PR #7, opened 2022-02-03
Update dependencies and use stable version

FireMasterK:
Use the stable version of `androidx.navigation:navigation-*`


alefvanoon:
Thanks Kavin!

Issue #27, opened 2022-02-14
Update channel page


dhishkyaon:
like this to use blank area better -

![Picsart_22-02-14_20-03-22-504.png](https://user-images.githubusercontent.com/88529066/153884263-c9a21cd6-154f-4160-924e-f22d3bb84f02.png)

dhishkyaon:
man i don't know c of coding. Just edited a screenshot. If you don't have enough time then its completly fine.

rimthekid:
No I have the rime thought you modified the code lol

dhishkyaon:
also please reduce the bottom navigation bar height.

PR #31, opened 2022-02-14
Update dependencies, improvements, TimeAgo changes, and StreamItem duration

### LINKED ISSUE #29: Show video time duration
something like this -

FireMasterK:
Closes #29


rimthekid:
Look at #27 should we change it to look like that?

FireMasterK:
That can be done too (in a different pr), the screenshot is to showcase the video duration on the thumbnails.

dhishkyaon:
can you change the transparency to zero?
Video duration will be more visible then.

FireMasterK:
> can you change the transparency to zero?
> Video duration will be more visible then.

The color was taken from NewPipe, I think it's good enough for now.

PR #70, opened 2022-03-16
YT subscriptions importer added

archroid:


app/build.gradle
```diff
@@ -75,4 +75,7 @@ dependencies {
     implementation 'com.squareup.retrofit2:converter-scalars:2.1.0'
 
     implementation 'com.arthenica:ffmpeg-kit-min:4.5.1.LTS'
+
+    implementation 'com.blankj:utilcode:1.30.0'
```
FireMasterK, line 79: Why add an additional library? You could just use a bufferedreader and use .lines() or .readLine()

lilcheti: He had some problems changing content:// to file path

Issue #42, opened 2022-02-19
[FEATURE] Pull Down to Refresh 


imyasharbakhshi:
As I noticed to Refresh each Tab we should Click again on Tab name , so I wonder it would be great to add pull down to Refresh in next build.

PR #77, opened 2022-03-17
Add pull down to refresh in subscriptions tab

Relwi:
https://github.com/libre-tube/LibreTube/issues/42
Added the SwipeRefreshLayout in the fragment_subscriptions.xml to refresh the content when user pulls down.

I keep the when user click in the tab it refresh the feed, because I don't know which UX do you prefer.


lilcheti:
Could you do it for Home too?

Relwi:
I will take a look tomorrow and if all is going well I will create a PR

FireMasterK:
Works perfectly on the feed page!

PR #79, opened 2022-03-17
Theme toggle switch and youtube subscription importer added!

archroid:


archroid:
If anyone has any other way to get the real URI without any extra libs just tell me.

archroid:
I tried so hard to just add the part of lib we want and remove unnecessary code but all files work with each other. It's wonderful how many lines he wrote to just get a correct uri for selected file :))) And it's not too much heavy. I think it's ok to just use it and don't ask for any other solutions.

lilcheti:
I should check the apk size after adding the library we do not want to make it any bigger than it's rn.

PR #84, opened 2022-03-18
Bottom navigation and toolbar behavior changes

Relwi:
Disabled the reselect behavior in BottomNavigationView, I think the feature to reselect the tabs is now I bit useless, since we have added the SwipeRefreshLayout.

Changes the toolbar navigation behavior to go back when in the settings fragment. Now there is only remains to hide the bottomNavigationView when on settings


Relwi:
Sincerely I couldn't find a way to hide the bottomNavigationView when user enters the settings fragment, so feel free to merge or close this PR.

PR #122, opened 2022-04-06
Add swipe to refresh on channel fragment

suraj0208:
Adds swipe to refresh on channels fragment.
Fix user unreadable strings that are shown while fetching channel details.

### DIFF
--- a/app/src/main/res/layout/fragment_channel.xml
+++ b/app/src/main/res/layout/fragment_channel.xml
@@ -1,92 +1,105 @@
 <?xml version="1.0" encoding="utf-8"?>
-<ScrollView xmlns:android="http://schemas.android.com/apk/res/android"
+<com.github.libretube.CustomSwipeToRefresh xmlns:android="http://schemas.android.com/apk/res/android"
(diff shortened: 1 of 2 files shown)


app/src/main/res/layout/fragment_channel.xml
```diff
@@ -1,92 +1,105 @@
 <?xml version="1.0" encoding="utf-8"?>
-<ScrollView xmlns:android="http://schemas.android.com/apk/res/android"
+<com.github.libretube.CustomSwipeToRefresh xmlns:android="http://schemas.android.com/apk/res/android"
```
suraj0208, line 2: Moved the ScrollView inside CustomSwipeToRefresh.
Added android:visibility="invisible" to ScrollView, so that it is not visible by default.
The ScrollView is made visible once channel details are fetched.
In case of any error, the screen remains blank. This is not the best way to handle this, once we implement network down error handling elsewhere either by toast or meessagebox, we can add it here as well.

FireMasterK:
Works as expected!

PR #133, opened 2022-04-16
Fix subscription imports on R.

FireMasterK:
Needs testing on devices below R, but I don't have any, all my devices are R


lilcheti:
I've tested it on android 5 and 10. It works!

PR #148, opened 2022-04-29
 Update channel page

Bnyro:
Changed the layout of the channel page (see picture) like suggested in #27 to make it take less space (if that is still a topic) and removed Description Label for better visuals (could add it again if needed but I think it looks better like that) 

Screenshot_20220429_194809

### DIFF
--- a/app/src/main/res/layout/fragment_channel.xml
+++ b/app/src/main/res/layout/fragment_channel.xml
@@ -15,7 +15,7 @@
 
         <LinearLayout
             android:layout_width="match_parent"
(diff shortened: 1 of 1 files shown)


Bnyro:
And pls tell me if there's sth to do to help you which is possible without much knowledge about Kotlin since I'm very new to it xD

FireMasterK (approved):

alefvanoon:
Thanks @Bnyro .
For better communication, you can join LibreTube room (#LibreTube:matrix.org) in matrix.

PR #158, opened 2022-05-03
Added navbar and changetheme strings

Bnyro:
Added some strings to strings.xml for better translation and already included them in German :)

### DIFF
--- a/app/src/main/java/com/github/libretube/ChannelFragment.kt
+++ b/app/src/main/java/com/github/libretube/ChannelFragment.kt
@@ -209,7 +209,7 @@ class ChannelFragment : Fragment() {
                 runOnUiThread {
                     view.findViewById<ScrollView>(R.id.channel_scrollView).visibility = View.VISIBLE
                     view.findViewById<TextView>(R.id.channel_name).text=response.name
-                    view.findViewById<TextView>(R.id.channel_subs).text=response.subscriberCount.formatShort() + " subscribers"
+                    view.findViewById<TextView>(R.id.channel_subs).text=response.subscriberCount.formatShort() + " " + resources.getString(R.string.subscribers)
                     view.findViewById<TextView>(R.id.channel_description).text=response.description
                     val bannerImage = view.findViewById<ImageView>(R.id.channel_banner)
                     val channelImage = view.findViewById<ImageView>(R.id.channel_image)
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/github/libretube/ChannelFragment.kt
```diff
@@ -209,7 +209,7 @@ class ChannelFragment : Fragment() {
                 runOnUiThread {
                     view.findViewById<ScrollView>(R.id.channel_scrollView).visibility = View.VISIBLE
                     view.findViewById<TextView>(R.id.channel_name).text=response.name
-                    view.findViewById<TextView>(R.id.channel_subs).text=response.subscriberCount.formatShort() + " subscribers"
+                    view.findViewById<TextView>(R.id.channel_subs).text=response.subscriberCount.formatShort() + " " + resources.getString(R.string.subscribers)
```
FireMasterK, line 212: Why not have the string as a format literal? In case, a language requires the count to be after the text

Bnyro: Good idea, but I don't know how to do this since I'm not so familiar with Kotlin. It would be nice if you could add it :)

Bnyro: That's a good idea. But the translators would have to implement into their translated string then? I don't think everybody's familiar with it.

Bnyro: > That's a good idea. But the translators would have to implement into their translated string then? I don't think everybody's familiar with it.

@FireMasterK

FireMasterK: I think it should be fine, if they use weblate since it would show the source string which would contains the formatting character too

For example, I already do this in Piped's web frontend.

Bnyro: K, gonna add it then

Bnyro: Seems like literals placed in the string.xml don't work because the literal in the strings.xml just gets replaced with nothing in the output while the other parts of the string can be seen (or I'm just too stupid to get it work)

Bnyro:
Didn‘t know that renaming a branch closes a pr, I‘m sry. Gonna reopen it later.

PR #165, opened 2022-05-05
Translation

Bnyro:


app/src/main/java/com/github/libretube/ChannelFragment.kt
```diff
@@ -209,7 +209,8 @@ class ChannelFragment : Fragment() {
                 runOnUiThread {
                     view.findViewById<ScrollView>(R.id.channel_scrollView).visibility = View.VISIBLE
                     view.findViewById<TextView>(R.id.channel_name).text=response.name
-                    view.findViewById<TextView>(R.id.channel_subs).text=response.subscriberCount.formatShort() + " subscribers"
+                    var subCount = response.subscriberCount.formatShort()
+                    view.findViewById<TextView>(R.id.channel_subs).text=resources.getString(R.string.subscribers).replace("#subCount", "$subCount")
```
FireMasterK, line 213: You should not do this, see: https://developer.android.com/guide/topics/resources/string-resource.html#FormattingAndStyling

Bnyro: Didn't know this is possible in Kotlin, gonna try it out later (sorry for wasting your time with sth like that :/)

FireMasterK: It's alright, it's not a waste of time :)

FireMasterK (approved):

PR #190, opened 2022-05-09
Subs and video count translation, video duration in search view

Bnyro:


lilcheti:
Could u add also video durations in the search?

Bnyro:
> Could u add also video durations in the search?

Sure, gonna do that tomorrow @lilcheti

PR #191, opened 2022-05-09
Added search history 

archroid:
Added new search history option. Still not completed. Any suggestions for the rest?


IceDBorn:
Any screenshots?

archroid:
> Any screenshots?

image

IceDBorn:
Seems nice, did you add an option to toggle this though?

archroid:
> Seems nice, did you add an option to toggle this though?

I will

app/src/main/java/com/github/libretube/SearchFragment.kt
```diff
@@ -143,4 +176,36 @@ class SearchFragment : Fragment() {
...
+        }
+
+
+
+        if (splited_history.size > 10) {
+            splited_history = splited_history.takeLast(10)
+        }
+
+
+        sharedPreferences.edit().putString("search_history", splited_history.joinToString("|"))
```
FireMasterK, line 200: What if I search for something that contains `|`?

archroid: Yeah you are right
Any suggestions?

FireMasterK: To write the objects: (I haven't tested if this works)
```kotlin
        val baos = ByteArrayOutputStream()
        ObjectOutputStream(baos).use {
            it.writeObject(arrayOf("1", "2", "3", "4", "5"))
        }
        baos.use {
            val encoded = Base64.encodeToString(it.toByteArray(), Base64.DEFAULT)
        }
```

archroid: Will test...
Thx

PR #227, opened 2022-05-17
Cronet

janisslsm:
I've attempted to implement cronet, I'm not sure if it makes any difference so I'd like some feedback to see if anyone notices a difference in speed.
A debug build should be available in checks from the CI

### DIFF
--- a/app/src/main/java/com/github/libretube/MainActivity.kt
+++ b/app/src/main/java/com/github/libretube/MainActivity.kt
@@ -45,6 +45,7 @@ class MainActivity : AppCompatActivity() {
         RetrofitInstance.engine = CronetEngine.Builder(applicationContext)
             .enableHttp2(true)
             .enableQuic(true)
+            .enableBrotli(true)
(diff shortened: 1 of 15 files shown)


FireMasterK:
I really don't like this approach of implementing cronet by copying it's source code, I would wait for them to release a maven dependency: https://github.com/google/cronet-transport-for-okhttp/issues/1

janisslsm:
> I really don't like this approach of implementing cronet by copying it's source code, I would wait for them to release a maven dependency: [google/cronet-transport-for-okhttp#1](https://github.com/google/cronet-transport-for-okhttp/issues/1)

Yeah, I'm aware of it. I'm just experimenting to see if cronet makes a difference

ag2s20150909:
exoplayer has cronet extensions

https://github.com/google/ExoPlayer/tree/release-v2/extensions/cronet

FireMasterK:
> exoplayer has cronet extensions
> 
> https://github.com/google/ExoPlayer/tree/release-v2/extensions/cronet

That's already being used since https://github.com/libre-tube/LibreTube/pull/130

PR #229, opened 2022-05-17
Autoplay and code formatting to satisfy ktlint

janisslsm:
This is an attempt at making the app automatically play the next video, it works fine but is this an acceptable solution?

@Bnyro @FireMasterK @lilcheti


IceDBorn:
You should add a switch to the settings menu.

Bnyro:
@IceDBorn Settings are already overfilled and I don't think that we would need a setting for that.

janisslsm:
What about a switch in the player screen?

Bnyro:
@janisslsm looks like a good idea :)

lilcheti:
Should I merge this now?

janisslsm:
> Should I merge this now?

Not yet, I haven't added a switch in the player yet. I've had a busy week

PR #230, opened 2022-05-17
Readme shorted + updated images

Bnyro:


Bnyro:
> Please don't remove fastlane image, We need them for f-droid.

@alefvanoon sry, didn't know that. Can I update them then?

PR #247, opened 2022-05-18
New icons

archroid:
Updated most of the icons to md3 outlined round version. 
image
image
image

Dont't know if its better or not


archroid:
Closed due to unpopularity.

FireMasterK:
I find these icons a bit lighter, and better fitting too.

Can't these icons be imported from a library or somewhere similar? I'm a bit unsure whether we should have our own drawables for each icon, although that's how it's currently done

archroid:
> I find these icons a bit lighter, and better fitting too.
> 
> Can't these icons be imported from a library or somewhere similar? I'm a bit unsure whether we should have our own drawables for each icon, although that's how it's currently done

I also found them better but people said old ones are better. Idk anyway.
Well that's what most of the apps do (vector drawables). And importing from a library will increase the apk size. It's unnecessary BTW.

Bnyro:
@archroid @FireMasterK We could try them, we can still roll back to the old ones if needed

archroid:
I will update icons and open another PR

PR #254, opened 2022-05-19
Format all code, in an attempt to fix the linter

FireMasterK:


Bnyro (approved):
Should be fine, at least the app still works :)

lilcheti:
we kinda broke the pr :')

PR #282, opened 2022-05-22
SponsorBlock Notifications Toggle

Bnyro:


app/src/main/java/com/github/libretube/PlayerFragment.kt
```diff
@@ -72,6 +72,7 @@ import java.io.IOException
 import java.net.URLEncoder
 import java.util.concurrent.Executors
 import kotlin.math.abs
+import org.chromium.base.ThreadUtils.runOnUiThread
```
FireMasterK, line 75: Should this really be used 🤔

I thought the activity class has a similar method

Bnyro: AutoImport xD

Bnyro: I don't even added sth that needs it

PR #485, opened 2022-06-13
App strings reworked 3

comradekingu:
Noticed when translating


app/src/main/res/values/strings.xml
```diff
@@ -164,13 +164,13 @@
...
-    <string name="empty_instance">You have to fill in the name and the API url.</string>
+    <string name="empty_instance">Fill in the name and the API URL.</string>
     <string name="clear_customInstances">Clear custom instances</string>
-    <string name="invalid_url">Please enter a valid url</string>
+    <string name="invalid_url">Please enter a URL that works</string>
     <string name="version">Version %1$s</string>
     <string name="about_summary">Get to know team LibreTube and how it all happens.</string>
     <string name="related_streams">Related streams</string>
-    <string name="related_streams_summary">Show related streams to videos.</string>
+    <string name="related_streams_summary">Show streams related to videos you watch.</string>
```
comradekingu, line 175:
```suggestion
    <string name="related_streams_summary">Show content related to what you watch.</string>
```

FireMasterK: This could be changed to "the video you are watching currently", or something similar.

PR #507, opened 2022-06-14
Wrapped thumbnail-duration text in a CardView

itsZECHS:
Just wrapped `thumnail_duration` TextView in a CardView to get those rounded corners

Before:
Screenshot from 2022-06-14 21-12-53

After:
Screenshot from 2022-06-14 21-12-59


Bnyro:
Looks amazing :)
It would be great if you could add the same thing to the video_search_row and the video_channel_row too so that it looks everywhere like that ;D

itsZECHS:
Updated other layouts too.

PR #544, opened 2022-06-18
"About" moved to the bottom

comradekingu:
"About" seems to always be at the bottom of menus (?)
outside of maybe "Exit", which isn't available in this one.


Bnyro:
Yes, you're right.
Unfortunately ktlint doesn't like you :)

PR #747, opened 2022-07-10
PlayerFragment.kt

NR8124:
made chapter name clickable in potrait mode too..

### DIFF
--- a/app/src/main/java/com/github/libretube/fragments/PlayerFragment.kt
+++ b/app/src/main/java/com/github/libretube/fragments/PlayerFragment.kt
@@ -374,7 +374,7 @@ class PlayerFragment : Fragment() {
         binding.linLayout.visibility = View.VISIBLE
         playerBinding.fullscreen.setImageResource(R.drawable.ic_fullscreen)
         playerBinding.exoTitle.visibility = View.INVISIBLE
-        playerBinding.chapterLL.isClickable = false
+        playerBinding.chapterLL.isClickable = true
 
         scaleControls(1F)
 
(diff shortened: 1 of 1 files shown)


Bnyro:
Why?
I disabled them because one should use the chapters menu in the video description in portrait mode...

NR8124:
because people have habits of this feature from the official youtube even me too

Bnyro:
The dialog doesn't include the chapter thumbnails so its worth than the other. But we can make the chapters in video description show when clicked on the chapter name.

Bnyro:
That's a better solution in my opinion

NR8124:
yeah that will help in better manner thanx
but some how it should highlight the ongoing chapter in description part too like
when we click on the chapter name.. it should take us to description and highlight which chapter we are watching

Bnyro:
Yes, true, I'll have a look later today maybe

NR8124:
can we make the changes to that chapters list dialogue itself
which will work in potrait as well as full screen mode

Bnyro:
I don't think that's possible

NR8124:
i think it can be possible, because it we every time open the description when clicked on chapter name in player.. i think doesn't seems good

instead we already have a dialogue which have all the chapters list.. so we can add the thumbnails right there and make it scrollable.

futher we just have to make some annimation change that the dialogue should appear from bottom and its height should be below player in potrait mode

this one dialogue of chapter can solve both potrait and full screen mode problem  regarding the chapters

Bnyro:
Dialogs can't show images

NR8124:
Not exactly but similar to this 


https://user-images.githubusercontent.com/88180887/178146819-60d665f8-5fbd-4af3-ad50-b83a8bc83071.mp4

we have our dialogue just need to rewrite to achieve this

Bnyro:
If "we" is you, then yes
Otherwise no

PR #852, opened 2022-07-22
Changed bindingBottomNav to Navigation to work on NavigationRail

Nephys222:
First pull request. Want to continue to contribute.


Bnyro:
First, I appreciate that your willing to help :)
But how should this work? :p
Viewbinding is used to connect the views in the xml files with the code. NavigationBarView is only a class, it doesn't reference to our actual navigation view. Your change would stop the app from running ^)

PR #941, opened 2022-08-01
in app crash handler

Bnyro:
Works but the app doesn't close when it should crash and shows a weird behavior (ExceptionHandler ll12). I tried throiwing the exception there with `throw exc` but it didn't work either...


Bnyro:
Rockets don't help me, I need someone to fix it lol

PR #942, opened 2022-08-01
fix for exiting searchBar

Nephys222:
Short fix for searchView exit


Bnyro:
Looks good, thank you <3

Nephys222:
Thanks. But it is a quick fix. There are some problems. When it is closed twice then it works as onBackpressed. Interpret there is no active search fragment so exit.

PR #950, opened 2022-08-02
App strings reworked 4

comradekingu:
List for future installments:
SponsorBlock options are a bit too long.
piped_summary" and similar aren't used?
Also,
Some options could be grayed out/removed when not active/available.


Bnyro:
Piped summary etc is shown by long pressing the Piped card in the about section
Agreed, SponsorBlock summaries could be shorter.

app/src/main/res/values/strings.xml
```diff
@@ -188,37 +188,37 @@
...
     <string name="deleteAccount_summary">Delete your Piped account</string>
     <string name="account">Account</string>
     <string name="restore">Restore</string>
     <string name="watch_history">Watch History</string>
     <string name="watch_positions">Remember position</string>
-    <string name="watch_positions_summary">Restore last watch position</string>
-    <string name="auth_instance">Authentication instance</string>
-    <string name="auth_instance_summary">Use a different instance for authenticated calls.</string>
+    <string name="watch_positions_summary">Continue from last playback position</string>
+    <string name="auth_instance">Log in elsewhere</string>
```
Bnyro, line 213: We should better leave that `Authentication instance`

app/src/main/res/values/strings.xml
```diff
@@ -188,37 +188,37 @@
...
+    <string name="auth_instance_summary">Uses a different instance.</string>
```
Bnyro, line 214: I would leave that too

app/src/main/res/values/strings.xml
```diff
@@ -227,66 +227,66 @@
...
     <string name="always">Always</string>
     <string name="selected">Selected</string>
     <string name="never">Never</string>
-    <string name="autoRotatePlayer">Auto fullscreen</string>
-    <string name="autoRotatePlayer_summary">Automatically switch to player fullscreen when the device gets turned.</string>
+    <string name="autoRotatePlayer">Auto-fullscreen</string>
+    <string name="autoRotatePlayer_summary">Fullscreen playback when the device gets turned.</string>
     <string name="pure_theme">Pure theme</string>
-    <string name="pure_theme_summary">Pure white/black theme</string>
+    <string name="pure_theme_summary">Only black and white</string>
```
Bnyro, line 243: I suggest keeping the old term here too (`Pure white/black theme`)

app/src/main/res/values/strings.xml
```diff
@@ -227,66 +227,66 @@
...
-    <string name="reset_watch_positions">Reset watch positions</string>
+    <string name="watch_positions_title">Remembered playback positions</string>
+    <string name="reset_watch_positions">Reset</string>
     <string name="system_caption_style">System caption style</string>
     <string name="captions">Captions</string>
     <string name="none">None</string>
-    <string name="update_now">Do you want to update the app now?</string>
-    <string name="seekbar_preview">Seekbar preview</string>
-    <string name="seekbar_preview_summary">Preview the video by seeking to the position when scrubbing the seekbar.</string>
+    <string name="update_now">Install the new LibreTube now?</string>
```
Bnyro, line 255: "Install the new LibreTube `version` now?" might be better

app/src/main/res/values/strings.xml
```diff
@@ -227,66 +227,66 @@
...
+    <string name="best_quality">Best</string>
+    <string name="worst_quality">Worst</string>
+    <string name="default_subtitle_language">Subtitle language</string>
     <string name="notifications">Notifications</string>
-    <string name="notify_new_streams">New streams notifications</string>
-    <string name="notify_new_streams_summary">Notify when new streams from subscriptions are available</string>
-    <string name="checking_frequency">Checking frequency</string>
-    <string name="new_streams_count">%1$s new streams are available</string>
-    <string name="new_streams_by">New streams by %1$s …</string>
+    <string name="notify_new_streams">Livestream notifications</string>
```
Bnyro, line 269: It has nothing to do with livestreams, it just notifies for new uploaded videos by channels the user subscribed to

app/src/main/res/values/strings.xml
```diff
@@ -227,68 +227,68 @@
...
-    <string name="reset_watch_positions">Reset watch positions</string>
+    <string name="watch_positions_title">Remembered playback positions</string>
+    <string name="reset_watch_positions">Reset</string>
     <string name="system_caption_style">System caption style</string>
     <string name="captions">Captions</string>
     <string name="none">None</string>
-    <string name="update_now">Do you want to update the app now?</string>
-    <string name="seekbar_preview">Seekbar preview</string>
-    <string name="seekbar_preview_summary">Preview the video by seeking to the position when scrubbing the seekbar.</string>
+    <string name="update_now">Install the new LibreTube now?</string>
```
comradekingu, line 255:
```suggestion
    <string name="update_now">Install the new LibreTube version now?</string>
```
Thought it sounded flashier without "version", to entice people to do it ;)

app/src/main/res/values/strings.xml
```diff
@@ -227,68 +227,68 @@
...
+    <string name="best_quality">Best</string>
+    <string name="worst_quality">Worst</string>
+    <string name="default_subtitle_language">Subtitle language</string>
     <string name="notifications">Notifications</string>
-    <string name="notify_new_streams">New streams notifications</string>
-    <string name="notify_new_streams_summary">Notify when new streams from subscriptions are available</string>
-    <string name="checking_frequency">Checking frequency</string>
-    <string name="new_streams_count">%1$s new streams are available</string>
-    <string name="new_streams_by">New streams by %1$s …</string>
+    <string name="notify_new_streams">Livestream notifications</string>
```
comradekingu, line 269:
```suggestion
    <string name="notify_new_streams">Subscription notifications</string>
```

app/src/main/res/values/strings.xml
```diff
@@ -227,68 +227,68 @@
...
+    <string name="notify_new_streams_summary">Look for new subscribed streams</string>
```
comradekingu, line 270:
```suggestion
    <string name="notify_new_streams_summary">Notifications about fresh content from creators you follow.</string>
```

app/src/main/res/values/strings.xml
comradekingu: My thinking is that "Black and white" is a natural linguistic concept.
Moreover, it could be just "Black and white" in the title, without any summary.
This again begs the question of why it isn't a selection in the theme alternatives to begin with.
Because it isn't a black and white version of the selected theme(?)
Also both require an app restart, so no loss(?) making it one category.

On my device, the Appearance section scrolls by one entry.
Doing away with the toggle would help if not fix that completely.

app/src/main/res/values/strings.xml
```diff
@@ -227,68 +227,68 @@
...
     <string name="always">Always</string>
     <string name="selected">Selected</string>
     <string name="never">Never</string>
-    <string name="autoRotatePlayer">Auto fullscreen</string>
-    <string name="autoRotatePlayer_summary">Automatically switch to player fullscreen when the device gets turned.</string>
+    <string name="autoRotatePlayer">Auto-fullscreen</string>
+    <string name="autoRotatePlayer_summary">Fullscreen playback when the device gets turned.</string>
     <string name="pure_theme">Pure theme</string>
-    <string name="pure_theme_summary">Pure white/black theme</string>
+    <string name="pure_theme_summary">Only black and white</string>
```
comradekingu, line 243:
```suggestion
    <string name="pure_theme_summary">Pure white/black theme</string>
```

app/src/main/res/values/strings.xml
comradekingu: The benefit escapes me.
I think that may be the official term, but it seems the actual meaning alludes people not already technically adept.
Yes it is a setting for technical users, but it is also important that regular ones understand not to use it, and more importantly aren't alarmed by its presence.

The question that arises is whether an authentication instance is needed in its own right.
This might lead people to try to input something, not knowing that there is already a sane default.
If the user doesn't know what "authenticated calls" are, in all likelihood it could be that the user thinks
that is an extra benefit available when connecting to a "different instance" to do so.

The direct translations aren't any better, and the attempted ones are more misleading.
They are also longer.

PR #981, opened 2022-08-06
fix fullscreen on devices with display cutouts

AlexGmach:


Bnyro:
Works great, thank you <3

PR #1026, opened 2022-08-12
Fix readme and compress some images

wahyuwiyoko:
Fixed images URL in readme and compress some images.


wahyuwiyoko:
How about changing the "Features" section to list instead of table like this:

- [x] User Accounts   
- [x] Subscriptions     
- [x] User Playlists    
- [x] Channel Playlists 
- [x] Search Filters   
- [x] SponsorBlock   
- [x] Subtitles      
- [x] Comments

PR #1214, opened 2022-09-05
Add Instance adminForge.de

Dominion0815:


FireMasterK:
Any reason why you're not adding your instance to the list Piped maintains? Libretube uses it, via https://piped-instances.kavin.rocks/

If you'd like me to add it there, let me know!

Dominion0815:
Oh how it's possible?

Bnyro:
Kavin is right, we don't use the array-list you changed for instances anymore.

Best way is to either open an issue on [the Piped repo](https://github.com/teampiped/piped) providing some info about the instance (region, CDN,...) or just provide the info here and make it up with Kavin :)

Bnyro:
Looks like Kavin already added it haha :p

PR #1215, opened 2022-09-06
Fixed Iran region has no videos in feed

archroid:


Bnyro:
We should rather remove it from the array of available countries instead of automatically selecting a different region without letting the user know.

(And using ktlint would be great for the next time)

Bnyro:
And the string should be translatable and not hard coded!

Bnyro:
I'll revert it and add something to handle that issue properly when I'm back home.

archroid:
OK then

alefvanoon:
Don't need to do this!
Iran has this problem in the YouTube app as well. Because YouTube is blocked in Iran.

archroid:
Yeah. My feed was almost empty for a month after implementing the auto
system region option and I was wondering what the problem was. I checked
the code and I found it but users may have problems with it. It would be
better to switch to US or any other supported region so the feed for
Iranians wouldn't be empty

Bnyro:
Maybe we should just show a toast that informs users it's recommended to try another region when the API didn't return any streams.

I don't really want to test all the regions and whether they work or not to exclude them manually if they don't (and it'd also be a bad idea to remove them completely because they might work again in the future).

Bnyro:
Switching region without the user knowing it can lead to confusion.

archroid:
Yeah. That's the best solution.

PR #1218, opened 2022-09-06
fix for searchview anormal behavior

Nephys222:
Cleared removeSearchFocus which led to open searchHistory. Cleared searchOnCloseListener which is dysfunctional anymore.


Bnyro:
Works great, thanks for the fix!

PR #1314, opened 2022-09-18
Use NotificationChannelCompat.

Isira-Seneviratne:
Use `NotificationChannelCompat` for notification channel creation.


Bnyro (approved):
LGTM, thank you!

PR #1327, opened 2022-09-19
Translate configs fields to portugues

karontyy:


Bnyro:
Thank you for your interest in contributing.
However, translations are handled via Weblate, please head over to https://hosted.weblate.org/projects/libretube/#languages and translate the missing strings there.
Thanks!

PR #1388, opened 2022-09-26
add option to limit hls to 1080p

### LINKED ISSUE #1337: add one resolution choice: "auto but no larger than 1080p"

Considering all experiences I had, regarding resolution and poor internet connection, I'm thinking about this:

Current resolution system: 
if I choose say 720p, if this video does not have 720p, it will play 480p; if it has 720p, it will stick to 720p. ( Problem is, if connection begins to get poor, playback will get stuck often, that's why Auto is a better choice.)
if I choose Auto, if connection is poor it will use say 360p, if connection is great it will use 2160p. ( Problem is, when it uses 2160p, it sometimes (once or twice for a video) breaks/stops to a black screen, and I have to click play button to continue, I don't know why it does not drop down to 1440p or 1080p instead. And for me I find 1080p is always fluent, when in good connection time of day.)  

How about:
add one setting: good enough resolution, say 1080p;
add one choice in resolution drop down menu: auto but no larger than good enough ?

Thus, not matter connection is poor or good, it will be fluent.  For people who has an excellent connection and 2160p is always easy, "good enough resolution" set to 2160p and nothing will go wrong.

Please consider it.

### Other details

_No response_

### Acknowledgements

Bnyro:
closes #1337

### DIFF
--- a/app/src/main/java/com/github/libretube/constants/PreferenceKeys.kt
+++ b/app/src/main/java/com/github/libretube/constants/PreferenceKeys.kt
@@ -77,6 +77,7 @@ object PreferenceKeys {
     const val PICTURE_IN_PICTURE = "picture_in_picture"
     const val PLAYER_RESIZE_MODE = "player_resize_mode"
     const val SB_SKIP_MANUALLY = "sb_skip_manually_key"
+    const val LIMIT_HLS = "limit_hls"
 
     /**
      * Background mode
@@ -101,8 +102,6 @@ object PreferenceKeys {
     const val CLEAR_WATCH_HISTORY = "clear_watch_history"
     const val CLEAR_WATCH_POSITIONS = "clear_watch_positions"
     const val SHARE_WITH_TIME_CODE = "share_with_time_code"
-    const val BACKUP_SETTINGS = "backup_settings"
-    const val RESTORE_SETTINGS = "restore_settings"
 
     /**
      * History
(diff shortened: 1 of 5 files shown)


ydqsbfsyi:
@Bnyro 
Is it possible to get back this option? - in recent versions, this option is lost, and I don't want to roll back to old version, since I do really need some new features recently:
```
pause and play on double click at the center;
option to not auto-insert related videos;
```
man! that's incredible.

Bnyro:
Currently you can watch HLS with 1080p when choosing it as quality in the player options menu. For limiting the automatic quality you could create a new issue, but I don't really get why you would need it as it works to set a fixed resolution for HLS.

ydqsbfsyi:
> Currently you can watch HLS with 1080p when choosing it as quality in the player options menu. For limiting the automatic quality you could create a new issue, but I don't really get why you would need it as it works to set a fixed resolution for HLS.

I wrote an issue before, based on which you kindly added this option. In short, for my bad internet connection, if a video has 2k or 4k, auto(HLS) format is easy to break and pause, I need to manually resume play. 

One question please: is fixed 1080p HLS or DASH? or could be both which I could control with "Use HLS" option? I could accept  a fixed HLS, which seems faster for me.

Bnyro:
The fixed resolution uses whatever you chose in settings. If you enable "Use HLS" in the audio/video settings it always uses HLS, otherwise it always uses DASH. So yes, you can change the resolution using HLS sincr the latest update :)

ydqsbfsyi:
> The fixed resolution uses whatever you chose in settings. If you enable "Use HLS" in the audio/video settings it always uses HLS, otherwise it always uses DASH. So yes, you can change the resolution using HLS sincr the latest update :)

I see. Thanks for your help.

PR #1428, opened 2022-09-30
Do not stat pip when open video externally

DennisBauer:
When opening a video in an external player picture in picture should not be started.


Bnyro (approved):
LGTM, thank you!

PR #1464, opened 2022-10-02
Aesthetic changes

GilesMunn:


app/src/main/java/com/github/libretube/ui/adapters/TrendingAdapter.kt
```diff
@@ -53,7 +53,7 @@ class TrendingAdapter(
             textViewTitle.text = trending.title
             textViewChannel.text =
                 trending.uploaderName + " • " +
-                trending.views.formatShort() + " • " +
+                trending.views.formatShort() + " views • " +
```
FireMasterK, line 56: I think this should be in `strings.xml` since this would need to be translated too.

GilesMunn: Good point, I'll move it!

PR #1534, opened 2022-10-10
GNU AFFERO GENERAL PUBLIC LICENSE  Version 3, 19 November 2007

bell-kevin:
GNU AFFERO GENERAL PUBLIC LICENSE  Version 3, 19 November 2007

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,5 +1,5 @@
-                    GNU GENERAL PUBLIC LICENSE
-                       Version 3, 29 June 2007
+                    GNU AFFERO GENERAL PUBLIC LICENSE
+                       Version 3, 19 November 2007
 
  Copyright (C) 2007 Free Software Foundation, Inc. <https://fsf.org/>
  Everyone is permitted to copy and distribute verbatim copies
@@ -7,17 +7,15 @@
 
                             Preamble
 
-  The GNU General Public License is a free, copyleft license for
-software and other kinds of works.
+  The GNU Affero General Public License is a free, copyleft license for
+software and other kinds of works, specifically designed to ensure
+cooperation with the community in the case of network server software.
 
   The licenses for most software and other practical works are designed
 to take away your freedom to share and change the works.  By contrast,
-the GNU General Public License is intended to guarantee your freedom to
+our General Public Licenses are intended to guarantee your freedom to
 share and change all versions of a program--to make sure it remains free
-software for all its users.  We, the Free Software Foundation, use the
-GNU General Public License for most of our software; it applies also to
-any other work released this way by its authors.  You can apply it to
-your programs, too.
+software for all its users.
 
   When we speak of free software, we are referring to freedom, not
 price.  Our General Public Licenses are designed to make sure that you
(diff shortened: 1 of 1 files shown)


Bnyro:
What's better or different about it?

bell-kevin:
Piped has an AGPL. I suggest we do the same. https://github.com/TeamPiped/Piped/blob/master/LICENSE

Bnyro:
But how exactly? I won't compare them now ;)

bell-kevin:
The GNU Affero General Public License is a modified version of the ordinary GNU GPL version 3. It has one added requirement: if you run a modified program on a server and let other users communicate with it there, your server must also allow them to download the source code corresponding to the modified version running there.

The purpose of the GNU Affero GPL is to prevent a problem that affects developers of free programs that are often used on servers.

Suppose you develop and release a free program under the ordinary GNU GPL. If developer D modifies the program and releases it, the GPL requires him to distribute his version under the GPL too. Thus, if you get a copy of his version, you are free to incorporate some or all of his changes into your own version.

But suppose the program is mainly useful on servers. When D modifies the program, he might very likely run it on his own server and never release copies. Then you would never get a copy of the source code of his version, so you would never have the chance to include his changes in your version. You may not like that outcome.

Using the GNU Affero GPL avoids that outcome. If D runs his version on a server that everyone can use, you too can use it. Assuming he has followed the license requirement to let the server's users download the source code of his version, you can do so, and then you can incorporate his changes into your version. (If he hasn't followed it, you have your lawyer complain to him.)

Both the ordinary GNU GPL, version 3, and the GNU Affero GPL have text allowing you to link together modules under these two licenses in one program.

The GNU Affero GPL does not address the problem of [Service as a Software Substitute (SaaSS)](https://www.gnu.org/philosophy/who-does-that-server-really-serve.html).

SaaSS means that users use someone else's web server to do their own computing. This requires them to send their data to the server, which does their computing for them and sends the results back to them. SaaSS is an injustice because the users cannot control their computing when it's done that way.

If some program on this server is released under the GNU Affero GPL, the server is required to offer the users the corresponding source of that program. That is good, but having this source code does not give them control over the computing done on that server. It also does not tell them what other software may be running on that server, examining or changing their data in other ways.

We don't see any sensible way to address the SaaSS problem with license conditions on particular programs. Even to write a legal condition to distinguish between SaaSS use and non-SaaSS use would be a challenge, and if we had that, we don't see anything that the program's license might require in the SaaSS case that would correct the inherent wrong of SaaSS. Thus, our solution to the problem of SaaSS is simple: refuse to use services that are SaaSS.

If a program is meant specifically and only for SaaSS, you shouldn't write it. But many programs are useful for a variety of kinds of services, including some that are SaaSS and some that are not. It's useful to write and release these programs so people can set up non-SaaSS services with them, and good to release them under the AGPL.

Bnyro:
Well, LibreTube is a client and not a server-side application so the AGPL wouldn't change anything except some unneeded clauses in the end. Therefore I'll close the PR, but anyways thanks for your interest in helping to secure the apps future!

bell-kevin:
You're welcome!

FireMasterK:
Also, changing license would require all past contributors to release their contributions under the AGPL. (Otherwise, their code would be under the GPL)

bell-kevin:
@FireMasterK , why is @piped AGPL?

Issue #1528, opened 2022-10-09
Simplify download notification


GilesMunn:
### Describe your suggested feature

At present, when downloading a video with audio, the notification shows as "2 files downloading" and has separate entries for both parts of the file. This is arguably unnecessary and over-complicated, it'd be nice if this were simplified to just the name of the video that is downloading.

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

shantanu1k:
@Bnyro @GilesMunn Can I work on this?

Bnyro:
> @Bnyro @GilesMunn Can I work on this?

Of course, I appreciate any help <3

PR #1640, opened 2022-10-22
Simplified the download notification

shantanu1k:
<h3> PR for issue #1528 <h3/>

* Before

photo_2022-10-22_16-10-21

* After

photo_2022-10-22_16-11-10


app/src/main/java/com/github/libretube/services/DownloadService.kt
```diff
@@ -74,17 +74,17 @@ class DownloadService : Service() {
...
+                    videoName,
                     getString(R.string.downloading),
                     videoUrl,
                     Uri.fromFile(
                         File(videoDownloadDir, videoName)
                     )
                 )
             }
-            if (downloadType in listOf(DownloadType.AUDIO, DownloadType.AUDIO_VIDEO)) {
+            else if (downloadType in listOf(DownloadType.AUDIO, DownloadType.AUDIO_VIDEO)) {
```
Bnyro, line 85: Adding an else here will prevent video and audio to be downloaded with one request so we will need a different way to handle that one. I think it would be the easiest to give the second created notification just a low priority so that only the first one with the video name as title is visible.

Bnyro (changes requested):

app/src/main/java/com/github/libretube/services/DownloadService.kt
shantanu1k: @Bnyro You're right! That's clumsy of me...😅 
The approach of using low priority(HIDDEN) notification needs the permission: _(android.permission.DOWNLOAD_WITHOUT_NOTIFICATION)_ 

If that's fine..Can I proceed with it?

Bnyro: Hmm, another alternative would be to just use the video title for both, the video and the audio notification but that would suck too as the title would exist twice then.
So I think we'll need to live with the additional permission, so feel free to just go on :)

app/src/main/java/com/github/libretube/services/DownloadService.kt
```diff
@@ -74,19 +75,21 @@ class DownloadService : Service() {
             )
             if (downloadType in listOf(DownloadType.VIDEO, DownloadType.AUDIO_VIDEO)) {
                 videoDownloadId = downloadManagerRequest(
-                    getString(R.string.video),
+                    videoName,
                     getString(R.string.downloading),
                     videoUrl,
+                    false,
```
Bnyro, line 81: Now, when only downloading the audio stream of a video, there simply won't be any notification :wink:

app/src/main/java/com/github/libretube/services/DownloadService.kt
```diff
@@ -132,6 +136,9 @@ class DownloadService : Service() {
                 .setDestinationUri(destination)
                 .setAllowedOverMetered(true) // Set if download is allowed on Mobile network
                 .setAllowedOverRoaming(true)
+        if (isAudio) {
```
Bnyro, line 139: I would just use a global variable `isNotificationSet` which is false by default and set it to `true` in that scope after creating the first download request with a notification, and check here whether `isNotificationSet` equals true and decide whether to show the notification accordingly :)

shantanu1k: Hi @Bnyro , I tried your solution....it works fine. But the problem occurs when I try to cancel the ongoing (video + audio) download, only video downloading gets cancelled because it is visible. The audio downloading remains unaffected because it is **hidden** notification & cannot be cancelled. The audio only & video only works fine 👍
<br/>

https://user-images.githubusercontent.com/57897297/197395923-ffa48cf4-a5d2-4151-a879-7f200ccf20db.mp4

<br/>
I'll need some time to figure out another way :)

Bnyro: Oh, damn, didn't think about that ...
We still have the option to just show the video name as title for both downloads which of course isn't the best UX-wise solution butter it'd be better than nothing ig :)
But of course there's still hope that we're finding a better way, I'll let you know if I got any other idea how it could be solved ;)

shantanu1k: For now, can we show the notifications like this?

* **Audio & Video** <br/>
 

* **Video** <br/>
 

* **Audio** <br/>

Bnyro: I'd be okay with that, we can still change it later if we'll ever find a better way to solve this :wink:

app/src/main/java/com/github/libretube/services/DownloadService.kt
```diff
@@ -74,7 +73,7 @@ class DownloadService : Service() {
             )
             if (downloadType in listOf(DownloadType.VIDEO, DownloadType.AUDIO_VIDEO)) {
                 videoDownloadId = downloadManagerRequest(
-                    getString(R.string.video),
+                    "[Video] $videoName",
```
Bnyro, line 76: Please use translatable strings for that (which already exist), you can access them e.g. with `getString(R.string.audio)`

Bnyro:
Last but not least, please run `ktlint --android -F` in the root directory of the project to fix the linter :)

app/src/main/java/com/github/libretube/services/DownloadService.kt
```diff
@@ -60,7 +60,7 @@ class DownloadService : Service() {
         TODO("Not yet implemented")
     }
 
-    private fun downloadManager() {
+    private fun downloadManager(videoName: String) {
```
Bnyro, line 63: `videoName` is reachable in that scope that means it doesn't need to get passed to the function manually.

shantanu1k:
lol! At the end this is literally a minor fix :P

Bnyro:
We spent hours with 2 changed lines now haha :)

Bnyro:
Thanks for your contribution, looking forward to your next one ;)

PR #1641, opened 2022-10-22
Added metadata to lockscreen. Closes #1470.

### LINKED ISSUE #1470: Show metadata on lockscreen

Currently, ExoPlayer does not share the title of the current video with the MediaMetadata function of Android. This means that LibreTube will display as "No title" on the lockscreen - similar to here: https://github.com/google/ExoPlayer/issues/6624.

It'd be nice if - as mentioned in this issue - we could share just the title of the current video with MediaMetadata so that we can see what is currently playing when the phone is locked.

### Other details

_No response_

### Acknowledgements

GilesMunn:
closes #1470


app/src/main/java/com/github/libretube/util/NowPlayingNotification.kt
```diff
@@ -169,6 +171,19 @@ class NowPlayingNotification(
             setUseFastForwardActionInCompactView(true)
             setUseRewindActionInCompactView(true)
         }
+        mediaSessionConnector = MediaSessionConnector(mediaSession)
```
Bnyro, line 174: That'd be duplicated code, see a few lines above

app/src/main/java/com/github/libretube/util/NowPlayingNotification.kt
```diff
@@ -129,6 +137,27 @@ class NowPlayingNotification(
...
+            override fun getMediaDescription(
+                player: Player,
+                windowIndex: Int
+            ): MediaDescriptionCompat {
+                return MediaDescriptionCompat.Builder().apply {
+                    setTitle(streams?.title!!)
+                    setSubtitle(streams?.uploader)
+                    val extras = Bundle()
+                    val appIcon = BitmapFactory.decodeResource(
+                        Resources.getSystem(),
```
Bnyro, line 150: `Resources.getSystem()` only accesses the system-specific resources, but it can't access LibreTube internal resources such as an app icon, so changing that line might fix the app icon :)

PR #1666, opened 2022-10-27
Action version update

rozari0:
This will help to get rid of deprecated warning in workflow.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: gradle/wrapper-validation-action@v1
-      - uses: actions/setup-python@v3
+      - uses: actions/setup-python@v4
         with:
           python-version: '3.x' # Version range or exact version of a Python version to use, using SemVer's version range syntax
           architecture: 'x64' # optional x64 or x86. Defaults to x64 if not specified


Bnyro:
I tried bumping it once but the Telegram bot failed then, not sure what the issue was ...

Bnyro:
I'll merge it and we see what happens!

Bnyro:
Fair, Github Mobile doesn't allow me to, we'll need to wait until I return back home

rozari0:
> I tried bumping it once but the Telegram bot failed then, not sure what the issue was ...

No error for me. 🤷‍♂️
ScreenShot

FireMasterK:
It seems to be failing previously too and fails even now as well. It seems like it's running out of ram, and hence being killed :confused:

FireMasterK:
Ugh, the script appears to be running a random binary: https://github.com/LibreTubeAlpha/Archive/blob/main/exec I can't find out what this does as there's no source code

rozari0:
> It seems to be failing previously too and fails even now as well. It seems like it's running out of ram, and hence being killed 😕

Nope, tg uploader script forcely closing python after completing upload. I'm writing alternative one with pyrogram. I'll make a pr later.

rozari0:
> Ugh, the script appears to be running a random binary: https://github.com/LibreTubeAlpha/Archive/blob/main/exec I can't find out what this does as there's no source code

This binary won't be needed anymore after merging #1669

PR #1669, opened 2022-10-28
TG upload with pyrogram.

rozari0:
@FireMasterK It's complete now. :p

### DIFF
--- a/.github/tg.py
+++ b/.github/tg.py
@@ -1,33 +1,34 @@
-import telegram
-from tgconfig import *
+import asyncio
 from json import load
-import multiprocessing
-from os import system
-from time import sleep as wait
+from os import listdir
 
-def deploy():
-    system(f'./exec --local --api-id={TG_API_ID} --api-hash={TG_API_HASH}')
+from pyrogram import Client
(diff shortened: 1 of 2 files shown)


.github/tg.py
```diff
@@ -1,33 +1,34 @@
...
+caption = f"""**Libretube {data['sha'][0:7]} // Alpha**
+
+<a href="{data['html_url']}">{data['commit']['message']}</a>
+
+Signed-off-by: {data['commit']['author']['name']}
+"""
+
+
+async def main():
+    async with Client("libretube", TG_API_ID, TG_API_HASH, bot_token=TG_TOKEN) as app:
```
Bnyro, line 27: The second `async` shouldn't be needed here as the `main` function is already async, or is there any specific reason for having async two times here?

rozari0: > The second `async` shouldn't be needed here as the `main` function is already async, or is there any specific reason for having async two times here?

Code won't work, you'll get an AttributeError. :(

Bnyro:
Thanks! <3

Issue #1654, opened 2022-10-24
Sharing video or channel via email


RNavarro3000:
### Describe your suggested feature

When sharing video or channel via email, include filled in subject field.

### Other details

Pipe title from video or channel to subject field for email.

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

shantanu1k:
@RNavarro3000 Like Autofill the subject field of composing email with video/channel title?

RNavarro3000:
Correct...

shantanu1k:
Hi @Bnyro , we might need to add additional parameter (videoTitle/channelTitle/playlistTitle) to ```ShareDialog``` constructor & to all of its instances. Lots of drilling......What do you think?

Bnyro:
> Hi @Bnyro , we might need to add additional parameter (videoTitle/channelTitle/playlistTitle) to ```ShareDialog``` constructor & to all of its instances. Lots of drilling......What do you think?

We could create a class `ShareData` for that which contains all the different possible information and just null them by default.
I'm on vacation so I won't work on that in the next time but feel free to do so :)

shantanu1k:
Hi @Bnyro , By creating a ```ShareData``` class , you mean Singleton class, right? Or something else? I've tried with Singleton & works for video(after opening the exo player) & channel. It doesn't work when we long press on video in the list & share with context menu.

Bnyro:
> Hi @Bnyro , By creating a ```ShareData``` class , you mean Singleton class, right? Or something else?
> 
No, I just mean creating a data class that is passed to the function with the variables `timeCode`, `channelName`,  etc.
All are nulled by default which means in the ShareDialog we just check whether `shareData.channelName != null` and run the corresponding code then!

Bnyro:
Not sure what you tried but in the end it should work for all the created share dialogs in the same way!

PR #1671, opened 2022-10-28
Sharing video or channel via email 

shantanu1k:
<h3>PR for issue #1654</h3>

* **Before**
_Subject is not auto filled_

* **After**

https://user-images.githubusercontent.com/57897297/198666694-0fd9bd34-a076-412b-bf30-bb6fe24b8e54.mp4


app/src/main/java/com/github/libretube/ui/dialogs/ShareDialog.kt
```diff
@@ -104,4 +107,14 @@ class ShareDialog(
         }
         return ""
     }
+    private fun getShareableTitle(shareData: ShareData): String {
+        if (shareData.currentChannel != null) {
```
Bnyro, line 111: I'd use the Kotlin `let` for that for a cleaner syntax, e.g.
```kotlin
shareData.currentChannel?.let {
  return it
}
```

app/src/main/java/com/github/libretube/ui/fragments/ChannelFragment.kt
```diff
@@ -29,7 +30,7 @@ class ChannelFragment : BaseFragment() {
 
     private var channelId: String? = null
     private var channelName: String? = null
-
+    private lateinit var shareData: ShareData
```
Bnyro, line 33: That variable doesn't need to be initialized here, it should be sufficient to just initialize where it's being used below

Bnyro:
Looks good to me except some minor things, thank you :)
Please don't forget to run `ktlint --android -F` in the root directory of the project to fix the linter :)

PR #1716, opened 2022-10-31
Auto-sizing the video options text

### LINKED ISSUE #1705: Align video options
### Steps to reproduce

Try to open any video and see options in French language, you will find that "save" option not aligned like others

### Expected behavior

Download option,"enregistrer" could be aligned
Screenshot_20221030-092248_LibreTube

### Actual behavior

There are a text that's not hided like "open", " ouvrir" button

Screenshot_20221030-092311_LibreTube

### LibreTube version

0.7.0

### Android version

Android 11

### Other details

_No response_

### Acknowledgements

shantanu1k:
<h2>Resolves #1705</h2>

* <h3>Before</h3>

* <h3>After</h3>

<h3>Greek</h3> 

<h3>French</h3> 

<h3>Portuguese</h3>

### DIFF
--- a/app/src/main/res/layout/fragment_player.xml
+++ b/app/src/main/res/layout/fragment_player.xml
@@ -165,12 +165,8 @@
                         <TextView
                             android:layout_width="wrap_content"
                             android:layout_height="wrap_content"
-                            app:autoSizeTextType="uniform"
-                            app:autoSizeMinTextSize="8sp"
-                            android:gravity="center"
-                            android:maxLines="1"
-                            android:text="@string/share"
-                            tools:targetApi="o" />
+                            style="@style/PlayerActionsText"
+                            android:text="@string/share" />
                     </LinearLayout>
 
                     <LinearLayout
@@ -186,12 +182,8 @@
                         <TextView
                             android:layout_width="wrap_content"
                             android:layout_height="wrap_content"
-                            app:autoSizeTextType="uniform"
-                            app:autoSizeMinTextSize="8sp"
-                            android:gravity="center"
-                            android:maxLines="1"
-                            android:text="@string/download"
-                            tools:targetApi="o" />
+                            style="@style/PlayerActionsText"
+                            android:text="@string/download" />
                     </LinearLayout>
 
                     <LinearLayout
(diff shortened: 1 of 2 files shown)


Bnyro:
I'd prefer if we had a new style for it defined in `values/style.xml` to get rid of the repeating code here, that'll also make it easier to potentially tweak it a bit in the future ;)

Bnyro:
Also, if you write something like `resolves #1705` or `closes #1705` GitHub will automatically close the issue once the PR is merged :)

VedxP:
Wouldn't it look irregular text sizing if it's set to auto?

I mean, buttons having long names would have a smaller font size whereas ones having shorter names would have larger font size :(

shantanu1k:
Yeah, that looks irregular **:(** 
@VedxP @Bnyro 
how about adding a horizontal scroll to that CardView just like YouTube app? then we can add even more options in the future if needed **:)**

Bnyro:
> how about adding a horizontal scroll to that CardView just like YouTube app? then we can add even more options in the future if needed **:)**
>
I'm not a fan of that as not all of the controls would be visible at the same time which means some would be more difficult to access for users than others. 
In my opinion we only have two options in the end: auto-sizing the TextViews or limiting it to one line and cut off the rest of the text without requiring a redesign of the player options.
So I think we should just take the auto-size approach and maybe align the text a bit differently to make it look more consistent :)

VedxP:
> 

That would be really beneficial since we don't have to make texts auto-resize and would fix  #1705 too.

But, all over to @Bnyro :)

shantanu1k:
Aligning the option texts at the `bottom` looks somewhat better but no option to align with `baseline` which would look even better @Bnyro @VedxP 


photo_2022-11-01_20-28-33

VedxP:
> Aligning the option texts at the `bottom` looks somewhat better but no option to align with `baseline` which would look even better @Bnyro @VedxP
> 
> photo_2022-11-01_20-28-33

Hmm, I would still say it'll create inconsistency in size, and even won't follow MD guideline if I'm not wrong.

Rather, trimming the text when it's long; like the `Open...` button, for all the buttons would be a better idea, in my opinion :)

Bnyro:
Hmm, but for example in French, if we only had `Enregi...` instead of `Enregistrer` I doubt that it would still be clear what the control is gonna be doing in the end.

The auto-size is not an ideal solution, I agree, but it's probably still the best one concerning UX, so if there are no other arguments against it I'd just merge it later.

Thank you @shantanu1k

VedxP:
Yups, go ahead :)

PR #1778, opened 2022-11-06
Playlist page redesign

Bnyro:


VedxP:
@Bnyro 
- Make Title Size Smaller, or something else so that the menu icon doesn't lose legibility when Title Text is Long.

> Screenshot_20221106-180723~2

PR #1867, opened 2022-11-16
Implement proper dash support

### LINKED ISSUE #1849: Inconsistent order of qualitys
### Steps to reproduce

Click on a video in 60fps 
And change resolution 

### Expected behavior

The resolution should get listed as following
1440p60
1080p60
720p60
720p
480p
...

### Actual behavior

It gets listed like this:
1440p60
1080p60
720p
720p60
480p
...

### LibreTube version

0.7.1

### Android version

Android 13

### Other details

_No response_

### Acknowledgements

FireMasterK:
This is so big, that this deserves its own release haha

closes #1849 as it's not a thing anymore

### DIFF
--- a/app/src/main/java/com/github/libretube/util/DashHelper.kt
+++ b/app/src/main/java/com/github/libretube/util/DashHelper.kt
@@ -69,7 +69,7 @@ object DashHelper {
             adapSetInfos.add(
                 AdapSetInfo(
                     stream.mimeType!!,
-                    null,
+                    stream.audioTrackId,
                     mutableListOf(stream)
                 )
             )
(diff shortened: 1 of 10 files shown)


app/src/main/java/com/github/libretube/util/DashHelper.kt
```diff
@@ -0,0 +1,176 @@
...
+import java.io.StringWriter
+import javax.xml.parsers.DocumentBuilder
+import javax.xml.parsers.DocumentBuilderFactory
+import javax.xml.transform.TransformerFactory
+import javax.xml.transform.dom.DOMSource
+import javax.xml.transform.stream.StreamResult
+
+// Based off of https://github.com/TeamPiped/Piped/blob/master/src/utils/DashUtils.js
+
+class DashHelper {
```
Bnyro, line 16: Maybe we could just use the Kotlin in-built `object` here instead of a class with a companion object? :)

FireMasterK: I have no clue how Kotlin works haha!

Bnyro: Just Java but in easier :p

PR #1878, opened 2022-11-17
Add option for runtime caching

### LINKED ISSUE #1838: limit thumbnail cache to runtime

I've made `@APPDATA@/files/coil` read only through root explorer to block thumbnail generation & the app became super snappy.

Thumbnail caching on storage media affects app performance a lot. Providing an option to limit thumb cache to runtime (memory), would be great. Because physical memory is always 10x faster than internal storage.

I'm sure thumb cache isn't necessary, as most users never revisit the older timeline multiple times. Even if they do, it wouldn't take much data & processing to reload those.

### Other details
I've researched a little bit about runtime mounts (tmpfs) on Android. But it's not possible although the Android file system is similar to Linux.

_No response_

### Acknowledgements

Bnyro:
closes #1838

### DIFF
--- a/app/src/main/java/com/github/libretube/util/ImageHelper.kt
+++ b/app/src/main/java/com/github/libretube/util/ImageHelper.kt
@@ -9,6 +9,7 @@ import android.widget.ImageView
 import coil.ImageLoader
 import coil.disk.DiskCache
 import coil.load
+import coil.request.CachePolicy
 import coil.request.ImageRequest
 import com.github.libretube.api.CronetHelper
 import com.github.libretube.constants.PreferenceKeys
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/github/libretube/util/ImageHelper.kt
```diff
@@ -25,17 +26,24 @@ object ImageHelper {
...
 
         imageLoader = ImageLoader.Builder(context)
             .callFactory(CronetHelper.callFactory)
-            .diskCache(diskCache)
+            .apply {
+                when (maxImageCacheSize) {
+                    "" -> {
+                        memoryCachePolicy(CachePolicy.DISABLED)
+                    }
+                    else -> diskCache(
```
FireMasterK, line 39: Would this mean caching is disabled by default?

Bnyro: Yes, is that a problem for the servers?
I'm always annoyed that the app size increases so much :/

FireMasterK: No, it's not a problem for servers but is a problem for app performance.
I think we should use both a memory and disk cache by default.

Let's say 20MB for disk, and 5MB of ram based caching.

Bnyro:
Oh, my bad, I disabled memory cache instead of disk cache haha

Bnyro:
Should be a oneliner though ...

PR #1899, opened 2022-11-18
Bookmark playlists

### LINKED ISSUE #1856: Bookmark playlists

When cloning playlists, they don't update when the original playlist gets updated. Hence it would be awesome to be able to set bookmarks to playlists.
This could also be implemented for videos but wouldn't update when the original video gets deleted or something similar to the watch history which is not a good solution in the end, hence the priority of the issue are playlists.

### Other details

_No response_

### Acknowledgements

Bnyro:
closes #1856


rozari0:
Where can I find my bookmarks other than home?

Bnyro:
> Where can I find my bookmarks other than home?
>
Only at home currently, I'll add it to the library tab too but it requires some changes first.

PR #1902, opened 2022-11-18
Implement proper handling of OTF videos.

FireMasterK:


app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
```diff
@@ -1193,13 +1193,18 @@ class PlayerFragment : BaseFragment(), OnlinePlayerOptions {
         if (!PreferenceHelper.getBoolean(PreferenceKeys.USE_HLS_OVER_DASH, false) &&
             streams.videoStreams.orEmpty().isNotEmpty()
         ) {
-            val manifest = DashHelper.createManifest(streams)
+            val uri = let {
```
Bnyro, line 1196: If the goal was to use dash if available that actually won't work like that :p

Bnyro: let will always take the last statement, it needs to be used differently :)

Bnyro: Or maybe I've just no idea how it works ...

FireMasterK: Well, it was actually:
```
if (dash != null) {
    return dash
}
```

IntelliJ / Android studio suggested that, and it worked like what I wrote!

PR #2040, opened 2022-11-25
Add support for swipe gesture

### LINKED ISSUE #1577: Option for playback gestures (volume & brightness)

Hi!
I'm already aware of:
- #60
- #62
- #71
- #381
- #545
- #980
- #1310
- #1541

but I think it's better to keep this issue open just in case someone else would like to implement one of the most requested feature.

Related discussion: #854

### Other details

Projects that already have it implemented (Possible reference): 
- https://github.com/videolan/vlc-android
- https://github.com/mpv-android/mpv-android
- https://github.com/TeamNewPipe/NewPipe
- https://github.com/SkyTubeTeam/SkyTube

### Acknowledgements

Kruna1Pate1:
Add support for swipe gestures to adjust brightness when in full-screen mode. By default, it is enabled which can be disabled from the player preference setting.

- Left screen swipe up/down to adjust brightness
- Right screen swipe up/down to adjust the volume
- Decrease brightness to < 0 will use system brightness

Closes #1577


app/src/main/java/com/github/libretube/util/PlayerGestureController.kt
```diff
@@ -0,0 +1,130 @@
...
+import android.os.Handler
+import android.os.Looper
+import android.os.SystemClock
+import android.view.GestureDetector
+import android.view.MotionEvent
+import android.view.View
+import com.github.libretube.ui.interfaces.PlayerGestureOptions
+import kotlin.math.abs
+
+class PlayerGestureController(context: Context, private val listner: PlayerGestureOptions) :
```
Bnyro, line 16: You probably meant `listener` here :)

app/src/main/res/xml/player_settings.xml
```diff
@@ -85,6 +85,13 @@
             app:key="picture_in_picture"
             app:title="@string/picture_in_picture" />
 
+        <SwitchPreferenceCompat
```
Bnyro, line 88: We should reorder the player preferences since the option gets kind of lost there as it's rather on the bottom while some unimportant settings like the default playback speed or the default resize mode are on the top. I'll probably change that later, just wanted to mention it here.

app/src/main/res/values/strings.xml
```diff
@@ -389,6 +389,11 @@
     <string name="local_playlists">Local playlists</string>
     <string name="not_enabled">Menu item not enabled!</string>
     <string name="select_other_start_tab">Please select an other start tab first!</string>
+    <string name="brightness">Brightness</string>
+    <string name="volume">Volume</string>
+    <string name="auto">Auto</string>
+    <string name="swipe_control">Swipe control</string>
```
Bnyro, line 395: I'd prefer `Swipe controls` as the English string here as your PR adds support for brightness as well as volume, we should change the string identifier name accordingly as well (`swipe_control` -> `swipe_controls`), same for the summary resource name.

app/src/main/java/com/github/libretube/ui/views/CustomExoPlayerView.kt
```diff
@@ -65,42 +76,68 @@ internal class CustomExoPlayerView(
...
-                eventPositionPercentageX > 0.6 -> forward()
-                else -> {
-                    player?.let { player ->
-                        if (player.isPlaying) {
-                            player.pause()
-                        } else {
-                            player.play()
-                        }
-                    }
+    private val playerGestureListner = object : PlayerGestureOptions {
```
Bnyro, line 79: Please let the `CustomExoPlayerView` extend the `PlayerOptionsInterface` and then override the actions for the `CustomExoPlayerView` instead of creating an extra variable for it to keep the code easier to overview :)

app/src/main/java/com/github/libretube/util/PlayerGestureController.kt
Kruna1Pate1: My bad :P

app/src/main/res/xml/player_settings.xml
Kruna1Pate1: Yes, some options can be moved below which are obvious and users generally don't use

app/src/main/res/values/strings.xml
Kruna1Pate1: My bad (English) again :P

app/src/main/java/com/github/libretube/ui/views/CustomExoPlayerView.kt
Kruna1Pate1: That sounds good. Implemented `PlayerGestureOptions` on `CustomExoPlayerView`.

Bnyro:
Thanks you! <3

swarnenduu:
I really wanted this gesture. Thanks

rahaaatul:
Unrelated, but can you do two fingers swipe up/down to increase/decrease video speed? And thank you for the feature we all wanted.

Kruna1Pate1:
> Unrelated, but can you do two fingers swipe up/down to increase/decrease video speed?

Not sure if It will work well with all gestures combined. But will try to do it.

PR #2071, opened 2022-11-26
Fix brightness inconsistency

### LINKED ISSUE #2046: Brightness doesn't return to normal after changing it from swipe control.
### Steps to reproduce

1. Open a video
2. Change it's brightness with swipe control
3. Close it and go back to LibreTubes home

### Expected behavior

Should return to default brightness that is set in device.

### Actual behavior

It remains the same. 

### LibreTube version

#2040

### Android version

Android 9

### Other details

_No response_

### Acknowledgements

Kruna1Pate1:
Brightness changes affect only when the player is in full-screen (landscape) mode. when orientation is changed to portrait, save brightness to shared preferences. It can be restored for later use.

Closes #2046


Bnyro (approved):
LGTM, thank you once again! <3

PR #2072, opened 2022-11-26
Pinch to zoom

### LINKED ISSUE #901: Pinch to zoom on Fullscreen

In Fullscreen mode pinching should switch between the normal viewing mode (Fit) and Zoom.

I would also be for removing the button which achieves the same. 

This would also mean to either:

a) remove the Fill mode completely,  I would be surprised if someone actually enjoys stretched videos

b) have a setting whether the user likes their videos zoomed or stretched.

### Other details

_No response_

### Acknowledgements

Bnyro:
closes #901

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/interfaces/PlayerGestureOptions.kt
+++ b/app/src/main/java/com/github/libretube/ui/interfaces/PlayerGestureOptions.kt
@@ -15,4 +15,8 @@ interface PlayerGestureOptions {
     fun onSwipeRightScreen(distanceY: Float)
 
     fun onSwipeEnd()
+
+    fun onZoom()
+
+    fun onMinimize()
 }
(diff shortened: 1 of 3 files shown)


Bnyro:
@Kruna1Pate1 I would appreciate if you could have a look at it (only if you have time of course).
Currently, when zooming in and out, it sometimes shows the player controller or interprets the gesture as a brightness/volume swipe (interesting lines are L46/47 in the `PlayerGestureController`). However, when changing the order/priority of the gesture detectors (so making the scale gesture listener check the event first and then only execute the normal gesture listener when the scale detector returned false), normal gestures and interactions with the player stop working.
If you have an idea feel free to chime in here, would be very appreciated :)

Kruna1Pate1:
I think we can try to let both gestures use the touch event. See this https://stackoverflow.com/a/47168465/10359348

Bnyro:
> I think we can try to let both gestures use the touch event. See this https://stackoverflow.com/a/47168465/10359348
>
Seems to work well now in most cases, thanks <3

PR #2080, opened 2022-11-26
Fix link opening issues

### LINKED ISSUE #1975: Incomplete YouTube Links
### Steps to reproduce

1. Open a video on LibreTube that has a YouTube video or playlist link on it's description. ⬇️
e.g. https://piped.video/watch?v=7zqNL-nzYfI

2. Click on that link

### Expected behavior

The link should open.

### Actual behavior

Video Link: The video stops(video screen turns black) and minimises to PIP mode.⬇️
Screenshot_20221122-104913
On clicking on PIP:⬇️
Screenshot_20221122-104949

Playlist Link: The video stops and minimises to PIP mode. Video player doesn't turn black.⬇️
Screenshot_20221122-105057
On clicking on PIP:⬇️
Screenshot_20221122-105104

### LibreTube version

0.8.0

### Android version

Android 12

### Other details

Supposed reason for the issue:
Screenshot_20221122-104354-625

Look in the above picture ⬆️
The video link has long url and is continued with '...' where the '...' is not hyperlinked and so url ends with 'mxE'. Same with playlist ⬇️
Screenshot_20221120-081543-905
But in YouTube website or app even the '...' is hyperlinked so we can open the complete link. Like YouTube website screenshot ⬇️
Screenshot_20221120-081717-372

But the links are completely hyperlinked in the comments section and so that works fine.⬇️
Screenshot_20221122-105805~2

These links also works completely fine in the piped.video website.

### Acknowledgements

Kruna1Pate1:
Fix opening playlist causing `NPE`.
Fix links are not actual hyperlinks. Which causes issues when opening ellipsed links.

There is still one issue, when clicking link it goes into PiP mode. What I have understood is,
When we click on link it trigers `MainActivity#onUserLeaveHint()` which then calls `PlayerFragment#onUserLeaveHint` which leads activity to go into PiP mode.
This problem won't occur when the player is paused because `PlayerFragment#shouldStartPiP()` check for it.

Possible solution could be,
- Pause player when the link is clicked. (Tried setting click listner but seems like Intent is send before the given event)
- Make PiP to fullscreen when the new Intent is received. (I don't think there is any straightforward way to do this)

Closes #1975

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/extensions/SetFormattedHtml.kt
+++ b/app/src/main/java/com/github/libretube/ui/extensions/SetFormattedHtml.kt
@@ -1,0 +1,10 @@
+package com.github.libretube.ui.extensions
+
+import android.text.util.Linkify
+import android.widget.TextView
+import androidx.core.text.HtmlCompat
+
+fun TextView.setFormattedHtml(text: String) {
+    Linkify.addLinks(this, Linkify.WEB_URLS)
+    this.text = HtmlCompat.fromHtml(text, HtmlCompat.FROM_HTML_MODE_LEGACY)
+}
--- a/app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
+++ b/app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
@@ -15,7 +15,6 @@ import android.os.Bundle
 import android.os.Handler
 import android.os.Looper
 import android.os.PowerManager
-import android.text.Html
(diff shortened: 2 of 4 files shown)


Bnyro:
Looks good to me, thank you!
I don't really know either what we could do about the PiP thing.
In theory the `MainActivity` is already forced to have only one instance at once which means the PiP window should "die" automatically when a link is being opened in LibreTube. If opening an external link I think it's fine if the app goes into PiP :)

PR #2083, opened 2022-11-26
Fix activity goes in PiP mode on link opening

Kruna1Pate1:
If the activity is running in PiP, then start the activity in front.
When the video is in `play` state sometimes it causes `kotlinx.coroutines.JobCancellationException` in `HomeFragment` but I think it's expected behavior since Job is closed forcefully.
Anyways, it's not affecting any behavior (other than making Toast which is handled now) so we can ignore it.


Bnyro:
Awesome, thank you :)

PR #2091, opened 2022-11-26
Fix plain text links and NPE when opening link from external source

Kruna1Pate1:
Some links are in plain text so that need to be considered.

When opening link from an external source (e.g. browser) `brightnessHelper` is not initialized which causes NPE


Bnyro:
Ty :)

PR #2100, opened 2022-11-27
Fix uploaded date accuracy

### LINKED ISSUE #1818: Duplicate Upload Date in Channel Page
### Steps to reproduce

- Open any channel, which gets video uploaded at least weekly 
- Scroll through videos 
- Many videos' upload date is same, while in actual, if you play the video then it'll show the correct date.

### Expected behavior

.

### Actual behavior

Seems, upload date gets cloned to videos.

### LibreTube version

0.7.0 & Latest Debug

### Android version

Android 12

### Other details

Screenshot_20221112-185115
Screenshot_20221112-185130
Screenshot_20221112-185152
Screenshot_20221112-185205
Screenshot_20221112-185214
Screenshot_20221112-185221

### Acknowledgements

Kruna1Pate1:
`video.uploaded` gives same value for some upload time range, I guess using `video.uploadedDate` which provides a response like `3 months ago` is better as discussed in issue. I didn't find any reason for using that `video.uploaded` and then parsing it,

Closes #1818

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/adapters/VideosAdapter.kt
+++ b/app/src/main/java/com/github/libretube/ui/adapters/VideosAdapter.kt
@@ -128,7 +128,7 @@ class VideosAdapter(
             videoInfo.text =
                 video.views.formatShort() + " " +
                 root.context.getString(R.string.views_placeholder) +
-                TextUtils.SEPARATOR + video.uploaded?.let { DateUtils.getRelativeTimeSpanString(it) }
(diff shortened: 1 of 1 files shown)


Bnyro:
Hmm, I'm not sure about that since the `video.uploadDate` is only an English string and not translated or adapted to the user's locale. But it might still be better than having inaccurate time spans.

FireMasterK:
Don't do this, that field is only for compatibility reasons. It's deprecated

PR #2134, opened 2022-11-28
Fix watch history deletion

### LINKED ISSUE #2127: Deleting video from watch history deletes wrong video
### Steps to reproduce

1. Go to Library -> Watch History
2. Delete any video, note which video u delete
3. Go back or any other page
4. Navigate back to Watch History you will notice the video deleted is above of that

For simplicity try deleting the first (top) video and continue from step 3. The video is not removed.
It is removed temporarily but not from db.

Also, when all video is removed `No history yet` page should be showing, which is missing.

### Expected behavior

Video on which delete action is performed should be deleted.

### Actual behavior

Video above it is deleted.

### LibreTube version

0.8.0 Debug

### Android version

Android 12

### Other details

_No response_

### Acknowledgements

Kruna1Pate1:
Fix threading issue, video was removed before the database query can use it for deletion which lead to remove wrong video.

Fix empty history view was not showing after removing all videos.

Closes #2127


Bnyro:
Ty :)

Issue #1811, opened 2022-11-11
Change play button icon into replay icon at end of video


terminal-cs:
### Describe your suggested feature

Title is self explanatory, makes more sense to have a replay icon instead.

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

PR #2171, opened 2022-12-02
Fix minimized player and repeat button

### LINKED ISSUE #2073: Sync miniplayer and player buttons

Sync player and miniplayer buttons, if you pause the video in the normal player, the miniplayer buttons should also be paused. 
Also change the play button at the end of a video to a replay button (like the miniplayer does)

### Other details

Screenshot_20221126-110515_LibreTube Debug

Screenshot_20221126-110520_LibreTube Debug

### Acknowledgements

Kruna1Pate1:
Synchronize mini-player buttons with playback state.

Use new image button for player repeat, because changing image resources of `exo_play_pause` will not work since It is frequently updated by [StyledPlayerControlView](https://github.com/google/ExoPlayer/blob/release-v2/library/ui/src/main/java/com/google/android/exoplayer2/ui/StyledPlayerControlView.java#L959).

Closes #2073

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
+++ b/app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
@@ -315,6 +315,12 @@ class PlayerFragment : BaseFragment(), OnlinePlayerOptions {
             BackgroundHelper.stopBackgroundPlay(requireContext())
         }
 
+        playerBinding.repeatBTN.setOnClickListener {
+            // Restart video if finished
+            exoPlayer.seekTo(0)
+            exoPlayer.play()
+        }
+
         binding.playImageView.setOnClickListener {
             if (!exoPlayer.isPlaying) {
                 // start or go on playing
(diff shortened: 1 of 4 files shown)


Bnyro:
Also closes #1811 🎉

Bnyro:
In theory, couldn't we just replace the `exo_play_pause` button by using the new repeat button for them as well and just always change the image resource according to the playback state? That might give us more freedom on controlling it's behavior in general :)

Bnyro:
And it would look cleaner than having two different buttons that we're always hiding and showing which might potentially be visible both at the same time when something goes wrong :)

Kruna1Pate1:
That's a good idea, because current implementation looks more like a workaround :)
let me try this.

Kruna1Pate1:
Made the changes looks fine now.
We can also get rid of 
```xml
    <drawable name="exo_styled_controls_play">@drawable/ic_play</drawable>
    <drawable name="exo_styled_controls_pause">@drawable/ic_pause</drawable>
```
but not sure if it is used somewhere else also.

Bnyro:
> Made the changes looks fine now.
> We can also get rid of 
> ```xml
>     <drawable name="exo_styled_controls_play">@drawable/ic_play</drawable>
>     <drawable name="exo_styled_controls_pause">@drawable/ic_pause</drawable>
> ```
> but not sure if it is used somewhere else also.
>
They are not used anywhere else so they can be removed, yes!

Kruna1Pate1:
Removed them. Looks fine now.

Bnyro:
Awesome, I'll quickly test it and merge it if I find no serious issues :)

Bnyro:
Works really well, thank you :tada:

PR #2226, opened 2022-12-04
Update Screenshots.

rozari0:


Bnyro:
Thank you! Indeed a good choice in terms of the search queries and visited channels ;)

rozari0:
> Thank you! Indeed a good choice in terms of the search queries and visited channels ;)

That's not actually mine, stolen from old screenshots. : )

Bnyro:
> > Thank you! Indeed a good choice in terms of the search queries and visited channels ;)
> 
> That's not actually mine, stolen from old screenshots. : )
>
I know, that's why I mentioned it 😂

PR #2264, opened 2022-12-06
Show/Hide exact view count with description toggle

### LINKED ISSUE #2138: Add an option to show exact view count

An option that makes the view count change from e.g. 1M views to 1,234,567 views

### Other details

Probably should be enabled by default or even be the only way of showing the data. Exact view count is already present in Piped

### Acknowledgements

Kruna1Pate1:
Show the exact view count when the description is open. Otherwise, show formated short count.

Closes #2138

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
+++ b/app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
@@ -463,10 +463,20 @@ class PlayerFragment : BaseFragment(), OnlinePlayerOptions {
             // hide the description and chapters
             binding.playerDescriptionArrow.animate().rotation(0F).setDuration(250).start()
             binding.descLinLayout.visibility = View.GONE
+
+            // show formated short view count
+            val viewInfo = context?.getString(R.string.views, streams.views.formatShort()) +
+                if (!isLive) TextUtils.SEPARATOR + streams.uploadDate else ""
+            binding.playerViewsInfo.text = viewInfo
         } else {
             // show the description and chapters
             binding.playerDescriptionArrow.animate().rotation(180F).setDuration(250).start()
             binding.descLinLayout.visibility = View.VISIBLE
+
+            // show exact view count
+            val viewInfo = context?.getString(R.string.views, String.format("%,d", streams.views)) +
+                if (!isLive) TextUtils.SEPARATOR + streams.uploadDate else ""
+            binding.playerViewsInfo.text = viewInfo
         }
         if (this::chapters.isInitialized && chapters.isNotEmpty()) {
             val chapterIndex = getCurrentChapterIndex()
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
```diff
@@ -463,10 +463,20 @@ class PlayerFragment : BaseFragment(), OnlinePlayerOptions {
             // hide the description and chapters
             binding.playerDescriptionArrow.animate().rotation(0F).setDuration(250).start()
             binding.descLinLayout.visibility = View.GONE
+
+            // show formated short view count
+            val viewInfo = context?.getString(R.string.views, streams.views.formatShort()) +
+                if (!isLive) TextUtils.SEPARATOR + streams.uploadDate else ""
```
Bnyro, line 469: Instead of using that line twice we might just want to save that text to a variable before the if/else condition and reuse it in the if and the else clause. Apart from that `getString()` should be sufficient, using `context?#getString()!!` shouldn't be needed as fragments implement their own method for it.

Kruna1Pate1: not sure then how to use `streams.views.formatShort()` and `String.format("%,d", streams.views)` in if else than.

> Apart from that getString() should be sufficient, using context?#getString()!! shouldn't be needed as fragments implement their own method for it.

changed that 👍🏼

Bnyro: > not sure then how to use `streams.views.formatShort()` and `String.format("%,d", streams.views)` in if else than.
>
I only meant saving `if (!isLive) TextUtils.SEPARATOR + streams.uploadDate else ""` to a variable before (just to avoid duplicating it) and later adding it to the end of the formatted views string.
The string formatting has to be done each time individually as it's already done in your PR.

Kruna1Pate1: ohk, got it.

app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
```diff
@@ -459,14 +459,23 @@ class PlayerFragment : BaseFragment(), OnlinePlayerOptions {
...
     private fun toggleDescription() {
+        var viewInfo = if (!isLive) TextUtils.SEPARATOR + streams.uploadDate else ""
         if (binding.descLinLayout.isVisible) {
             // hide the description and chapters
             binding.playerDescriptionArrow.animate().rotation(0F).setDuration(250).start()
             binding.descLinLayout.visibility = View.GONE
+
+            // show formated short view count
+            viewInfo = getString(R.string.views, streams.views.formatShort()) + viewInfo
+            binding.playerViewsInfo.text = viewInfo
```
Bnyro, line 470: We could even move that line below the if/else condition because it's used in both the if and the else condition now :)

Kruna1Pate1: Yep, didn't noticed :)

Bnyro:
Thanks! :tada:

PR #2343, opened 2022-12-10
Change Language Strings

matchboxbananasynergy:
This PR changes the list of languages to use their spelling instead of having them all in English.

This list was used:

https://en.wikipedia.org/wiki/List_of_language_names

Someone should double-check it to make sure there are no mistakes.

### DIFF
--- a/app/src/main/res/values/array.xml
+++ b/app/src/main/res/values/array.xml
@@ -74,7 +74,7 @@
         <item>Euskara</item>
         <item>বাংলা</item>
         <item>Català</item>
-        <item>汉语</item>
+        <item>简体中文</item>
         <item>古文</item>
         <item>Czech</item>
         <item>Dansk</item>
--- a/app/src/main/res/values/array.xml
+++ b/app/src/main/res/values/array.xml
@@ -76,7 +76,7 @@
         <item>Català</item>
         <item>简体中文</item>
         <item>繁體中文</item>
-        <item>Czech</item>
+        <item>čeština</item>
         <item>Dansk</item>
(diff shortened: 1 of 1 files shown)


app/src/main/res/values/array.xml
```diff
@@ -69,47 +69,47 @@
...
-        <item>Bengali</item>
-        <item>Catalan</item>
-        <item>Chinese (simplified)</item>
-        <item>Chinese (traditional)</item>
+        <item>العربية</item>
+        <item>Azərbaycan dili</item>
+        <item>Euskara</item>
+        <item>বাংলা</item>
+        <item>Català</item>
+        <item>汉语</item>
```
raymond-nee, line 77:
```suggestion
        <item>简体中文</item>
```
For the Chinese (Simplified), The term “简体中文” (which is a direct translation for "Simplified Chinese" ) is more commonly used.

app/src/main/res/values/array.xml
```diff
@@ -69,47 +69,47 @@
...
+        <item>古文</item>
```
raymond-nee, line 78:
```suggestion
        <item>繁體中文</item>
```
For Chinese (Traditional), it may use “繁體中文“ (direct translation is "Complex Chinese").

Bnyro:
Does anyone know whether čeština or český is the right term for Czech?

xIUPITERx:
https://www.italki.com/en/post/question-211686
If they are right it's "čeština".

Bnyro:
Thanks!

PR #2469, opened 2022-12-21
New custom downloader

Kruna1Pate1:
Implemented a new downloader using `HttpURLConnection` with `Okio`.

- [x] Progress of download
- [x] Pause/Resume download
- [x] Notification with progress and play/pause action
- [x] Download subtitle
- [x] Update expired link (probably after 6 hours). (format and quality are saved in db, should be easy to implement)
- [ ] Automatic retry on failiur
- [x] Limit concurrent download
- [ ] Load subtitle in offline player
- [x] Save file with it's title name and only filter reserved chars

Removed view info (probably not needed as it get frequently updated anyways) from the download fragment to get some space for showing download size :)

Currently using bound service and flow to update `DownloadFragment` progress.
Let me know if something should be changed or not working.

Some unused code still need to be removed like notification channel etc.

Not tried with large fie but should work similarly


FireMasterK:
Why not use cronet instead for the connections? 🤔

Kruna1Pate1:
> Why not use cronet instead for the connections? 

I don't know much about cronet. Will try to change implementation.

Kruna1Pate1:
I have completed most of the changes required.
I think using the downloaded subtitle functionality would be better in a separate PR.
Retry on connection timeout is implemented. But an error when downloading a file would still need to start downloading manually. I think it's not a big issue and can be added eventually.
Let me know if any changes are needed.

Bnyro:
Could you please run `ktlint --android -F` in the root directory of the project at first? I'll go through the code and leave some comments (if needed) later today :)

Kruna1Pate1:
> Could you please run ktlint --android -F in the root directory of the project at first?

I did that before every commit. But no idea why checks are failed.

Bnyro:
Hmm, are you using the latest version of ktlint?

Kruna1Pate1:
> Hmm, are you using the latest version of ktlint?

No i didn't updated for while. May be that's issue. Will check.

Kruna1Pate1:
Yep, version issue :) now looks fine.

Bnyro:
We should probably also resolve https://github.com/libre-tube/LibreTube/pull/2469/checks?check_run_id=10294742980 before merging, I don't know how tbh

app/src/main/java/com/github/libretube/constants/IntentData.kt
```diff
@@ -10,4 +10,10 @@ object IntentData {
     const val fileName = "fileName"
     const val openQueueOnce = "openQueue"
     const val playlistType = "playlistType"
+    const val videoFormat = "videoFormate"
+    const val videoQuality = "videoQuality"
+    const val audioFormat = "audioFormate"
```
Bnyro, line 15: *"audioFormat"

app/src/main/java/com/github/libretube/constants/IntentData.kt
```diff
@@ -10,4 +10,10 @@ object IntentData {
     const val fileName = "fileName"
     const val openQueueOnce = "openQueue"
     const val playlistType = "playlistType"
+    const val videoFormat = "videoFormate"
```
Bnyro, line 13: *"videoFormat"

app/src/main/java/com/github/libretube/db/dao/DownloadDao.kt
```diff
@@ -0,0 +1,51 @@
...
+    @Insert(onConflict = OnConflictStrategy.REPLACE)
+    fun insertDownloadItem(downloadItem: DownloadItem): Long
+
+    @Update(onConflict = OnConflictStrategy.REPLACE)
+    fun updateDownload(download: Download)
+
+    @Update(onConflict = OnConflictStrategy.REPLACE)
+    fun updateDownloadItem(downloadItem: DownloadItem)
+
+    @Transaction
```
Bnyro, line 42: Why are you using a transaction here?

app/src/main/java/com/github/libretube/enums/FileType.kt
```diff
@@ -0,0 +1,7 @@
+package com.github.libretube.enums
+
+enum class FileType {
```
Bnyro, line 3: Maybe that could somewhat be combined with the enum `DownloadType` we already have?

app/src/main/java/com/github/libretube/receivers/NotificationReceiver.kt
```diff
@@ -0,0 +1,33 @@
...
+        if (intent?.action == null) return
+
+        val serviceIntent = Intent(context, DownloadService::class.java)
+        serviceIntent.action = intent.action
+
+        val id = intent.getIntExtra("id", -1)
+        if (id == -1) return
+        serviceIntent.putExtra("id", id)
+
+        if (Build.VERSION.SDK_INT > Build.VERSION_CODES.O) {
```
Bnyro, line 20: Since we use that multiple times within the whole code we could write a function/compat for that (starting foreground services if supported).

app/src/main/res/values/strings.xml
```diff
@@ -182,9 +182,17 @@
     <string name="playerVideoFormat">Video format for player</string>
     <string name="no_audio">No audio</string>
     <string name="no_video">No video</string>
+    <string name="no_subtitle">No subtitle</string>
     <string name="audio">Audio</string>
     <string name="video">Video</string>
     <string name="downloading">Downloading…</string>
+    <string name="download_paused">Download paused</string>
+    <string name="download_completed">Download completed</string>
```
Bnyro, line 190: We already have a `downloadsucceeded` string, so the `download_completed` string here shouldn't be needed and instead replaced with the `downloadsucceeded` string resource.

app/src/main/java/com/github/libretube/services/DownloadService.kt
```diff
@@ -1,182 +1,524 @@
...
+                            uploadDate = streams.uploadDate,
+                            uploader = streams.uploader ?: ""
+                        )
+                    )
+                }
+                streams.thumbnailUrl?.let { url ->
+                    ImageHelper.downloadImage(
+                        this@DownloadService,
+                        url,
+                        File(
```
Bnyro, line 120: We should create a function that returns the download directory only based on one argument like DownloadHelper#THUMBNAIL_DIR  in order to not repeat that code for creating a file multiple times.

app/src/main/java/com/github/libretube/services/DownloadService.kt
```diff
@@ -1,182 +1,524 @@
...
-    override fun onBind(intent: Intent?): IBinder? {
-        TODO("Not yet implemented")
+    /**
+     * Initiate download [Job] using [DownloadItem] by creating file according to [FileType]
+     * for the requested file.
+     */
+    private fun start(item: DownloadItem) {
+        val file: File = when (item.type) {
+            FileType.AUDIO -> {
+                val audioDownloadDir = DownloadHelper.getDownloadDir(
```
Bnyro, line 155: That (and the two following statements for videos and subtitles) should be using the above mentioned function as well.

app/src/main/java/com/github/libretube/services/DownloadService.kt
```diff
@@ -1,182 +1,524 @@
...
-                    "[${getString(R.string.video)}] $videoName",
-                    getString(R.string.downloading),
-                    videoUrl,
-                    Uri.fromFile(
-                        File(videoDownloadDir, videoName)
-                    )
-                )
+            // Set start range where last downloading was held.
+            val con = CronetHelper.getCronetEngine().openConnection(url) as HttpURLConnection
+            con.requestMethod = "GET"
```
Bnyro, line 212: `con.apply {}` could be used here instead.

app/src/main/java/com/github/libretube/services/DownloadService.kt
```diff
@@ -1,182 +1,524 @@
...
-            unregisterReceiver(onDownloadComplete)
-        } catch (e: Exception) {
-            e.printStackTrace()
+    private fun getPauseAction(id: Int): NotificationCompat.Action {
+        val intent = Intent(this, NotificationReceiver::class.java)
+
+        intent.action = ACTION_DOWNLOAD_PAUSE
+        intent.putExtra("id", id)
+
+        val flags = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
```
Bnyro, line 486: That could be moved into a function called `getNotificationFlags()` as its used for the other  notification at `getResumeAction()` too.

app/src/main/java/com/github/libretube/enums/FileType.kt
Kruna1Pate1: `DownloadType` is not used anywhere so we can remove it or rename `FileType` to `DownloadType`.

app/src/main/java/com/github/libretube/db/dao/DownloadDao.kt
Kruna1Pate1: Deleting records from `Download` table needs to delete all related entries from `DownloadItem` also, so I think it is required.

app/src/main/java/com/github/libretube/enums/FileType.kt
Bnyro: I'd say we rename `FileType` to `DownloadType` because that's what it's currently being used for.

app/src/main/java/com/github/libretube/db/dao/DownloadDao.kt
Bnyro: Makes sense, thanks for the explanation :)

Kruna1Pate1:
> We should probably also resolve https://github.com/libre-tube/LibreTube/pull/2469/checks?check_run_id=10294742980 before merging, I don't know how tbh

According to [recommendation](https://codeql.github.com/codeql-query-help/go/go-path-injection/#recommendation), validate user input and disallow path traversal char. That is already done [here](https://github.com/libre-tube/LibreTube/pull/2469/files#diff-868e7632feaab8b5f8aa07fce48d7b1130c44fd826703a58fff9fe3327d23c33R45-R63).

app/src/main/java/com/github/libretube/services/DownloadService.kt
Kruna1Pate1: Not sure wdym. current `getDownloadDir` takes two args that are required.

Bnyro:
You're right, but that still doesn't stop CodeQL from complaining unfortunately. Maybe we can just apply the filename Regex to the filename where CodeQL complained (although it's not needed for the code), only to try making CodeQL happy again?

Bnyro:
I want to avoid having broken CI for all future PRs. Otherwise we could try to set a CodeQL rule to ignore that case but I don't know if that's possible.

app/src/main/java/com/github/libretube/services/DownloadService.kt
Bnyro: Always when calling
```kotlin
File(DownloadHelper.getDownloadDir(), fileName)
```
here in the DownloadService, the `fileName` and the `context` parameter are always the name. Hence I suggest to create a function that only takes the `path` parameter of `DownloadHelper#getDownloadDir` and creates a file as mentioned above from it. Hope that's understandable, if not feel free to ask :)

Kruna1Pate1:
> I want to avoid having broken CI for all future PRs. Otherwise we could try to set a CodeQL rule to ignore that case but I don't know if that's possible.

Agree, will look if we can find some proper way for this. But, i am not available for some days so i will take a look after some days.

app/src/main/java/com/github/libretube/services/DownloadService.kt
Kruna1Pate1: Alright, i will try that.

FireMasterK:
https://github.com/github/codeql/issues/9298#issuecomment-1136938130

Apparently, we can dismiss warnings through some sort of UI :thinking:

Bnyro:
We're going to merge that after the next release as we should probably allocate some time for testing after we merged it.

xIUPITERx:
Could we merge the pull request so that we still have as much time as possible to find errors in the debug versions?
Or should we wait for 11.0?

Bnyro:
We still need to resolve CodeQL and some merge conflicts as well as some of my comments. Then we're good to merge it.

Bnyro:
Next version will probably be 0.11.0 though, updates will arrive with more time in between (like 3 to 4 weeks now) instead of once per two weeks.

Bnyro:
Nvm, I'll merge it now and then try to resolve stuff before the release.

Bnyro:
Btw, awesome work on that @Kruna1Pate1, downloads really feel a lot faster and more consistent now!

PR #2528, opened 2022-12-28
Update AndroidManifest.xml

Roguo-Kato:
Due to the OS, writing permission is required before Pie.


Bnyro:
In fact I wanted to avoid requiring any storage permissions, are you aware of any directory that the app can access without permissions below Android Pie?

Bnyro:
Then we could just use that directory for those Android versions instead of requiring storage permissions.

Roguo-Kato:
I think if it use the file picker and DocumentFile, it will haven't request WRITE_EXTERNAL_STORAGE. But DownloadManager requires WRITE_EXTERNAL_STORAGE.

Issue #2059, opened 2022-11-26
Can't download video/audio on android 5.1


Nidal-Bakir:
### Steps to reproduce

Download any video/audio will reproduce the bug on `Android 5.1`

### Expected behavior

The video been downloaded.

### Actual behavior

The app crashes.

### LibreTube version

0.8.0

### Android version

Android 5.1

### Other details

> java.lang.RuntimeException: Unable to start service com.github.libretube.services.DownloadService@21f5ddfc with Intent { cmp=com.github.libretube/.services.DownloadService (has extras) }: java.lang.SecurityException: need WRITE_EXTERNAL_STORAGE permission to use DESTINATION_FILE_URI: uid 10121 does not have android.permission.WRITE_EXTERNAL_STORAGE.
> 	at android.app.ActivityThread.handleServiceArgs(ActivityThread.java:3975)
> 	at android.app.ActivityThread.access$2300(ActivityThread.java:218)
> 	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1842)
> 	at android.os.Handler.dispatchMessage(Handler.java:102)
> 	at android.os.Looper.loop(Looper.java:145)
> 	at android.app.ActivityThread.main(ActivityThread.java:7007)
> 	at java.lang.reflect.Method.invoke(Native Method)
> 	at java.lang.reflect.Method.invoke(Method.java:372)
> 	at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1404)
> 	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1199)
> Caused by: java.lang.SecurityException: need WRITE_EXTERNAL_STORAGE permission to use DESTINATION_FILE_URI: uid 10121 does not have android.permission.WRITE_EXTERNAL_STORAGE.
> 	at android.os.Parcel.readException(Parcel.java:1546)
> 	at android.database.DatabaseUtils.readExceptionFromParcel(DatabaseUtils.java:185)
> 	at android.database.DatabaseUtils.readExceptionFromParcel(DatabaseUtils.java:137)
> 	at android.content.ContentProviderProxy.insert(ContentProviderNative.java:476)
> 	at android.content.ContentResolver.insert(ContentResolver.java:1260)
> 	at android.app.DownloadManager.enqueue(DownloadManager.java:1336)
> 	at com.github.libretube.services.DownloadService.downloadManagerRequest(DownloadService.kt:53)
[...]

PR #2537, opened 2022-12-28
File permission on Pie or below

Roguo-Kato:


Roguo-Kato:
Sorry, I made a mistake in using github.

> In fact I wanted to avoid requiring any storage permissions, are you aware of any directory that the app can access without permissions below Android Pie?

> Then we could just use that directory for those Android versions instead of requiring storage permissions.

I think if it use the file picker and DocumentFile, it will haven't request WRITE_EXTERNAL_STORAGE. But DownloadManager requires WRITE_EXTERNAL_STORAGE.

Bnyro:
Independent of that, please only create one PR and not two separated ones for the same thing.
Apart from that I still don't really want to add storage permissions to the app and would prefer finding a different solution for the issue.

Roguo-Kato:
Thank you for your advice. I use Android 12 and 5.0, and there are some other suggestions, but is it annoying to make PR each time?

Bnyro:
Your two PRs do are targeting the same issue so one would be preferred, apart from it such a change should have discussed in a bug report first.

Roguo-Kato:
I realized that I was making patch1 and patch2. I repaired.
>it such a change should have discussed in a bug report first.
I had watched it, but the discussion were closed without fixing.
(https://github.com/libre-tube/LibreTube/issues/2059)
I had understood that DownloadManager.enqueue was the cause, so I just wanted to make a suggestion for a dream-like app.

PR #2596, opened 2023-01-05
Only load 10 videos at once in private playlists

### LINKED ISSUE #2289: Having too much videos in saved playlist causes a crash
### Steps to reproduce

1. Have a Piped playlist with ~500+ videos (IMPORTANT: Not a bookmarked playlist)
2. Try to open it from an app

### Expected behavior

After some time playlist should load without problems

### Actual behavior

Loading takes too much time, app then crashes.

### LibreTube version

0.9.0

### Android version

Android 12

### Other details

I believe it's happening because app is trying to load all videos in a playlist at once. It doesn't matter when there are 100 videos present, but when it's 1000, everything breaks. I think that loading only a part of playlist at a time would help (like bookmarked and public playlists do that)

### Acknowledgements

Bnyro:
closes #2289


Bnyro:
How is that related to the PR?

PR #2659, opened 2023-01-11
Remove multidex.

Isira-Seneviratne:
Remove the multidex library, as it is not needed on Android 5.0 and later.


Bnyro:
Didn't know that, thank you!

PR #2703, opened 2023-01-15
Simplify LocaleHelper methods.

Isira-Seneviratne:
Simplify the `LocaleHelper` methods using Kotlin's extensions and property syntax.

### DIFF
--- a/app/src/main/java/com/github/libretube/util/LocaleHelper.kt
+++ b/app/src/main/java/com/github/libretube/util/LocaleHelper.kt
@@ -17,7 +17,7 @@ object LocaleHelper {
         val languageName = PreferenceHelper.getString(PreferenceKeys.LANGUAGE, "sys")
         val locale = when {
             languageName == "sys" -> Locale.getDefault()
-            languageName.contains("-") == true -> {
+            languageName.contains("-") -> {
                 val languageParts = languageName.split("-")
                 Locale(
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/github/libretube/util/LocaleHelper.kt
```diff
@@ -39,101 +41,50 @@ object LocaleHelper {
...
-    private fun detectSIMCountry(context: Context): String? {
-        try {
-            val telephonyManager =
-                context.getSystemService(Context.TELEPHONY_SERVICE) as TelephonyManager
-            return telephonyManager.simCountryIso
-        } catch (e: Exception) {
-            e.printStackTrace()
-        }
-        return null
+    private fun detectSIMCountry(context: Context): String {
```
Bnyro, line 57: Instead of returning empty strings here, maybe we could let it return null instead and then do
```kotlin
private fun getDetectedCountry(context: Context): String {
     return detectSIMCountry(context) ?: detectNetworkCountry(context) ?: detectLocaleCountry(context) ?: "UK"
}
```

Isira-Seneviratne: The methods return empty strings if the country is not available, so those would have to be handled.

Bnyro:
Awesome, thank you!

PR #2717, opened 2023-01-16
Fixed #2670 : Timestamp click behaviour in the description.

### LINKED ISSUE #2670: Strange Behavior When Clicking on Timestamps
### Steps to reproduce

1. Open a video that has timestamps for different parts. (like [this](https://piped.video/watch?v=l6fVZ0GRbhQ)) 
2.  Click on a timestamp from description 

### Expected behavior

Smooth transition when attempting to play the video from a timestamp

### Actual behavior

https://user-images.githubusercontent.com/50032603/212070703-42c8037e-5a33-4a33-90cf-191fb517f7e8.mp4

### LibreTube version

0.10. 0

### Android version

Android 12

### Other details

_No response_

### Acknowledgements

### LINKED ISSUE #2723: Player disappears when entered in app from media notification
### Steps to reproduce

1. Play any video
2. Back to home or switch app
3. Tap on player notification 

### Expected behavior

App should open in either player screen or with mini player 

### Actual behavior

Player screen completely disappears 

### LibreTube version

Latest debug

### Android version

Android 12

### Other details

https://user-images.githubusercontent.com/88180887/212749007-9802fe17-837c-4b5f-9f8c-b9ba89d0497c.mp4

### Acknowledgements

faisalcodes:
- Now clicking on a timestamp in the description plays the video at the specific time.
- Youtube video links are open within the player instead of working like a normal link.

closes #2670 
closes #2723


Bnyro:
Could you please run `ktlint --android -F` in the root directory of the project and push the changes to your branch? Thanks!

faisalcodes:
@Bnyro I have updated the branch.

app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
```diff
@@ -1040,6 +1036,119 @@ class PlayerFragment : BaseFragment(), OnlinePlayerOptions {
...
+                                    // get the time in seconds
+                                    val timeInSeconds = getTimeInMillis(time)
+                                    if (timeInSeconds != -1L) {
+                                        // seek to the time
+                                        exoPlayer.seekTo(timeInSeconds)
+                                    }
+                                }
+                                // youtube link without time
+                                // open new player
+                                playNextVideo(videoId)
```
Bnyro, line 1070: That apparently is still being called even though the player previously seeked to another position in line 1065, so the video will be reloaded although we already figured out it's only a timestamp which is not what we want to achieve.

app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
```diff
@@ -1040,6 +1036,119 @@ class PlayerFragment : BaseFragment(), OnlinePlayerOptions {
...
+        if (description.contains("<") && description.contains(">")) {
+            descTextView.movementMethod = LinkMovementMethod.getInstance()
+            descTextView.text = HtmlParser.createSpannedText(
+                description,
+                LinkHandler { link ->
+                    // check if the link is a youtube link
+                    if (link.contains("youtube.com") || link.contains("youtu.be")) {
+                        // check if the link is a video link
+                        val videoId = getVideoIdIfVideoLink(link)
+                        if (!videoId.isNullOrEmpty()) {
```
Bnyro, line 1054: Instead of using lots of if conditions here, we should create a separate function which for example directly opens a new link and returns when the `videoId` is null or empty. That could also be partly applied for the following if conditions, just to keep the code easier to understand.

app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
```diff
@@ -1040,6 +1036,119 @@ class PlayerFragment : BaseFragment(), OnlinePlayerOptions {
...
+                }
+            )
+        } else {
+            // Links can be present as plain text
+            descTextView.autoLinkMask = Linkify.WEB_URLS
+            descTextView.text = description
+        }
+    }
+
+    private fun getVideoIdIfVideoLink(link: String): String? {
```
Bnyro, line 1097: That's not needed, instead we should first parse the URL with `Uri#parse` and then use `uri.getQueryParameter` instead of manually splitting strings

app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
```diff
@@ -1040,6 +1036,119 @@ class PlayerFragment : BaseFragment(), OnlinePlayerOptions {
...
+        if (link.contains("youtube.com")) {
+            // check if the link is a channel link
+            val videoId = link.substringAfter("v=").substringBefore("&")
+            if (videoId.isNotEmpty() && videoId != link) {
+                return videoId
+            }
+        } else if (link.contains("youtu.be")) {
```
Bnyro, line 1104: Instead of splitting strings, we should use the path from the Uri generated above

app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
```diff
@@ -1040,6 +1036,119 @@ class PlayerFragment : BaseFragment(), OnlinePlayerOptions {
...
+        if (time.contains("m")) {
+            val longOrNull = time.substringBefore("m").substringAfter("s").toLongOrNull()
+            if (longOrNull != null) {
+                timeInSeconds += longOrNull * 60
+                found = true
+            }
+        }
+        // check if the time is in hours
+        if (time.contains("h")) {
+            val longOrNull = time.substringBefore("h").substringAfter("m").toLongOrNull()
```
Bnyro, line 1134: Instead, please use something like ```kotlin
time.substringBefore("h").substringAfter("m").toLongOrNull()?.let {
    timeInSeconds += longOrNull * 60 * 60
    found = true
}
```
here

File: app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt [hunk 6]
```diff
@@ -1040,6 +1036,119 @@ class PlayerFragment : BaseFragment(), OnlinePlayerOptions {
...
+            if (longOrNull != null) {
+                timeInSeconds += longOrNull * 60 * 60
+                found = true
+            }
+        }
+
+        if (!found) {
+            val longOrNull = time.toLongOrNull()
```
Bnyro, line 1142: Same as the comment a few lines above

app/src/main/java/com/github/libretube/util/LinkHandler.kt
```diff
@@ -0,0 +1,54 @@
...
+class LinkHandler(private val clickCallback: ((String) -> Unit)?) : HtmlParser.TagHandler {
+    private var linkTagStartIndex = -1
+    private var link: String? = null
+    override fun handleTag(
+        opening: Boolean,
+        tag: String?,
+        output: Editable?,
+        attributes: Attributes?
+    ): Boolean {
+        if (output != null) {
```
Bnyro, line 20: These if conditions should be replaced with with some return statements instead.

app/src/main/java/com/github/libretube/util/HtmlParser.kt
```diff
@@ -0,0 +1,134 @@
...
+import android.text.Html
+import android.text.Spanned
+import androidx.core.text.HtmlCompat
+import org.xml.sax.Attributes
+import org.xml.sax.ContentHandler
+import org.xml.sax.Locator
+import org.xml.sax.SAXException
+import org.xml.sax.XMLReader
+
+class HtmlParser private constructor(private val handler: TagHandler) :
```
Bnyro, line 13: Why using a private constructor here?

Bnyro:
I'm not really a fan of that `HTMLParser` as it's quite large in code size although it doesn't really do much, maybe we can find some better way to do that or remove some unused/unneeded parts of it.

app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
faisalcodes: I did this because the link could also be a channel or playlist link. In future we can open channel or playlist within the app instead of opening an external intent.

app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
faisalcodes: Sorry, here I forgot an else condition. I must fix this by merging both  `playNextVideo(videoId)`.

app/src/main/java/com/github/libretube/util/HtmlParser.kt
faisalcodes: To get what we need, we do need to override the link handling. This is the only way I find to be useful to handle timestamps because we could not find any callback from the default implementations.

app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
faisalcodes: Fortunately video was not reloading because in `playNextVideo` it checks for difference.

app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
faisalcodes: Pardons but i thought this would be more readable.

app/src/main/java/com/github/libretube/util/HtmlParser.kt
faisalcodes: I am making it short.

Bnyro:
Works awesome, thank you!

PR #2726, opened 2023-01-16
Better compability on Android 11 in full screen

demirdegerli:
Implemented libs with compat ones. Also I don't know Kotlin so there might be bugs or type errors. Please edit the code if you find a mistake.

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/activities/MainActivity.kt
+++ b/app/src/main/java/com/github/libretube/ui/activities/MainActivity.kt
@@ -15,6 +15,9 @@ import android.view.ViewGroup
 import android.view.WindowInsets
 import android.view.WindowInsetsController
 import android.view.WindowManager
+import android.view.WindowCompat
+import android.view.WindowInsetsControllerCompat
+import android.view.WindowInsetsCompat
 import android.widget.ScrollView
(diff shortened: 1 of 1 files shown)


Bnyro:
As the PR adds the use of a compat, we should probably get rid of the if condition and the deprecated method below.
Did you test whether it fixes the issue?

demirdegerli:
I tested in a blank Kotlin project, built it and it worked. Because of I couldn't build LibreTube.

demirdegerli:
Also I didn't remove the non-compat ones in case of something breaks.

Bnyro:
The CI is broken too so no wonder it's not building for you. Apart from that I wonder how using the compat should have changed something here really because according to what I found the methods called would be still the same as before the change on Android 9 and up

demirdegerli:
First tap shows the system UI, second tap shows the OSD. It's like there is something between the front and LibreTube UI. An invisible divider dissolves at first tap.
https://user-images.githubusercontent.com/67432156/212762346-d448fbfb-44b2-4330-885d-2694beeba41b.mp4

Bnyro:
I've been writing a more elegant solution now (#2727) which should do what you initially intended to, feel free to test the debug builds!

PR #2745, opened 2023-01-17
strings.xml

Ilhamiii:
Use a more correct word.


Bnyro:
Watch history is the proper term for that

PR #2749, opened 2023-01-18
Add PendingIntentCompat helper methods.

Isira-Seneviratne:
Add helper methods to `PendingIntentCompat` to avoid issues on API levels < 23.

### DIFF
--- a/app/src/main/java/com/github/libretube/compat/PendingIntentCompat.kt
+++ b/app/src/main/java/com/github/libretube/compat/PendingIntentCompat.kt
@@ -1,20 +1,30 @@
 package com.github.libretube.compat
 
 import android.app.PendingIntent
+import android.content.Context
+import android.content.Intent
 import android.os.Build
 
+// TODO: Use AndroidX's PendingIntentCompat class instead once it becomes available.
 object PendingIntentCompat {
-    val updateCurrentFlags: Int
-        get() = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/github/libretube/compat/PendingIntentCompat.kt
```diff
@@ -1,20 +1,30 @@
 package com.github.libretube.compat
 
 import android.app.PendingIntent
+import android.content.Context
+import android.content.Intent
 import android.os.Build
 
+// TODO: Use AndroidX's PendingIntentCompat class instead once it becomes available.
```
Bnyro, line 8: Have there been any announcements for a `PendingIntentCompat` being added to the `androidx` library in the near future recently? (just out of interest)

Bnyro (approved):

app/src/main/java/com/github/libretube/compat/PendingIntentCompat.kt
Isira-Seneviratne: It's there in the repo: https://github.com/androidx/androidx/blob/androidx-main/core/core/src/main/java/androidx/core/app/PendingIntentCompat.java

Bnyro: So it's probably going to be added with the next version bump of `androidx.core`?

Isira-Seneviratne: Seems like it.

PR #2755, opened 2023-01-18
Switch to Kotlinx Serialization.

Isira-Seneviratne:
* Use [Kotlinx Serialization](https://github.com/Kotlin/kotlinx.serialization) instead of Jackson for JSON parsing.
* Serialize/deserialize dates and times using [kotlinx-datetime](https://github.com/Kotlin/kotlinx-datetime).


Isira-Seneviratne:
> What will this change?🤔

Kotlinx Serialization has native Kotlin support (it's by the same devs behind Kotlin), and doesn't rely on reflection.

FireMasterK:
The kotlinx serialization library could also be potentially slower than Jackson too, I will be trying some benchmarking. However, since Libretube is just a client where we won't be deserializing much data, I don't think it would matter much even if it's a bit slower than Jackson if kotlinx provides some QOL code benefits in some way.

I also think using kotlinx serialization would reduce the apk size, but I haven't tested this.

Bnyro:
What's also quite nice about kotkinx is that we don't need to always provide some default constructors which will make the code more elegant as kotlinx is native to Kotlin and can generate its own default constructors.
What should still be tested before merging is whether everything, that had a `null` default constructor previously, still works fine, because the API might return `null` in some edge cases for fields we don't allow it to.
Apart from that I'll have a look on how the app size of a release build changes, just to make sure we're still below 20MB (which shouldn't be an issue I think), as that's a requirement for the IzzyOnDroid repo.

Bnyro:
I've built a release with and without your changes:
Universal APK before: 16.86, after: 16.66
Arm v8 before: 9.60, after: 9.26
So the app size increased notably   :tada:

I've also tested the changes, everything is working as expected.
Awesome work, thank you!

Bnyro (approved):

Bnyro:
Looks like unfortunately we have #2813 as side effect, there seem to be two different issues:
1. `bannerUrl` sometimes is`null` for some channel responses which causes a crash.
2. When scrolling down in a channel page on the videos tab, the newly fetched related videos cause a crash.
I'm not at home today so I can't look into it, if anyone else wants to I wouldn't mind :)

Isira-Seneviratne:
Just got home, I'll take a look at the issues mentioned.

Issue #2261, opened 2022-12-06
Horizontal spacing to seek bar/options dropdown


Arkhtek:
### Describe your suggested feature

Right now, in landscape mode, notches/holepunch cameras block the options dropdown or part of the seek bar. I suggest that you add some form of dynamic adjustment so that this doesn't happen. 

### Other details

this is really minor but its been bugging me for a bit lmao
![Screenshot_20221205-222711](https://user-images.githubusercontent.com/89922603/205824406-dd9e5d7a-2a55-4854-aad2-58a96df8f070.png)

if it helps at all, i use a google pixel 5

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

xIUPITERx:
Can't you detect if a smartphone has a notch etc.

Bnyro:
What difference would that make? I feel like that margin is a good idea for landscape mode in general because the corners of the phone are not that comfortable to reach.

xIUPITERx:
It's more of a thing you're used to from other apps, the spacing is just pretty drastic (over half a centimeter) and I'm missing something design-wise in the empty space.

xIUPITERx:
![Screenshot_20230120_194531_LibreTube Debug.png](https://user-images.githubusercontent.com/97352673/213780959-9076a6af-545b-4487-92f3-456f65ed1eb7.png)

PR #2777, opened 2023-01-20
Only show extra padding when there's a notch

Bnyro:
Follow up to #2261

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/views/CustomExoPlayerView.kt
+++ b/app/src/main/java/com/github/libretube/ui/views/CustomExoPlayerView.kt
@@ -568,6 +568,9 @@ internal class CustomExoPlayerView(
             it.layoutParams = params
         }
 
+        // don't add extra padding if there's no cutout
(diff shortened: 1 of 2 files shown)


xIUPITERx (commented):
What effect exactly are the changes supposed to have now?
I don't see any difference.

Bnyro:
> What effect are the changes supposed to have? I don't see any difference.
>
https://github.com/libre-tube/LibreTube/issues/2261#issuecomment-1370849352

xIUPITERx:
But I think you not currently don't check what kind of cut-out the smartphone has. 
If it's a notch, waterdrop etc. in the middle of the screen, the cut-out doesn't actually interfere with the player control.
You would only need the extra padding if the camera is in one of the corners.

xIUPITERx:
https://developer.android.com/reference/android/view/DisplayCutout
https://developer.android.com/reference/androidx/core/view/WindowInsetsCompat#getDisplayCutout()
I think that could help.

PR #2830, opened 2023-01-22
[Audio mode] Show video options when clicking thumbnail

Bnyro:


NeeRaj-2401:
@Bnyro  I guess this commit should be reverted cause 
Video options such as
**Play in background, play next, add to queue**, (as we are already playing that song in background )
**Share** ( already have that option below thumbnail)
All these are no needed at all.

Except download and add to playlist
 Also if u will keep the download then there will be demand for download audio only etc.

The only option remaining is to keep that add to playlist part which can be adjusted with the other options i.e. below thumbnail

Bnyro:
Well, I didn't find an other place for the download and add to playlist option and that was by far the easiest way to implement it.

NeeRaj-2401:
So while implementing eliminating the rest would be helpful?

PR #2841, opened 2023-01-23
Auto Release on TG after GH Release.

rozari0:
You have to set a new Secret **TG_CID** with Value **libretube** and Add the bot in libretubes channel as well.

For example, check  https://t.me/abuneko/467

### DIFF
--- a/.github/runOnRelease.py
+++ b/.github/runOnRelease.py
@@ -1,0 +1,64 @@
+from pyrogram import Client
+from pyrogram.types import (
+    InlineKeyboardButton,
+    InlineKeyboardMarkup,
+    InputMediaDocument,
+)
+from requests import get
+from tgconfig import TG_API_HASH, TG_API_ID, TG_CID, TG_TOKEN
+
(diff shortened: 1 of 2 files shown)


Bnyro:
Why should we use a secret for `TG_CID` if it's `libretube` anyways?
And where is the bot referenced inside the code, is it the same as we're already using for debug builds? (I know we already have a token set, I just don't know what account it's from)

rozari0:
> Why should we use a secret for `TG_CID` if it's `libretube` anyways?

I needed to check if that works correctly, that's why it's not hardcoded. 
 
> And where is the bot referenced inside the code

In line 57.

> is it the same as we're already using for debug builds?
 
Yes, it's the same. Check debug channel, there shouldn't be more than one bot. 🤷‍♂️

Bnyro:
I assume https://t.me/LibreTubeBot needs writing permissions?

rozari0:
> I assume https://t.me/LibreTubeBot needs writing permissions?

Not any extra permission needed. Just the default one.

PR #2886, opened 2023-01-27
Keep screen on only while video is playing #2877

### LINKED ISSUE #2877: Screen stays on when video is paused (no timeout)
### Steps to reproduce

pause the video

### Expected behavior

Screen should timeout if video is not playing for X minutes 

### Actual behavior

Screen stays on until battery runs out 

### LibreTube version

0.11.1

### Android version

13

### Other details

_No response_

### Acknowledgements

praveenrajput:
- Fixes Screen stays on when video is paused (no timeout)
- closes #2877


app/src/main/java/com/github/libretube/ui/views/CustomExoPlayerView.kt
```diff
@@ -172,17 +172,9 @@ internal class CustomExoPlayerView(
                     )
                 ) {
                     updatePlayPauseButton()
+                    updateScreenOnState()
```
Bnyro, line 175: Why creating a new function here instead of running the code directly?
The function only contains one line and is only called once, so that seems like a bad practice.

Bnyro (changes requested):

app/src/main/java/com/github/libretube/ui/views/CustomExoPlayerView.kt
praveenrajput: I put a single line of code in a function to make the code more readable. If it's not a good way to do it, I'll change it and submit the code again for review.

Bnyro (approved):

PR #2895, opened 2023-01-29
Use HttpUrl in TextUtils.

Isira-Seneviratne:
Use OkHttp's [`HttpUrl`](https://square.github.io/okhttp/4.x/okhttp/okhttp3/-http-url/) class to query URL values.

### DIFF
--- a/app/src/main/java/com/github/libretube/util/TextUtils.kt
+++ b/app/src/main/java/com/github/libretube/util/TextUtils.kt
@@ -1,13 +1,12 @@
 package com.github.libretube.util
 
-import android.net.Uri
-import java.net.URL
 import java.time.format.DateTimeFormatter
 import java.time.format.FormatStyle
 import java.util.*
 import kotlin.time.Duration
 import kotlinx.datetime.LocalDate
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/github/libretube/util/TextUtils.kt
```diff
@@ -66,19 +52,12 @@ object TextUtils {
...
-
-        if (link.contains("youtube.com")) {
-            // the link may be in an unsupported format, so we should try/catch it
-            return try {
-                uri.getQueryParameter("v")
-            } catch (e: Exception) {
-                null
+        return link.toHttpUrlOrNull()?.let {
+            when (it.host) {
+                "youtube.com" -> it.queryParameter("v")
```
FireMasterK, line 57: Couldn't this fail when the host starts with `www.`?

Isira-Seneviratne: Yeah, my bad. Updated.

Bnyro (approved):

PR #2907, opened 2023-01-30
Add string resource for the update disabled message.

Isira-Seneviratne:
Create a string resource for the update disabled message to allow it to be localized.


Bnyro:
The message is only available in debug versions so I don't really think it makes sense to let users translate it, since only users of debug versions will ever see that, and those are mostly people contributing to the project, either translations or code, so they're familiar with English.

Isira-Seneviratne:
Fair enough, I'll close this PR.

Issue #2815, opened 2023-01-22
Unwatched videos marked as watched


noClaps:
### Steps to reproduce

1. search for a video
2. scroll through the list of result

### Expected behavior

The watched video thumbnails should have the duration played bar at the bottom while the unwatched video thumbnails shouldn't.

### Actual behavior

some of the unwatched video thumbnails have the duration played bars

### LibreTube version

0.11.0

### Android version

Android 13

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

Bnyro:
Can't reproduce, also tested it on a fresh installation but can't reproduce either.

akano12:
Faced same issue multiple time, some videos that I've never watched would have watch progress marker.
However, i don't recall facing this on recent version of LT

akano12:
> Can't reproduce, also tested it on a fresh installation but can't reproduce either.

Found out how to reproduce:

1. Search for something
2. Play first video from results (just move the seekbar to anywhere u want)
3. Now every ninth video from below the played video will have the same watch marker as the video you have played

PR #2910, opened 2023-01-30
Added the original comment as the first item in replies.

faisalcodes:

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/fragments/CommentsRepliesFragment.kt
+++ b/app/src/main/java/com/github/libretube/ui/fragments/CommentsRepliesFragment.kt
@@ -5,10 +5,12 @@ import android.util.Log
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
+import androidx.core.view.updatePadding
 import androidx.fragment.app.Fragment
 import androidx.fragment.app.activityViewModels
 import androidx.recyclerview.widget.LinearLayoutManager
 import com.github.libretube.api.RetrofitInstance
+import com.github.libretube.api.obj.Comment
 import com.github.libretube.api.obj.CommentsPage
 import com.github.libretube.constants.IntentData
 import com.github.libretube.databinding.FragmentCommentsBinding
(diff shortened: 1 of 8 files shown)


Bnyro:
The highlighting doesn't fit the rest of the app's theme, instead we should use a surface color for a certain elevation, e.g. 10dp.

Bnyro:
I saw you also hardcode `text_color_secondary` which we should definitely change before the next release because there will be issues with the contrast otherwise.

faisalcodes:
Makes sense.

faisalcodes:
Elevation does not look good so I am only using `R.attr.colorSurface`. It's better now.

faisalcodes:
I have also fixed #2815. I will be doing a PR another day as the changes are in a different computer.

Bnyro:
Awesome!
I will test the PRs changes later and merge it then.

app/src/main/java/com/github/libretube/ui/fragments/CommentsRepliesFragment.kt
```diff
@@ -40,12 +43,16 @@ class CommentsRepliesFragment : Fragment() {
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
         val videoId = arguments?.getString(IntentData.videoId) ?: ""
-        val nextPage = arguments?.getString(IntentData.replyPage) ?: ""
+        val comment = Json.decodeFromString(
+            Comment.serializer(),
+            arguments?.getString(IntentData.comment) ?: """{}"""
```
Bnyro, line 48: Why are you triple quoting `"""{}"""` here?

faisalcodes: I saw this in official Kotlinx-serialization documentation. Anyway that code is never reached because the comment in the argument is guaranteed.

Bnyro: But then we should return instead to make that clear instead of that triple quotes statement

FireMasterK: I think that might be a multi-line string if I'm not wrong

Bnyro: Or instead use `!!`

Bnyro (approved):

ARBoyGo:
@faisalcodes Thanks your making to awesome 😌

PR #2944, opened 2023-02-02
Add Mastodon Widget

XelXen:
Merge Website repo PR first


Bnyro:
I would like that to become the second social media icon (after Matrix) because it's open source, thus we should promote it more than the others.

XelXen:
okay

PR #2960, opened 2023-02-04
Use POST requests for unauthenticated subscriptions

Bnyro:
see https://github.com/TeamPiped/Piped-Backend/pull/512


FireMasterK:
I want to use the GET method if its 100 or lesser channels, POST only if it's more than that.

Bnyro:
> I want to use the GET method if its 100 or lesser channels, POST only if it's more than that.
>
Why though?

FireMasterK:
The get method can be cached by the reverse proxy, but the post method cannot be cached.

PR #2993, opened 2023-02-08
Use LifecycleService.

Isira-Seneviratne:
Use the [`LifecycleService`](https://developer.android.com/reference/androidx/lifecycle/LifecycleService) class with the download and background mode services.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -92,6 +92,7 @@ dependencies {
     implementation libs.lifecycle.viewmodel
     implementation libs.lifecycle.runtime
     implementation libs.lifecycle.livedata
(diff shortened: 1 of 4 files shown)


Bnyro:
So if I understand it correctly, this will allow us to use view models in these services?

Isira-Seneviratne:
> So if I understand it correctly, this will allow us to use view models in these services?

No, they allow any launched coroutines to be bound to the service's life cycle.

Bnyro:
Oh, my bad, I mixed up `lifecycleOwner` and `viewLifecycleOwner` ;)

PR #3037, opened 2023-02-12
Use view binding extensions in PlayerFragment.

Isira-Seneviratne:
* Add the [ViewBindingPropertyDelegate](https://github.com/androidbroadcast/ViewBindingPropertyDelegate) library.
* Use view binding extensions in the player fragment to avoid memory leaks.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -122,6 +122,9 @@ dependencies {
     /* Room */
     kapt libs.room.compiler
     implementation libs.room
+
+    /* View binding */
+    implementation libs.viewBindingPropertyDelegate
 }
 
 static def getUnixTime() {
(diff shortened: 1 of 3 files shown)


Bnyro:
I think we should consider something as described [here](https://proandroiddev.com/make-android-view-binding-great-with-kotlin-b71dd9c87719) to avoid having the same code to release and get the bindings for each fragment.

Isira-Seneviratne:
> I think we should consider something as described [here](https://proandroiddev.com/make-android-view-binding-great-with-kotlin-b71dd9c87719) to avoid having the same code to release and get the bindings for each fragment.

Sounds good.

Bnyro (approved):

Bnyro:
Unfortunately I had to undo some of the changes within the `PlayerFragment` via #3341  since there have been some issues as tracked in #3245.

PR #3119, opened 2023-02-18
Fix no connection with gnirehtet reverse tethering

Linus789:
I’m not too sure why the old code was commented out, so please take a look if it’s done right this time. Now it doesn’t say anymore that I have no connection if I’m using [Gnirehtet](https://github.com/Genymobile/gnirehtet).

### DIFF
--- a/app/src/main/java/com/github/libretube/helpers/NetworkHelper.kt
+++ b/app/src/main/java/com/github/libretube/helpers/NetworkHelper.kt
@@ -2,6 +2,8 @@ package com.github.libretube.helpers
 
 import android.content.Context
 import android.net.ConnectivityManager
+import android.net.NetworkCapabilities
+import android.os.Build
 import androidx.core.content.getSystemService
 
 object NetworkHelper {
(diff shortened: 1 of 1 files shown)


Linus789:
Do I need to pass the reviewdog?

Bnyro:
Would be good, yes, please run `ktlint --android -F` in the root directory of the project.

Bnyro:
Otherwise if you can't get ktlint running I can do that too before merging :)

app/src/main/java/com/github/libretube/helpers/NetworkHelper.kt
```diff
@@ -2,40 +2,34 @@ package com.github.libretube.helpers
...
-        }
-         */
+            if (connectivityManager.activeNetworkInfo?.isConnected == true) {
+                return true
+            }
 
-        @Suppress("DEPRECATION")
-        return connectivityManager?.activeNetworkInfo?.isConnected ?: false
+            // VPNs might not show up in the activeNetworkInfo, so check each one by one
+            return connectivityManager.allNetworkInfo.filterNotNull().any { networkInfo ->
```
Linus789, line 29: Actually, it might be possible to use `connectivityManager.getNetworkInfo(ConnectivityManager.TYPE_VPN)` instead of iterating over all of them. I will take a look at it.

Linus789: Okay, seems to work. I changed the PR accordingly.

Bnyro (approved):

Bnyro:
Stills seems to work as expected, thanks! I hope some others using a VPN can test the debug builds in the next time just to make sure it still works for them, since we had issues with that earlier.

PR #3124, opened 2023-02-19
Use lifecycle coroutine scope extensions.

Isira-Seneviratne:


Bnyro:
After that change, I get `Job was cancelled` as error quite often, apart from that the library doesn't seem to update anymore after creating a new playlist while being logged in :/

Bnyro:
Isn't it the normal behavior that `lifecycleScope#launch` gets cancelled when the dialog is closed, which would be contra-productive in our case?

Isira-Seneviratne:
> After that change, I get `Job was cancelled` as error quite often, apart from that the library doesn't seem to update anymore after creating a new playlist while being logged in :/

That's weird, I'll check it out.

Isira-Seneviratne:
> Isn't it the normal behavior that `lifecycleScope#launch` gets cancelled when the dialog is closed, which would be contra-productive in our case?

I made a mistake; the call to dismiss the playlist creation dialog was placed outside the launched coroutine, I'll fix that.

Bnyro:
The creation of playlists seems to be fixed now :tada:
However, I still got issues with `Job was cancelled` when renaming a playlist by long pressing it (with account).
And now the `CreatePlaylist` dialog stays visible while the playlist is getting created, the reason why we used `CoroutineScope(Dispatchers.IO)` there is for creating the playlist in the background, so that the dialog doesn't need to stay visible.

Bnyro (approved):

Bnyro:
> Are there any dialogs I missed? I updated the clone, create and rename dialogs a few hours ago.
>
The `AddToPlaylist` dialog had required some changes too since I wasn't able to add videos to playlists after the changes, I quickly pushed a commit fixing that. Anything else works fine and hiding the dialogs seems to be a nice hack! :)

Isira-Seneviratne:
> > Are there any dialogs I missed? I updated the clone, create and rename dialogs a few hours ago.
> >
> The `AddToPlaylist` dialog had required some changes too since I wasn't able to add videos to playlists after the changes, I quickly pushed a commit fixing that.

Thanks, I missed that one.

PR #3214, opened 2023-03-01
Format watch history item dates.

Isira-Seneviratne:
Format the dates for watch history items in the same style as in the video description.


Bnyro:
Is that going to work flawlessly without any database migrations? Since we usually need to run migrations when changing columns of a table.

Isira-Seneviratne:
> Is that going to work flawlessly without any database migrations? Since we usually need to run migrations when changing columns of a table.

Yeah, the type stored in the database is still a string, this just provides a convenient mechanism to convert to/from `LocalDate`.

Bnyro:
Alright, I just wasn't sure whether Room recognizes that the database structure didn't change in fact.

PR #3264, opened 2023-03-07
strings.xml

Qaz-7:
"?" fixed🤭


Bnyro:
The string is being used as an statement and not as a question inside the app.

PR #3290, opened 2023-03-11
Add option to disable Piped proxy

### LINKED ISSUE #3274: Add ability to not proxy through Piped

Hey there LibreTube team,

I have a feature request that I believe would benefit many users of the app. As you know, LibreTube currently proxies YouTube videos through instances. However, I think it would be great to have the option for users to not proxy videos through instances, for a few reasons.

Firstly, many privacy-focused users of LibreTube are likely already using services like VPNs to hide their IP addresses. For those users, adding an additional proxy like Piped may not be necessary, and may even be redundant. By giving users the option to bypass Piped, you could free up server resources and potentially save users money on proxy costs.
I'm for example using LibreTube because it looks so much better than NewPipe and I'm using a VPN - I would be very open to only use that VPN and not use any of Piped resources.

Additionally, as the name "LibreTube" suggests, the app is focused on user freedom and choice. Allowing users to choose how they connect to Google and YouTube would align with this principle and give users more control over their network traffic.

I know this may sound a bit strange at first, but I hope you'll consider this feature request. Thank you for all the hard work you've put into the app so far!

### Other details

_No response_

### Acknowledgements

Bnyro:
closes #3274


FireMasterK:
What if it's a music video? These will break if you just disable proxying, as they're IP restricted

Bnyro:
Well, but can we distinguish normal videos and music videos by the API response?

FireMasterK:
That's true too, not entirely sure what we can do. I think we can send a HEAD request to any streaming URL, and see if that 403s and then maybe decide whether we forcibly keep the proxy enabled.

app/src/main/java/com/github/libretube/helpers/ProxyHelper.kt
```diff
@@ -27,4 +27,24 @@ object ProxyHelper {
...
+        ) {
+            return url
+        }
+
+        return url.toHttpUrlOrNull()?.let {
+            it.newBuilder()
+                .host(it.queryParameter("host").orEmpty())
+                .removeAllQueryParameters("host")
+                .build()
+        }.toString()
```
Isira-Seneviratne, line 48:
```suggestion
        return url.toHttpUrlOrNull()?.let {
            it.newBuilder()
                .host(it.queryParameter("host").orEmpty())
                .removeAllQueryParameters("host")
                .build()
                .toString()
        }
```

This avoids returning a "null" string if the URL is invalid.

Bnyro:
I've added a note that it might not work with content from YT music now, if there are many complaints about it we should indeed consider doing HEAD requests before starting the playback.

techgamers2:
thanks for adding this feature it helps in certain situation <3

PR #3311, opened 2023-03-16
Use java.nio.file APIs in the download functionality.

Isira-Seneviratne:
* Switch to the [`desugar_jdk_libs_nio`](https://developer.android.com/studio/write/java11-nio-support-table) artifact.
* Use the [`Path`](https://developer.android.com/reference/java/nio/file/Path) and [`Files`](https://developer.android.com/reference/java/nio/file/Files) APIs in the download functionality.

### DIFF
--- a/app/src/main/java/com/github/libretube/helpers/ImageHelper.kt
+++ b/app/src/main/java/com/github/libretube/helpers/ImageHelper.kt
@@ -6,6 +6,7 @@ import android.graphics.BitmapFactory
 import android.net.Uri
 import android.widget.ImageView
 import androidx.core.graphics.drawable.toBitmap
+import androidx.core.net.toUri
 import coil.ImageLoader
 import coil.disk.DiskCache
 import coil.load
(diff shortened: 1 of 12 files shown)


app/src/main/java/com/github/libretube/helpers/ImageHelper.kt
```diff
@@ -54,9 +56,10 @@ object ImageHelper {
         if (!DataSaverMode.isEnabled(target.context)) target.load(url, imageLoader)
     }
 
-    fun downloadImage(context: Context, url: String, path: String) {
+    fun downloadImage(context: Context, url: String, path: Path) {
         getAsync(context, url) { bitmap ->
-            saveImage(context, bitmap, Uri.fromFile(File(path)))
+            @Suppress("NewApi") // The Path class is desugared.
+            saveImage(context, bitmap, path.toFile().toUri())
```
Bnyro, line 62: I guess we can use `Path#toAndroidUri` here too.

Bnyro (approved):

PR #3375, opened 2023-03-24
Correct Kurdish (Central) native language name

GoRaN909:
Correct Kurdish (Central) native language name and moved to below of Catalan language by language code ckb


Bnyro:
Thanks! May I ask why you have been changing its position?

Bnyro (approved):

PR #3385, opened 2023-03-25
Use CompactDecimalFormat to format counts.

Isira-Seneviratne:
Use [`CompactDecimalFormat`](https://developer.android.com/reference/android/icu/text/CompactDecimalFormat) to format counts on Android 7.0 and later.

**Screenshots**

Before

Screenshot_20230325_190050_LibreTube

After

Screenshot_20230325_190101_LibreTube Debug

### DIFF
--- a/app/src/main/java/com/github/libretube/extensions/FormatShort.kt
+++ b/app/src/main/java/com/github/libretube/extensions/FormatShort.kt
@@ -1,14 +1,22 @@
 package com.github.libretube.extensions
 
+import android.icu.text.CompactDecimalFormat
+import android.os.Build
+import java.util.*
 import kotlin.math.pow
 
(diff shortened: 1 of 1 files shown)


Bnyro (approved):
That looks great, pretty unfortunate that it didn't get added before SDK 24.

Isira-Seneviratne:
> That looks great, pretty unfortunate that it didn't get added before SDK 24.

The ICU4J library could be included for older Android versions: https://developer.android.com/guide/topics/resources/internationalization

Bnyro:
> The ICU4J library could be included for older Android versions: https://developer.android.com/guide/topics/resources/internationalization
>
Hmm, I don't think that would be worth the trouble.

PR #3411, opened 2023-03-27
Add picture-in-picture helper classes.

Isira-Seneviratne:


Bnyro (approved):
Looks good to me!
Just one note: `com.github.libretube.ui.extensions#SetAspectRatio` isn't needed anymore after the change, so it'd be great if you could already remove that within this PR.

PR #3445, opened 2023-03-29
Fix dark status bar icons when player is in fullscreen

anilbeesetti:
Before:
Screenshot_20230328-201725_LibreTube_Debug
After:
Screenshot_20230328-203936_LibreTube_Debug


app/src/main/java/com/github/libretube/ui/extensions/Activity.kt
```diff
@@ -0,0 +1,22 @@
...
+        }
+    }
+
+fun Activity.setLightStatusBarIcons() {
+    WindowCompat.getInsetsController(window, window.decorView).isAppearanceLightStatusBars = false
+}
+
+fun Activity.setDarkStatusBarIcons() {
+    WindowCompat.getInsetsController(window, window.decorView).isAppearanceLightStatusBars = true
+}
```
Isira-Seneviratne, line 22: Looks like the method names should be swapped.

anilbeesetti: They are correct. Here we have to change the statusbar icon color but there's no direct method to change icon color directly. So if we set the statusbar as light then icons will be dark

Isira-Seneviratne: Right, missed that.

app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
```diff
@@ -520,6 +526,13 @@ class PlayerFragment : Fragment(), OnlinePlayerOptions {
             enableTransition(R.id.yt_transition, true)
         }
 
+        // set status bar icon color back to theme color
+        if (mainActivity.isLightTheme) {
+            mainActivity.setDarkStatusBarIcons()
+        } else {
+            mainActivity.setLightStatusBarIcons()
+        }
```
Isira-Seneviratne, line 534:
```suggestion
        windowInsetsControllerCompat.isAppearanceLightStatusBars = when (resources.configuration.uiMode and Configuration.UI_MODE_NIGHT_MASK) {
            Configuration.UI_MODE_NIGHT_YES -> false
            Configuration.UI_MODE_NIGHT_NO -> true
            else -> true
        }
```

app/src/main/java/com/github/libretube/ui/extensions/Activity.kt
anilbeesetti: Yes we can do that. It's just a preference. By doing what you suggested the code is not reusable and using functions make code more readable in my opinion

Isira-Seneviratne: True, but it's not like the code is being used anywhere else in your changeset. If it needs to be used in other places as well, it can be easily refactored.

Isira-Seneviratne (approved):
Looks good to me.

Bnyro (approved):

Bnyro:
Thank you!

Issue #3069, opened 2023-02-13
Navigation buttons get in the way of player UI element


Pentaphon:
### Steps to reproduce

See screenshot. The last player UI element on the top right is covered by the navigation button bar. 
![Screenshot_20230213-115754](https://user-images.githubusercontent.com/49374058/218563270-383fbd26-2043-4cb7-8788-0412af5de86c.png)

### Expected behavior

Navigation button bar should not get in the way of the player UI.

### Actual behavior

Navigation button bar overlaps the player UI. When the navigation buttons go away, so does the player UI, which reduces functionalily. 

### LibreTube version

0.12

### Android version

Android 11

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

Pentaphon:
> Shouldn't the navigationbar be hidden in fullscreen mode?

It shows up when you tap once.

KrychuTM:
https://user-images.githubusercontent.com/79592592/218858905-8e85a4d2-b014-4c6c-a43c-5291d73199f0.mp4

https://user-images.githubusercontent.com/79592592/218858904-36bcbbc8-1ba3-43af-a80b-6016d1d90f30.mp4

socynx:
https://user-images.githubusercontent.com/125415316/218890228-9129bc33-5831-484f-bc88-a7314bfd68ac.mp4

xipe-totek:
I have this problem on tablet (HTC a101, Android 11). system navigation bar completely hides player buttons. but if you tap on the triangle, it become like on the second screenshot. so it is more inconvinience than a problem. because you need tap on that triangle button every time. ![Screenshot_20230228-094745.png](https://user-images.githubusercontent.com/81915152/221777873-271e35f0-fd47-480d-9aa8-b1e5d8325448.png)
![Screenshot_20230228-094830.png](https://user-images.githubusercontent.com/81915152/221778375-0db7c0f3-6600-46bc-ad5d-b25365d56759.png)

Bnyro:
Has the latest release fixed that?

Pentaphon:
> Has the latest release fixed that?

The buttons don't show up anymore in fullscreen so technically it has. I'd wait to hear what other people think.

CactiChameleon9:
> If nobody responds in a week, I'll close this issue as fixed.
[...]

PR #3450, opened 2023-03-30
Fix two taps required to interact with player when in fullscreen mode

anilbeesetti:
previously we had to tap twice to show controllers. On first tap the status bar will be shown and on the next tap player controller will be shown. This pr fixes this by showing controller on single tap

### DIFF
--- a/app/src/main/java/com/github/libretube/helpers/WindowHelper.kt
+++ b/app/src/main/java/com/github/libretube/helpers/WindowHelper.kt
@@ -20,25 +20,24 @@ object WindowHelper {
 
         WindowCompat.setDecorFitsSystemWindows(window, !isFullscreen)
 
-        val controller = WindowCompat.getInsetsController(window, window.decorView)
-        val flags = WindowInsetsCompat.Type.systemBars() or WindowInsetsCompat.Type.navigationBars()
         if (isFullscreen) {
-            controller.hide(flags)
+            activity.hideSystemBars()
         } else {
-            controller.show(flags)
+            activity.showSystemBars()
         }
+    }
+}
 
-        controller.systemBarsBehavior = if (isFullscreen) {
-            WindowInsetsControllerCompat.BEHAVIOR_SHOW_TRANSIENT_BARS_BY_SWIPE
-        } else {
-            WindowInsetsControllerCompat.BEHAVIOR_SHOW_BARS_BY_TOUCH
-        }
+fun Activity.hideSystemBars() {
+    WindowCompat.getInsetsController(window, window.decorView).apply {
+        systemBarsBehavior = WindowInsetsControllerCompat.BEHAVIOR_SHOW_TRANSIENT_BARS_BY_SWIPE
+        hide(WindowInsetsCompat.Type.systemBars() or WindowInsetsCompat.Type.navigationBars())
+    }
+}
 
-        val layoutFlag = WindowManager.LayoutParams.FLAG_LAYOUT_NO_LIMITS
-        if (isFullscreen) {
-            window.setFlags(layoutFlag, layoutFlag)
-        } else {
-            window.clearFlags(layoutFlag)
-        }
+fun Activity.showSystemBars() {
+    WindowCompat.getInsetsController(window, window.decorView).apply {
+        systemBarsBehavior = WindowInsetsControllerCompat.BEHAVIOR_SHOW_TRANSIENT_BARS_BY_SWIPE
+        show(WindowInsetsCompat.Type.systemBars() or WindowInsetsCompat.Type.navigationBars())
     }
 }
--- a/app/src/main/java/com/github/libretube/helpers/WindowHelper.kt
+++ b/app/src/main/java/com/github/libretube/helpers/WindowHelper.kt
@@ -6,6 +6,8 @@ import android.view.WindowManager
 import androidx.core.view.WindowCompat
 import androidx.core.view.WindowInsetsCompat
 import androidx.core.view.WindowInsetsControllerCompat
+import com.github.libretube.ui.extensions.hideSystemBars
+import com.github.libretube.ui.extensions.showSystemBars
 
 object WindowHelper {
     fun toggleFullscreen(activity: Activity, isFullscreen: Boolean) {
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/github/libretube/ui/extensions/Activity.kt
```diff
@@ -0,0 +1,20 @@
...
+
+import android.app.Activity
+import androidx.core.view.WindowCompat
+import androidx.core.view.WindowInsetsCompat
+import androidx.core.view.WindowInsetsControllerCompat
+
+fun Activity.hideSystemBars() {
+    WindowCompat.getInsetsController(window, window.decorView).apply {
+        systemBarsBehavior = WindowInsetsControllerCompat.BEHAVIOR_SHOW_TRANSIENT_BARS_BY_SWIPE
+        hide(WindowInsetsCompat.Type.systemBars() or WindowInsetsCompat.Type.navigationBars())
```
Bnyro, line 11: The navigation bars shouldn't be made visible here as that's just a bad practice for UX. We also had some issues open about that previously.

Bnyro (changes requested):

Bnyro:
Also, please remove the unused imports in the `WindowHelper` file.

app/src/main/java/com/github/libretube/ui/extensions/Activity.kt
```diff
@@ -0,0 +1,20 @@
...
+    }
+}
+
+fun Activity.showSystemBars() {
+    WindowCompat.getInsetsController(window, window.decorView).apply {
+        systemBarsBehavior = WindowInsetsControllerCompat.BEHAVIOR_SHOW_TRANSIENT_BARS_BY_SWIPE
```
Isira-Seneviratne, line 17:
```suggestion
        systemBarsBehavior = WindowInsetsControllerCompat.BEHAVIOR_SHOW_BARS_BY_TOUCH
```

Isira-Seneviratne: Also, these two methods could be easily combined into one, e.g. `toggleSystemBars(show: Boolean)`.

anilbeesetti: No. `WindowInsetsControllerCompat.BEHAVIOR_SHOW_BARS_BY_TOUCH` conflict with touch to show player controller. i.e now we have to tap twice to show player controller

app/src/main/java/com/github/libretube/ui/views/CustomExoPlayerView.kt
```diff
@@ -202,6 +205,22 @@ internal class CustomExoPlayerView(
...
+                playerViewModel?.isFullscreen?.value?.let { isFullscreen ->
+                    if (!isFullscreen) return@let
+                    when (visibility) {
+                        View.VISIBLE -> {
+                            activity.showSystemBars(WindowInsetsCompat.Type.statusBars())
+                        }
+                        View.GONE -> {
+                            activity.hideSystemBars(WindowInsetsCompat.Type.statusBars())
+                        }
+                    }
```
Isira-Seneviratne, line 219:
```suggestion
                    val windowInsetsControllerCompat = WindowCompat.getInsetsController(activity.window, activity.window.decorView)
                    windowInsetsControllerCompat.systemBarsBehavior = WindowInsetsControllerCompat.BEHAVIOR_SHOW_TRANSIENT_BARS_BY_SWIPE
                    if (visibility == View.VISIBLE) {
                        windowInsetsControllerCompat.show(WindowInsetsCompat.Type.statusBars())
                    } else {
                        windowInsetsControllerCompat.hide(WindowInsetsCompat.Type.statusBars())
                    }
```

anilbeesetti: What's wrong with when

Isira-Seneviratne: According to the documentation for the listener, only View.VISIBLE and View.GONE are passed, so this is simpler.

app/src/main/java/com/github/libretube/helpers/WindowHelper.kt
```diff
@@ -20,25 +21,10 @@ object WindowHelper {
 
         WindowCompat.setDecorFitsSystemWindows(window, !isFullscreen)
 
-        val controller = WindowCompat.getInsetsController(window, window.decorView)
-        val flags = WindowInsetsCompat.Type.systemBars() or WindowInsetsCompat.Type.navigationBars()
         if (isFullscreen) {
-            controller.hide(flags)
+            activity.hideSystemBars(WindowInsetsCompat.Type.systemBars())
```
Bnyro, line 25: Due to #3069, hiding the navigation bars here too might fix issues for some devices.

app/src/main/java/com/github/libretube/ui/views/CustomExoPlayerView.kt
```diff
@@ -202,6 +205,22 @@ internal class CustomExoPlayerView(
...
+        setControllerVisibilityListener(
+            ControllerVisibilityListener { visibility ->
+                playerViewModel?.isFullscreen?.value?.let { isFullscreen ->
+                    if (!isFullscreen) return@let
+                    when (visibility) {
+                        View.VISIBLE -> {
+                            activity.showSystemBars(WindowInsetsCompat.Type.statusBars())
+                        }
+                        View.GONE -> {
+                            activity.hideSystemBars(WindowInsetsCompat.Type.statusBars())
```
Bnyro, line 217: Same as above: Due to #3069, hiding the navigation bars here too might fix issues for some devices.

Bnyro:
I feel like we could also just create a function ```
fun Activity.toggleSystemBars(showBars: Boolean)
```
which then calls the corresponding methods since we're currently doing two boolean if expressions and different places which could be simplified that way. Though I'm not sure whether that's really needed since it's only being used twice in total, but I probably wouldn't be too bad.

app/src/main/java/com/github/libretube/helpers/WindowHelper.kt
anilbeesetti: Screenshot_20230401-122249_MX_Player_Pro

Do want navigation bars this way or not appear when controls are visible

Bnyro: They should not appear, but if the user manually shows them they should be hidden again automatically.

anilbeesetti:
see this video for reference

https://user-images.githubusercontent.com/66936649/229292694-205d9840-d41d-4f30-9df6-6b7ab8f33fbd.mp4

Bnyro (approved):

Bnyro:
Thank you!

PR #3487, opened 2023-04-04
Add single item insertion methods.

Isira-Seneviratne:


FireMasterK:
A single insert is likely also slower, but I don't think it matters much here, since we won't be inserting many items.

Isira-Seneviratne:
I thought it would be useful to avoid having to create lists when inserting single items.

FireMasterK (approved):
Seems fine to me!

Bnyro (approved):

PR #3502, opened 2023-04-08
Free fragment view bindings when their fragments are destroyed.

Isira-Seneviratne:
Free the fragment view bindings when their host fragments are destroyed to avoid memory leaks.


Bnyro (approved):
The changes look good to me!
However, we should make sure to test them well before the next release in order to prevent possible crashes caused by any `Handler#postDelayed` calls that try to access destroyed bindings or something similar, but it seems like you already took care of a lot of these possible issues.

PR #3510, opened 2023-04-09
Switch to AndroidX's PendingIntentCompat class.

Isira-Seneviratne:


Bnyro:
I assume you added the AndroidX dependency directly because we can use a later version of it this way?

Bnyro (approved):

Isira-Seneviratne:
> I assume you added the AndroidX dependency directly because we can use a later version of it this way?

Yeah.

PR #3511, opened 2023-04-09
In-built browser intent chooser

Bnyro:
Newer Android versions do always open links inside the default browser of the phone instead of letting the user choose which browser to use. Using `Intent#createChooser` behaves the same way. In order to allow users to choose which browser they actually want to use, the app queries for available packages. Silent contribution from @AudricV, because they're working hard to support the development in the background :)

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/sheets/IntentChooserSheet.kt
+++ b/app/src/main/java/com/github/libretube/ui/sheets/IntentChooserSheet.kt
@@ -1,0 +1,31 @@
+package com.github.libretube.ui.sheets
+
+import android.content.pm.ResolveInfo
+import android.os.Bundle
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import androidx.recyclerview.widget.GridLayoutManager
+import com.github.libretube.databinding.BottomSheetBinding
+import com.github.libretube.ui.adapters.IntentChooserAdapter
+
+class IntentChooserSheet(
+    private val packages: List<ResolveInfo>,
+    private val url: String
+) : BaseBottomSheet() {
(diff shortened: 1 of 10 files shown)


Qaz-7:
In general, can all YouTube links be opened with LibreTube?🤔

Bnyro:
> In general, can all YouTube links be opened with LibreTube?🤔
>
Yes, but that's not related to this PR.

app/src/main/java/com/github/libretube/ui/sheets/IntentChooserSheet.kt
```diff
@@ -0,0 +1,31 @@
...
+        savedInstanceState: Bundle?
+    ): View {
+        binding = BottomSheetBinding.inflate(inflater)
+        return binding.root
+    }
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        binding.optionsRecycler.layoutManager = GridLayoutManager(context, 3)
+        binding.optionsRecycler.adapter = IntentChooserAdapter(packages, url)
+    }
```
Isira-Seneviratne, line 30:
```suggestion
    private var _binding: BottomSheetBinding? = null
    private val binding get() = _binding!!

    override fun onCreateView(
        inflater: LayoutInflater,
        container: ViewGroup?,
        savedInstanceState: Bundle?
    ): View {
        _binding = BottomSheetBinding.inflate(inflater)
        return binding.root
    }

    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
        binding.optionsRecycler.layoutManager = GridLayoutManager(context, 3)
        binding.optionsRecycler.adapter = IntentChooserAdapter(packages, url)
    }
    
    override fun onDestroyView() {
        _binding = null
    }
```

PR #3525, opened 2023-04-10
Audio/background player for downloads

Bnyro:


app/src/main/java/com/github/libretube/services/OfflinePlayerService.kt
```diff
@@ -0,0 +1,114 @@
...
+import com.google.android.exoplayer2.MediaItem
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
+
+/**
+ * A service to play downloaded audio in the background
+ */
+class OfflinePlayerService : Service() {
```
Isira-Seneviratne, line 29:
```suggestion
class OfflinePlayerService : LifecycleService() {
```

app/src/main/java/com/github/libretube/services/OfflinePlayerService.kt
```diff
@@ -0,0 +1,114 @@
...
+            .setSmallIcon(R.drawable.ic_launcher_lockscreen)
+            .build()
+
+        startForeground(PLAYER_NOTIFICATION_ID, notification)
+    }
+
+    override fun onStartCommand(intent: Intent, flags: Int, startId: Int): Int {
+        val videoId = intent.getStringExtra(IntentData.videoId)
+
+        CoroutineScope(Dispatchers.IO).launch {
```
Isira-Seneviratne, line 48:
```suggestion
        lifecycleScope.launch(Dispatchers.IO) {
```

Isira-Seneviratne (changes requested):
Overall, this looks good to me. I've suggested a couple of small changes.

mxwmnn:
Works partially, you can't skip songs it only plays one audio and doesn't have continuous play. Would be nice. Thanks for the effort guys

PR #3527, opened 2023-04-11
Switch to repeatOnLifecycle extension.

Isira-Seneviratne:
Switch from the `launchWhenCreated` extension to the `repeatOnLifecycle` extension, as the former has been deprecated.


app/src/main/java/com/github/libretube/ui/fragments/HomeFragment.kt
```diff
@@ -75,13 +78,13 @@ class HomeFragment : Fragment() {
     }
 
     private fun fetchHomeFeed() {
-        lifecycleScope.launchWhenCreated {
```
Bnyro, line 78: The goal here has been to load some of them simultaneously in order to benefit from faster loading times in the home page in general. Thus, I don't think this should be changed.

Isira-Seneviratne: In that case, I could run them concurrently using `async` blocks.

Bnyro (approved):

PR #3542, opened 2023-04-13
Display stream notification thumbnails.

Isira-Seneviratne:
Display the thumbnails of new streams in their notifications.

**Screenshots**

Android 5.0
Screenshot on Android 5

Android 8.1
Screenshot_20230413-082019_LibreTube Debug
Screenshot_20230413-082124_LibreTube Debug

Settings page
Screenshot_20230413-172032_LibreTube Debug

### DIFF
--- a/app/src/main/java/com/github/libretube/constants/PreferenceKeys.kt
+++ b/app/src/main/java/com/github/libretube/constants/PreferenceKeys.kt
@@ -107,6 +107,7 @@ object PreferenceKeys {
      * Notifications
      */
     const val NOTIFICATION_ENABLED = "notification_toggle"
+    const val SHOW_STREAM_THUMBNAILS = "show_stream_thumbnails"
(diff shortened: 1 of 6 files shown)


Bnyro:
I'm not sure about this in terms of data usage. Most users probably don't expect the notifications to show thumbnails and if enabled on mobile data this could use quite a lot of it.
Maybe we should consider turning it into an option, at least?

Isira-Seneviratne:
> I'm not sure about this in terms of data usage. Most users probably don't expect the notifications to show thumbnails and if enabled on mobile data this could use quite a lot of it. Maybe we should consider turning it into an option, at least?

Yeah, the data saving mode could be checked. It's already being done for the image loading.

Bnyro:
Hmm, I personally would just prefer a dedicated option for it here as I don't want to use the data saver mode since I only use the app on WiFi, but still want to receive notifications about the new streams on mobile data, which should be as data saving as possible.

PR #3585, opened 2023-04-18
Upgrade to AGP 8.0.

Isira-Seneviratne:
* Update the Android Gradle Plugin to 8.0.
* Set the JDK version to 17.
* Remove some unnecessary API suppressions (the lint issue is [not present](https://issuetracker.google.com/issues/276929548) on Android Studio Flamingo).


Bnyro:
Android Studio does not yet support the latest AGP version as far as I know.

Isira-Seneviratne:
> BTW, do you have a reason to use version 8.0.0 of the Gradle build tool? That version was released on 13th February. Since then there was some patch releases. The latest version is Gradle [8.1.1](https://docs.gradle.org/8.1.1/release-notes.html).

It's required by AGP 8.0.

Bnyro:
That makes sense, however we need to make sure F-Droid supports later versions of gradle than 8.0 too, recently we weren't updated on F-Droid for a month due to bumping the gradle version to one that's not been supported by F-Droid.

Bnyro:
Unfortunately I'm not able to test this because I'm using IntelliJ which doesn't yet support the AGP 8.0. I'll probably need to switch to Android Studio then, which however is unfortunately not part of my distro's  package repos. 
So apologies for the delay, I'll see whether I can find an other way, e.g. to update the Android Plugin in IntelliJ.

Bnyro:
I've tested the changes and it now seems to be all good to me, the only thing we should get running before merging is CodeQL.

Bnyro:
Thanks @FireMasterK and @Isira-Seneviratne

Bnyro (approved):

Issue #3610, opened 2023-04-23
create new playlist issue


nayeemdv:
### Steps to reproduce

1. click FAB to create new playlist
2. click "create playlist" button (playlist name should be empty)
3. enter playlist name And click on "create playlist" button

https://user-images.githubusercontent.com/91174350/233821392-460fe726-68f1-4e11-af10-9346bb76bd7c.mp4

### Expected behavior

After entering the playlist name, the playlist should be created

### Actual behavior

"create playlist" button gets disabled

### LibreTube version

0.14.0

### Android version

Android 11

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

PR #3611, opened 2023-04-23
creating new playlist

nayeemdv:
Fixed an issue "creating new playlist"(#3610 )

I try to spam a lot when creating a playlist but it doesn't create multiple playlists
so there is no point to disabling OnClickListener
Please correct me if I'm wrong!!


Bnyro (approved):
You're right, thank you!

Issue #2922, opened 2023-01-31
Ability to interact with player while viewing comments


ARBoyGo:
### Describe your suggested feature

After the PR #2878 
The comment Multi level, We need ability to interact with the player. 

When reading a comment plus player a video, There are some moments we need to pause and read it.
Hope you understand my feature request. 

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

faisalcodes:
You are talking about touches not being dispatched to the player when the comments dialog is visible.

ARBoyGo:
Yes, Like I'm reading a comment, Maybe related to a time stamp, I need to pause and read. It's good to interact with the player, When we are reading comments.

akano12:
https://github.com/libre-tube/LibreTube/issues/2522#issuecomment-1366154175

PR #3624, opened 2023-04-25
Ability to interact with player while viewing comments

xz-dev:
Fix way: https://stackoverflow.com/questions/54798186/bottomsheetdialogfragment-allow-user-to-touch-outside
Fix https://github.com/libre-tube/LibreTube/issues/2922


xz-dev:
Edit git log and remove one useless line in code

Bnyro:
Works like a charme, thank you!

PR #3632, opened 2023-04-26
Half expand comment dialog when roll player to avoid comments dialog hide the player

xz-dev:
Fix a bug from https://github.com/libre-tube/LibreTube/pull/3624


xz-dev:
I think everything is working well, any more other suggestions?
(Also, if you think my work is helping you, please consider [donating me :smile: ](https://ko-fi.com/xzdev/goal?g=0))

app/src/main/java/com/github/libretube/ui/sheets/ExpandedBottomSheet.kt
```diff
@@ -36,4 +29,18 @@ open class ExpandedBottomSheet : BottomSheetDialogFragment() {
...
     }
+
+    fun expand() {
+        bottomSheet?.let { fl ->
+            val bottomSheetInfoBehavior = BottomSheetBehavior.from(fl)
+            bottomSheetInfoBehavior.state = BottomSheetBehavior.STATE_EXPANDED
+        }
+    }
+
+    fun halfExpand() {
```
Bnyro, line 40: I think we should instead add a parameter called `expandHalf` (boolean) to the function `expand` instead of creating two different functions doing basically the same thing.

xz-dev: You are right, I had want to hide it, but I find it will just dismiss the dialog, so I change it to halfExpand

xz-dev:
> 

yeah, I agree with you, I love to code opensource APP.
But I'm really close to not being able to pay my rent this month, and due to some personal issues, I can't go to work for a while, which is difficult.

xz-dev:
> LibreTube is not a commercial project, so its contributors shouldn't aim to make money with it either.

yeah, I agree with you, I love to code opensource APP.
But I'm really close to not being able to pay my rent this month, and due to some personal issues, I can't go to work for a while, which is difficult.

Bnyro:
Tested, can confirm it fixes the issue!
I found one other issue that the comments sheet are still visible when going into fullscreen mode, I'm not sure if we should hide the comments in that case or not. Though doesn't block the PR from being merged.

NeeRaj-2401:
> I'm not sure if we should hide the comments in that case or not.

I think just leave that as it is 😂. That leads us to read comments in full screen something like official YT app.

Bnyro (approved):

xz-dev:
> > But I'm really close to not being able to pay my rent this month, and due to some personal issues, I can't go to work for a while, which is difficult.
> 
> I'm very sorry to hear that, however I doubt that donations will be the proper way to get out of that situation. Anyways, best of luck dealing with it! <3



> > But I'm really close to not being able to pay my rent this month, and due to some personal issues, I can't go to work for a while, which is difficult.
> 
> I'm very sorry to hear that, however I doubt that donations will be the proper way to get out of that situation. Anyways, best of luck dealing with it! <3

Sorry, I just had a bout of depression, I slept through it and I'm much better now. :smile_cat:

Bnyro:
No worries, there's nothing to excuse for.

Isira-Seneviratne (approved):
This looks good to me.

PR #3715, opened 2023-05-05
Improve stream notifications

Isira-Seneviratne:
* Use the stream/channel IDs as unique notification identifiers to avoid having to keep track of the active notification count.
* Avoid creating intermediate lists when filtering the new streams for which notifications should be created.


Bnyro (approved):
Looks much better than the previous approach, awesome!

PR #3718, opened 2023-05-05
ChannelFragment NPE fix

jatzuk:
Hi there!
Newcomer here =)
Was first time using your app, all was great except a [runtime error](https://drive.google.com/file/d/1FGallGO954fwpWqRmLzHBcNVAl18Lt9p/view?usp=sharing). 
Great that you have crash handling so i can see an error.
Actually it will be nice to separate domain/network code from fragment but I preferred to stick with the existing codebase.


Bnyro:
Thanks for your PR!
I agree, we should have started to separate the data/network logic and the UI logic, however now it'd need to be changed on a lot of places so quite some work.

About the PR: I don't really see a reason why `null` could be returned at API side, but it surely makes sense to handle it properly.

Bnyro (approved):

PR #3762, opened 2023-05-12
Improve notification creation performance

Isira-Seneviratne:
* Display stream notifications all at once instead of one by one, and create the notification objects concurrently to improve performance.
* Avoid creating single notifications, as there is the possibility of summary notifications not being created in later checks.

### DIFF
--- a/app/src/main/java/com/github/libretube/workers/NotificationWorker.kt
+++ b/app/src/main/java/com/github/libretube/workers/NotificationWorker.kt
@@ -25,11 +25,11 @@ import com.github.libretube.helpers.ImageHelper
 import com.github.libretube.helpers.PreferenceHelper
 import com.github.libretube.ui.activities.MainActivity
 import com.github.libretube.ui.views.TimePickerPreference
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/github/libretube/workers/NotificationWorker.kt
```diff
@@ -125,47 +128,46 @@ class NotificationWorker(appContext: Context, parameters: WorkerParameters) :
...
         }
     }
 
-    private suspend fun showStreamNotification(
+    private suspend fun createStreamNotification(
         group: String,
         stream: StreamItem,
-        isSingleNotification: Boolean,
-    ) {
+    ): Triple<Int, Long?, Notification> {
```
Bnyro, line 170: I think returning a `Triple` is not really easy to read as the types `Int` and `Long` are very generic and don't really indicate what they are.
It might be a good idea to add comments about what all the returned values are or create a simple, private data class to become returned here.

Bnyro (approved):

PR #3785, opened 2023-05-15
refactor(UI): increase scrubber size

154pinkchairs:
Currently the seek button is frustratingly small and even with smaller hands one has to tap the screen a few times before being able to actually use it. This increases the button size by 1.5 factor.

### DIFF
--- a/app/src/main/res/layout/exo_styled_player_control_view.xml
+++ b/app/src/main/res/layout/exo_styled_player_control_view.xml
@@ -335,8 +335,8 @@
                 app:bar_height="2dp"
                 app:played_color="?attr/colorSecondary"
                 app:scrubber_color="?attr/colorSecondary"
-                app:scrubber_dragged_size="16dp"
-                app:scrubber_enabled_size="12dp" />
+                app:scrubber_dragged_size="24dp"
+                app:scrubber_enabled_size="18dp" />
 
         </LinearLayout>
 


Bnyro:
Please share a screenshot of the change, I'd like to see how it looks in proportion to the other components. Apart from that this change probably also requires changes in the `MarkableSeekbar`.

terminal-cs:
I've been frustrated by this a few times, is there a sort-of material-you themed scrubber widget? The thicker nature of material you would work for better targeting and would fit the app theme more

154pinkchairs:
@Bnyro 
Old portrait:
image
Old landscape:
image
New portrait:
image
New landscape:
image

So it seems like padding needs to be increased as well. Aside from that this looks accessible enough.

Device: Pixel 6, 502 dp smallest width

Bnyro:
From my testings, the large scrubber looks very ugly compared to the time bar and some parts of the scrubber are cut off.
Hence, we're not going to increase the scrubber size for now.
Thanks for your PR and efforts anyways!

terminal-cs:
> From my testings, the large scrubber looks very ugly compared to the time bar and some parts of the scrubber are cut off. Hence, we're not going to increase the scrubber size for now. Thanks for your PR and efforts anyways!

would it be possible to just increase the touch area of the scrubber, and not the visual size itself? Like just increase it's hitbox to be as tall as the circle marker is. I feel that would make sense

PR #3798, opened 2023-05-20
support import/export for freetube subscriptions

### LINKED ISSUE #3677: support import from Freetube exports: subscriptions, history, playlists

Currently importing from Freetube data is not supported, although Freetube can export subscription for Newpipe format and then allows LibreTube to import such subscription successfully as a workaround.

There are no workaround for the other 2 feature for importing Freetube data: history and playlists.

To add support for these 3 features, I propose to follow the similar UI flow from FreeTube:
1. Prompt the user which format to import from (so that we can build a more deterministic parsing logic for the data format. This way we can prevent & isolate potential issues in the future if Newpipe or Freetube decided to update their data format)
2. Trim unnecessary data (such as `"_id": "xcmrX4RAO1IOj2Mr"` from FreeTube's playlists. Tbh, idk what this id do for the Freetube client, if anyone knows more, pls provide context)
3. Persist data into LibreTube client

### Other details

related to discussion from #3675 

### Acknowledgements

### Tasks

- [x] create enum to encapsulate types of supported clients
- [x] impl alert dialog on ui                                                    
- [ ] translate "Import from"                                                    
- [x] add serializable classes for freetube, libretube, newpipe                 
- [ ] test parsing logic

yoguut:
## Description

Add import & export support for Freetube & Youtube subscription file format.

## Issue

closes #3677


app/src/main/java/com/github/libretube/obj/FreetubeSubscriptions.kt
```diff
@@ -0,0 +1,11 @@
+package com.github.libretube.obj
+
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+@Serializable
+data class FreetubeSubscriptions (
+    @SerialName("_id") val id: String = "",
+    @SerialName("name") val name: String = "",
```
Bnyro, line 9: @SerialName("name") is not needed here.

Bnyro:
There's no point of introducing a `LibreTube` subscriptions format since it's exactly the same as the one from NewPipe, please just remove that.

app/src/main/java/com/github/libretube/enums/SupportedClient.kt
```diff
@@ -0,0 +1,11 @@
+package com.github.libretube.enums
+
+enum class SupportedClient(val value: Int) {
```
Bnyro, line 3: Please call that enum `ImportFormat`.

app/src/main/java/com/github/libretube/helpers/ImportHelper.kt
```diff
@@ -45,15 +50,34 @@ object ImportHelper {
      * Get a list of channel IDs from a file [Uri]
      */
     @OptIn(ExperimentalSerializationApi::class)
-    private fun getChannelsFromUri(activity: Activity, uri: Uri): List<String> {
+    private fun getChannelsFromUri(activity: Activity, uri: Uri, supportedClient: SupportedClient): List<String> {
         return when (val fileType = activity.contentResolver.getType(uri)) {
```
Bnyro, line 54: Currently, when the user imports a CSV file from YouTube they still get prompted to choose a format (e.g. FreeTube, NewPipe) although that won't actually be used.
I think it makes sense here to remove that file type conditions and only allow the user to select the format they want to import. Hence, please also add a `YouTube` entry to the enum class containing the different available formats.

app/src/main/java/com/github/libretube/helpers/ImportHelper.kt
```diff
@@ -45,15 +50,34 @@ object ImportHelper {
...
             "application/json", "application/*", "application/octet-stream" -> {
-                // NewPipe subscriptions format
-                val subscriptions = activity.contentResolver.openInputStream(uri)?.use {
-                    JsonHelper.json.decodeFromStream<NewPipeSubscriptions>(it)
-                }
-                subscriptions?.subscriptions.orEmpty().map {
-                    it.url.replace("https://www.youtube.com/channel/", "")
+                when (supportedClient) {
```
Bnyro, line 56: Currently, when the user imports a CSV file from YouTube they still get prompted to choose a format (e.g. FreeTube, NewPipe) although that won't actually be used.
I think it makes sense here to remove that file type conditions above and only allow the user to select the format they want to import. Hence, please also add a `YouTube` entry to the enum class containing the different available formats.

app/src/main/java/com/github/libretube/obj/FreetubeSubscription.kt
```diff
@@ -0,0 +1,11 @@
+package com.github.libretube.obj
+
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+@Serializable
+data class FreetubeSubscription(
+    @SerialName("name") val name: String,
```
Bnyro, line 8: @SerialName("name") is not needed here as the variable name is already the same as the serial name.

app/src/main/java/com/github/libretube/ui/preferences/BackupRestoreSettings.kt
```diff
@@ -77,18 +87,54 @@ class BackupRestoreSettings : BasePreferenceFragment() {
...
+
     override fun onCreatePreferences(savedInstanceState: Bundle?, rootKey: String?) {
         setPreferencesFromResource(R.xml.import_export_settings, rootKey)
 
         val importSubscriptions = findPreference<Preference>("import_subscriptions")
         importSubscriptions?.setOnPreferenceClickListener {
-            getSubscriptionsFile.launch("*/*")
+            val importOptions = SupportedClient.values().toList()
+                .map { it.toString().lowercase().replaceFirstChar(Char::titlecase) }
+            createDialog("Import subscriptions from", importOptions) {
```
Bnyro, line 115: Please make that string translatable and add it to `values.xml`

app/src/main/java/com/github/libretube/ui/preferences/BackupRestoreSettings.kt
```diff
@@ -77,18 +87,54 @@ class BackupRestoreSettings : BasePreferenceFragment() {
...
+                try {
```
Bnyro, line 116: That try catch isn't needed since we only allow the user to pick available options.

app/src/main/java/com/github/libretube/ui/preferences/BackupRestoreSettings.kt
```diff
@@ -77,18 +87,54 @@ class BackupRestoreSettings : BasePreferenceFragment() {
...
             true
         }
 
         val exportSubscriptions = findPreference<Preference>("export_subscriptions")
         exportSubscriptions?.setOnPreferenceClickListener {
-            createSubscriptionsFile.launch("subscriptions.json")
+            val exportOptions = SupportedClient.values().toList()
+                .map { it.toString().lowercase().replaceFirstChar(Char::titlecase) }
+            createDialog("Export subscriptions to", exportOptions) {
+                try {
```
Bnyro, line 131: (same as above)
That try catch isn't needed since we only allow the user to pick available options.

app/src/main/java/com/github/libretube/ui/preferences/BackupRestoreSettings.kt
```diff
@@ -77,18 +87,54 @@ class BackupRestoreSettings : BasePreferenceFragment() {
         }
     }
 
+    private fun createDialog(
```
Bnyro, line 90: Please move the function content (dialog logic) into the `onClickListener` below where it's called. We only use it once and it's only a very simple function, so there's no need to have it in a separate function.

app/src/main/java/com/github/libretube/ui/preferences/BackupRestoreSettings.kt
```diff
@@ -77,18 +87,54 @@ class BackupRestoreSettings : BasePreferenceFragment() {
...
+        var selectedIndex = 0
+        MaterialAlertDialogBuilder(this.requireContext())
+            .setTitle(title)
+            .setSingleChoiceItems(items.toTypedArray(), selectedIndex) { _, i ->
+                selectedIndex = i
+            }
+            .setPositiveButton(
+                R.string.okay
+            ) { _, _ -> onConfirm(selectedIndex) }
+            .setNegativeButton(R.string.cancel) { _, _ -> }
```
Bnyro, line 104: Please instead use `setNegativeButton(R.string.cancel, null)` here.

app/src/main/java/com/github/libretube/helpers/ImportHelper.kt
```diff
@@ -45,26 +48,43 @@ object ImportHelper {
      * Get a list of channel IDs from a file [Uri]
      */
     @OptIn(ExperimentalSerializationApi::class)
-    private fun getChannelsFromUri(activity: Activity, uri: Uri): List<String> {
+    private fun getChannelsFromUri(activity: Activity, uri: Uri, importFormat: ImportFormat): List<String> {
         return when (val fileType = activity.contentResolver.getType(uri)) {
             "application/json", "application/*", "application/octet-stream" -> {
```
Bnyro, line 53: I think we should just drop that lines entirely (the file type check) and just act depending on the selected import format by the user only.

app/src/main/java/com/github/libretube/helpers/ImportHelper.kt
```diff
@@ -74,21 +94,37 @@ object ImportHelper {
...
+                val freeTubeChannels = subs.map {
+                    FreetubeSubscription(it.name, "", "https://www.youtube.com${it.url}")
+                }
+                val freeTubeSubscriptions = FreetubeSubscriptions(subscriptions = freeTubeChannels)
+                activity.contentResolver.openOutputStream(uri)?.use {
+                    JsonHelper.json.encodeToStream(freeTubeSubscriptions, it)
+                }
+            }
+
+            ImportFormat.YOUTUBECSV -> Unit
```
Bnyro, line 127:
```suggestion
            ImportFormat.YOUTUBECSV -> throw IllegalArgumentException()
```

app/src/main/java/com/github/libretube/ui/preferences/BackupRestoreSettings.kt
```diff
@@ -47,28 +50,33 @@ class BackupRestoreSettings : BasePreferenceFragment() {
     ) {
         it?.let {
             lifecycleScope.launch(Dispatchers.IO) {
-                ImportHelper.importSubscriptions(requireActivity(), it)
+                importFormat.let { format ->
+                    ImportHelper.importSubscriptions(requireActivity(), it, format)
+                }
```
Bnyro, line 55:
```suggestion
               ImportHelper.importSubscriptions(requireActivity(), it, importFormat)
```

app/src/main/java/com/github/libretube/ui/preferences/BackupRestoreSettings.kt
```diff
@@ -77,18 +85,56 @@ class BackupRestoreSettings : BasePreferenceFragment() {
         }
     }
 
+    private fun createDialog(
```
Bnyro, line 88:
```suggestion
    private fun createImportFormatDialog(
```

app/src/main/java/com/github/libretube/ui/preferences/BackupRestoreSettings.kt
```diff
@@ -77,18 +85,56 @@ class BackupRestoreSettings : BasePreferenceFragment() {
...
+                selectedIndex = i
+            }
+            .setPositiveButton(
+                R.string.okay
+            ) { _, _ -> onConfirm(selectedIndex) }
+            .setNegativeButton(R.string.cancel, null)
+            .show()
+    }
+
+    private fun importFormatList(): List<String> {
```
Bnyro, line 106:
```suggestion
    private fun getTranslatedImportFormat(): List<String> {
```

app/src/main/java/com/github/libretube/enums/SupportedClient.kt
```diff
@@ -0,0 +1,11 @@
+package com.github.libretube.enums
+
+enum class ImportFormat(val value: Int) {
```
Bnyro, line 3: I think we could save the string resource ids here, so
```kotlin
enum class ImportFormat(@StringRes val stringRes: Int) {
    NEWPIPE(R.string.import_format_newpipe)
```
etc

app/src/main/java/com/github/libretube/enums/SupportedClient.kt
```diff
@@ -0,0 +1,11 @@
...
+    NEWPIPE(0),
+    FREETUBE(1),
+    YOUTUBECSV(2);
+
+    companion object {
+        fun fromInt(value: Int) = ImportFormat.values().first { it.value == value }
```
Bnyro, line 9: I think that should be dropped, and instead we should use
```kotlin
ImportFormat.values()[value]
```
which should just work the same way.

app/src/main/java/com/github/libretube/ui/preferences/BackupRestoreSettings.kt
```diff
@@ -77,18 +85,56 @@ class BackupRestoreSettings : BasePreferenceFragment() {
...
+                ImportFormat.NEWPIPE -> getString(R.string.import_format_newpipe)
+                ImportFormat.FREETUBE -> getString(R.string.import_format_freetube)
+                ImportFormat.YOUTUBECSV -> getString(R.string.import_format_youtube_csv)
+            }
+        }
+    }
+
+    private fun exportFormatList(): List<String> {
+        return listOf(getString(R.string.import_format_newpipe), getString(R.string.import_format_freetube))
+    }
```
Bnyro, line 118: These functions should be removed in favor of the approach of having the string resource directly inside the enum (see my suggestion in the `SupportedClient.kt` file (which should be renamed to `ImportFormat.kt` by the way).

app/src/main/java/com/github/libretube/ui/preferences/BackupRestoreSettings.kt
```diff
@@ -77,18 +85,56 @@ class BackupRestoreSettings : BasePreferenceFragment() {
         }
     }
 
+    private fun createDialog(
+        titleStringId: Int,
```
Bnyro, line 89:
```suggestion
        @StringRes titleStringId: Int,
```

Bnyro:
Only some minor nitpicks left, thanks for already applying the other suggestions!

app/src/main/java/com/github/libretube/ui/preferences/BackupRestoreSettings.kt
```diff
@@ -77,18 +83,56 @@ class BackupRestoreSettings : BasePreferenceFragment() {
...
+                ImportFormat.NEWPIPE -> getString(R.string.import_format_newpipe)
+                ImportFormat.FREETUBE -> getString(R.string.import_format_freetube)
+                ImportFormat.YOUTUBECSV -> getString(R.string.import_format_youtube_csv)
+            }
+        }
+    }
+
+    private fun exportFormatList(): List<String> {
+        return listOf(getString(R.string.import_format_newpipe), getString(R.string.import_format_freetube))
+    }
```
Bnyro, line 116:
```suggestion
    private val importFormatList = listOf(ImportFormat.NEWPIPE, ImportFormat.FREETUBE, ImportFormat.YOUTUBECSV).map { getString(it.value) }

    private val exportFormatList = listOf(ImportFormat.NEWPIPE, ImportFormat.FREETUBE).map { getString(it.value) }
```

PR #3810, opened 2023-05-22
Avoid possible PIP-related exception on API levels below 33.

Isira-Seneviratne:
The [`PictureInPictureParams#getActions`](https://developer.android.com/reference/android/app/PictureInPictureParams#getActions()) method is only available on API 33 and later, so an error would be thrown on earlier Android versions.


Bnyro:
Oh, that's my bad, I only added that code for testing. It should be removed totally!

FireMasterK:
Isn't this change already made in the Media3 migration PR? 🤔

Bnyro:
No, PiP is not related to Media3 in that manner.

Bnyro (approved):

Bnyro:
Awesome that you found out about that before the release @Isira-Seneviratne, otherwise that could have gone badly.

Issue #3816, opened 2023-05-22
support import from Freetube exports: playlists


yoguut:
### Describe your suggested feature

When user go to `Settings -> Backup & Restore -> Import Playlists`, it should be prompted with options to choose for [NewPipe, FreeTube, YouTube]

When user go to `Settings -> Backup & Restore -> Export Playlists`, it should be prompted with options to choose for [NewPipe, FreeTube] 

### Other details

> Currently, if user tries to import playlists of different formats, it will crash the app.

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

yoguut:
I'd like to work on this issue pls :)

PR #3821, opened 2023-05-24
support freetube playlists import/export

yoguut:
Closes https://github.com/libre-tube/LibreTube/issues/3816

[x] tested import FreeTube playlists to app
[x] tested export Piped / LibreTube playlists from app
[x] tested export FreeTube playlists from app

### DIFF
--- a/app/src/main/java/com/github/libretube/api/PlaylistsHelper.kt
+++ b/app/src/main/java/com/github/libretube/api/PlaylistsHelper.kt
@@ -12,7 +12,9 @@ import com.github.libretube.enums.PlaylistType
 import com.github.libretube.extensions.toID
 import com.github.libretube.helpers.PreferenceHelper
 import com.github.libretube.helpers.ProxyHelper
-import com.github.libretube.obj.ImportPlaylist
+import com.github.libretube.obj.FreeTubeImportPlaylist
+import com.github.libretube.obj.FreeTubeVideo
+import com.github.libretube.obj.PipedImportPlaylist
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.async
 import kotlinx.coroutines.awaitAll
@@ -133,7 +135,41 @@ object PlaylistsHelper {
         }
     }
 
-    suspend fun importPlaylists(playlists: List<ImportPlaylist>) = withContext(Dispatchers.IO) {
+    suspend fun importFreeTubePlaylists(playlists: List<FreeTubeImportPlaylist>) = withContext(Dispatchers.IO) {
+        playlists.map { playlist ->
+            val playlistId = createPlaylist(playlist.name)
+            async {
+                playlistId ?: return@async
+                // if logged in, add the playlists by their ID via an api call
+                if (loggedIn) {
+                    addToPlaylist(
+                        playlistId,
+                        *playlist.videos.map {
+                            StreamItem(url = "$YOUTUBE_FRONTEND_URL/watch?v=${it.videoId}")
+                        }.toTypedArray(),
+                    )
+                } else {
+                    // if not logged in, all video information needs to become fetched manually
+                    // Only do so with 20 videos at once to prevent performance issues
+                    playlist.videos.mapIndexed { index, id -> id to index }
+                        .groupBy { it.second % 20 }.forEach { (_, videos) ->
+                            videos.map {
+                                async {
+                                    runCatching {
+                                        val stream = RetrofitInstance.api.getStreams(it.first.videoId).toStreamItem(
+                                            it.first.videoId
+                                        )
+                                        addToPlaylist(playlistId, stream)
+                                    }
+                                }
+                            }.awaitAll()
+                        }
+                }
+            }
+        }.awaitAll()
+    }
+
+    suspend fun importNewPipePlaylists(playlists: List<PipedImportPlaylist>) = withContext(Dispatchers.IO) {
         playlists.map { playlist ->
             val playlistId = createPlaylist(playlist.name!!)
             async {
@@ -167,7 +203,7 @@ object PlaylistsHelper {
         }.awaitAll()
     }
 
-    suspend fun exportPlaylists(): List<ImportPlaylist> = withContext(Dispatchers.IO) {
+    suspend fun exportPipedPlaylists(): List<PipedImportPlaylist> = withContext(Dispatchers.IO) {
         getPlaylists()
             .map { async { getPlaylist(it.id!!) } }
             .awaitAll()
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/github/libretube/api/PlaylistsHelper.kt
```diff
@@ -167,18 +203,33 @@ object PlaylistsHelper {
...
     }
 
-    suspend fun exportPlaylists(): List<ImportPlaylist> = withContext(Dispatchers.IO) {
+    suspend fun exportPipedPlaylists(): List<PipedImportPlaylist> = withContext(Dispatchers.IO) {
         getPlaylists()
             .map { async { getPlaylist(it.id!!) } }
             .awaitAll()
             .map {
                 val videos = it.relatedStreams.map { item ->
                     "$YOUTUBE_FRONTEND_URL/watch?v=${item.url!!.toID()}"
```
yoguut, line 212: Question to reviewer: why force unwrap optional `url`?

Bnyro:
Maybe we can rename the string of the Piped option to "Piped / LibreTube"? Otherwise users that exported it earlier probably don't know which format they exported it to.

app/src/main/java/com/github/libretube/api/PlaylistsHelper.kt
Bnyro: `url` is nullable and `toID()` only works with non-null strings. So we either need to use `!!` or `?`. Which one we use doesn't matter, because it should never become null. But asserting it as non-null is better in this case because `?` indicates it can be null.

app/src/main/java/com/github/libretube/api/PlaylistsHelper.kt
```diff
@@ -133,7 +135,41 @@ object PlaylistsHelper {
         }
     }
 
-    suspend fun importPlaylists(playlists: List<ImportPlaylist>) = withContext(Dispatchers.IO) {
+    suspend fun importFreeTubePlaylists(playlists: List<FreeTubeImportPlaylist>) = withContext(Dispatchers.IO) {
```
Bnyro, line 138: That's just a lot of duplicated code. Instead, please keep the old `importPlaylists` function and call it no matter if it's a Piped or a FreeTube playlist that's being imported.
That means we should "convert" the `List<FreeTubeImportPlaylist` to a `List<PipedImportPlaylist>` using 
```kotlin
freeTubePlaylists.map { PipedImportPlaylist(it.title, ...) }
```
(the above is just a simple example).

app/src/main/java/com/github/libretube/helpers/ImportHelper.kt
```diff
@@ -144,47 +189,49 @@ object ImportHelper {
...
+                }
+                try {
+                    PlaylistsHelper.importNewPipePlaylists(importPlaylists)
+                    activity.toastFromMainDispatcher(R.string.success)
+                } catch (e: Exception) {
+                    Log.e(TAG(), e.toString())
+                    e.localizedMessage?.let {
+                        activity.toastFromMainDispatcher(it)
+                    }
+                }
```
Bnyro, line 205: That try catch block is being called three times here (for each playlist type). Instead, it should only be there once after the `when` clause (which will work after applying my suggestion from `PlaylistsHelper.kt`.

app/src/main/java/com/github/libretube/helpers/ImportHelper.kt
```diff
@@ -76,6 +77,8 @@ object ImportHelper {
                     }
                 }.orEmpty()
             }
+            // Piped case not use for subscriptions
+            ImportFormat.PIPED -> emptyList()
```
Bnyro, line 81:
```suggestion
            else -> throw IllegalArgumentException()
```

app/src/main/java/com/github/libretube/helpers/ImportHelper.kt
```diff
@@ -123,12 +126,38 @@ object ImportHelper {
...
+                importPlaylists.forEach { playlist ->
+                    playlist.videos = playlist.videos.map { it.takeLast(11) }
+                }
+            }
+            ImportFormat.FREETUBE -> {
+                val playlistFile = activity.contentResolver.openInputStream(uri)?.use {
+                    JsonHelper.json.decodeFromStream<List<FreeTubeImportPlaylist>>(it)
+                }
+                val playlists = playlistFile?.map { playlist ->
+                    // convert FreeTube videos to listOf string
```
Bnyro, line 149:
```suggestion
                    // convert FreeTube videos to list of string
```

app/src/main/java/com/github/libretube/helpers/ImportHelper.kt
```diff
@@ -144,23 +173,13 @@ object ImportHelper {
...
-                val message = activity.getString(R.string.unsupported_file_format, fileType)
-                activity.toastFromMainDispatcher(message)
-                return
             }
-        }
-
-        // convert the YouTube URLs to videoIds
-        importPlaylists.forEach { playlist ->
-            playlist.videos = playlist.videos.map { it.takeLast(11) }
+            ImportFormat.NEWPIPE -> Unit
```
Bnyro, line 182:
```suggestion
            else -> throw IllegalArguemntException()
```

app/src/main/java/com/github/libretube/api/PlaylistsHelper.kt
```diff
@@ -133,52 +139,69 @@ object PlaylistsHelper {
...
-                    playlist.videos.mapIndexed { index, id -> id to index }
-                        .groupBy { it.second % 20 }.forEach { (_, videos) ->
-                            videos.map {
-                                async {
-                                    runCatching {
-                                        val stream = RetrofitInstance.api.getStreams(it.first).toStreamItem(
-                                            it.first,
-                                        )
-                                        addToPlaylist(playlistId, stream)
+    suspend fun importPlaylists(playlists: List<PipedImportPlaylist>) =
```
Bnyro, line 142: Please try to undo these formatting changes since they are not related to your PR really

Bnyro:
Only some minor things from my side, looks good to me otherwise!

Bnyro (approved):

Bnyro:
Have you tested whether it works well after the latest changes?

yoguut:
> Have you tested whether it works well after the latest changes?

_Yes and kind of. I don't have Piped account or Youtube account. So my playlists come from FreeTube export. And then once imported the FreeTube playlists in the app, I exported it to Piped & FreeTube formats._

_It would be great if someone has either an Piped account or Youtube account to help testing by exporting playlists, then importing into the app to ensure the most current formats of these 2 services are consistent to FreeTube's export formats._


[x] tested import FreeTube playlists to app
[x] tested export Piped / LibreTube playlists from app
[x] tested export FreeTube playlists from app

Bnyro:
Tested the Piped format and the YouTube CSV and works as expected. Thanks for your PR!

PR #3831, opened 2023-05-25
Fix search bar collapse while search in progress

### LINKED ISSUE #1965: Search Text Controller Disposed When Playing/Dismissing A Video
### Steps to reproduce

1. Search for something
2. Play the video (playing and clicking back btn would cause the issue) 
3. Close the video (dismissing a video cause the same issue) 

### Expected behavior

Nothing should affect the text controller 

### Actual behavior

Search controller is disposed  

### LibreTube version

0.8.0

### Android version

12

### Other details

Playing or dismissing a video would clear the search text controller and close the search bar. 

https://user-images.githubusercontent.com/50032603/203103233-63f6f03b-839c-487b-a089-f749ca494a20.mp4

### Acknowledgements

SajalRG:
This is an AOSP bug where destroy of fragment calls invalideMenu of Activity even though the fragment does not participate in menu updates.

https://issuetracker.google.com/issues/244336571

Hence added an workaroud to not invalidate menu when search is in progress.

Closes #1965

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/activities/MainActivity.kt
+++ b/app/src/main/java/com/github/libretube/ui/activities/MainActivity.kt
@@ -280,6 +280,22 @@ class MainActivity : BaseActivity() {
         return super.onPrepareOptionsMenu(menu)
     }
 
+    private fun isSearchInProgress(): Boolean {
+        if (!::navController.isInitialized) return false
+        val id = navController.currentDestination?.id ?: return false
+        return id in listOf(R.id.searchFragment, R.id.searchResultFragment)
+    }
+
+    override fun invalidateMenu() {
+        // Don't invalidate menu when in search in progress
+        // this is a workaround as there is bug in android code
+        // details of bug: https://issuetracker.google.com/issues/244336571
+        if (isSearchInProgress()) {
+            return
+        }
+        super.invalidateMenu()
+    }
+
     override fun onCreateOptionsMenu(menu: Menu): Boolean {
         // Inflate the menu; this adds items to the action bar if it is present.
         menuInflater.inflate(R.menu.action_bar, menu)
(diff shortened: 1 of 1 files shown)


SajalRG:
Verified all cases below working as expected
1. Search then open player and close player
2. Search then open player and minimize player with back key
3. Close player while search in progress and player minimized
4. Switch videos from search release and then press back key
5. Switch in bottom navigation while search in progress
6. Open player from search, switch to audio mode and then press back

FireMasterK:
https://issuetracker.google.com/issues/244336571#comment22

Could this issue also be fixed if we just include the latest version of the dependency in LibreTube directly?

app/src/main/java/com/github/libretube/ui/activities/MainActivity.kt
```diff
@@ -280,6 +280,22 @@ class MainActivity : BaseActivity() {
         return super.onPrepareOptionsMenu(menu)
     }
 
+    private fun isSearchInProgress(): Boolean {
+        if (!::navController.isInitialized) return false
+        val id = navController.currentDestination?.id ?: return false
+        return ((id == R.id.searchFragment) || (id == R.id.searchResultFragment))
```
Bnyro, line 286: I'd prefer something like
```suggestion
        return  id in listOf(R.id.searchFragment, R.id.searchResultFragment)
```
for better readability.

Bnyro:
Tested it and can confirm this fixes the issue, awesome work, happy we can finally close this issue!

app/src/main/java/com/github/libretube/ui/activities/MainActivity.kt
SajalRG: Thanks, will update same commit with your suggestion.

SajalRG:
Hi @Bnyro , updated commit as per your suggestion. Please review. Thanks :)

Bnyro (approved):

PR #3878, opened 2023-06-01
Use DASH in the audio only/background player

Bnyro:
This significantly improves the loading times when playing audio in the background.
Note that this change requires us to disable the audio quality preference for now, it can be re-enabled later (when we added some logic to select audio tracks in the audio player mode) theoretically.


buawf:
> disable the audio quality preference 
 
 Where is the preference to disable ? 
 
 Currently using v0.15.0

PR #3915, opened 2023-06-03
Welcome Activity to choose instance on first app startup

Bnyro:
## Why

Previously, kavin.rocks has always been selected as the default instance when installing the app.
This behavior now changes, so that users can choose the instance they want to at the first app startup, without any being deselected. This should help us to spread users across more instances, which could reduce the costs for the official instance and also if the official instance was down, not everybody would affected (decentralization, ...).

## Testers needed

Since I'm not 100% sure if the current UI approach is the best one for new users (and self-explaining!!!), I need some people to please, please test it. It can be found after clearing the app's data and then opening the app (like after a clean installation).
CI builds are [here](https://github.com/libre-tube/LibreTube/actions/runs/5165317075).

ping @SameenAhnaf, @rozari0, @luckkmaxx, @xIUPITERx, @akano12, @FireMasterK, @Isira-Seneviratne, @ARBoyGo


luckkmaxx:
Sure! I will test and coment asap the commit is merged.

Bnyro:
You can already test the apk at the bottom here: https://github.com/libre-tube/LibreTube/actions/runs/5165317075. (It's called "app" and only visible when you're logged in). Then you just need to unzip it and install the version for the architecture you need.

Bnyro:
You need to uninstall LibreTube Debug before, they're not signed with the same key.

luckkmaxx:
Ok tested. It works, and have suggestions..:
1. Since the list is too long, maybe an indicator or whatever resources material design gives to indicate the list continues.
2. Flags could be explained above, that this is the server locations (not languages)
3. It could be nice to set a custom instance from there.

Bnyro:
That's the kind of feedback I've been looking for!

I agree with 1, however I'm not really sure what's the proper component for indicating that (if there is any :thinking:).

I guess for 2. that's a good point, we should probably add that to the text above the instances?

That seems reasonable, however we should watch out that new users don't get confused by it, so I'm not sure if we just create a small button in a corner called sth like `Custom (Advanced users)` or just keep it in the settings only. I don't think it's really hard for "advanced" users that are familiar with the app to quickly head over to the settings.

luckkmaxx:
I edited the above comment with some tips.
Regarding the custom instance, thats ok if you decide to not complicate other users. However, the way could be another button near the "ok" button called "add custom", who will leave you in instance settings with the window as if you were tapped in "add custom".

luckkmaxx:
> I agree with 1, however I'm not really sure what's the proper component for indicating that (if there is any 🤔).

Maybe a slider in the right?

luckkmaxx:
BTW, while adding a custom instance, there could be a checkbox to set it as current.

luckkmaxx:
> > Maybe a slider at the right of screen?
> 
> You mean a scrollbar? ([Slider](https://m3.material.io/components/sliders/overview)) I'm not sure, if they are part of material you.
> 
> The only thing I saw in the guidelines is a progress indicator, but I'm not sure if it fits here.

Yes, scrollbar! sorry LOL
Not sure neither if matches MD3, but there are apps (like MDGram) using scrollbars.

ARBoyGo:
Tested, Personally, This is good for new users. As most of them get comfused in a hole about instances. 
As of suggestions, I think, At the end a input to choose custom instances - if you own a Instances, Add it in here.
Custom instances has three input, So it should be something like in our settings. 

A small note, Choose instances based on you are country or nearby country. Flag itself doesn't explain much. 

While, Commenting, Noticed your reply on Custom Instances, Indeed, Only advance users use it, So it's pretty reasonable.

ARBoyGo:
> 2 add an option to sort it by country (if it's not to complex)
> 3 the Guidelines also recommend using a dropdown menu (so you can search your country or server) instead of a list if it's to long

Dropdown menu is good instead of scrolling. Though scrolling is might looking appealing to the users.

luckkmaxx:
In case of hundred instances, a search filter would be comfotable to users who are looking for some.

luckkmaxx:
After opening #3918 I thought about this startup wizard.

Here you could add optional login fields.

app/src/main/java/com/github/libretube/ui/adapters/InstancesAdapter.kt
```diff
@@ -0,0 +1,41 @@
...
+import android.view.ViewGroup
+import androidx.recyclerview.widget.RecyclerView
+import com.github.libretube.api.obj.Instances
+import com.github.libretube.databinding.InstanceRowBinding
+import com.github.libretube.ui.viewholders.InstancesViewHolder
+
+class InstancesAdapter(
+    private val instances: List<Instances>,
+    private val onSelectInstance: (index: Int) -> Unit
+): RecyclerView.Adapter<InstancesViewHolder>() {
```
Isira-Seneviratne, line 14: This could be a [`ListAdapter`](https://developer.android.com/reference/androidx/recyclerview/widget/ListAdapter) instead. It has additional methods, e.g. getItem.

app/src/main/java/com/github/libretube/ui/viewholders/InstancesViewHolder.kt
```diff
@@ -0,0 +1,8 @@
+package com.github.libretube.ui.viewholders
```
Isira-Seneviratne, line 1: The class could be moved to the InstancesAdapter file.

Isira-Seneviratne:
This looks good to me, it would be useful for new users. Added a couple of minor suggestions.

akano12:
The design is neat, I like it.
Regarding the "custom instance" suggestion, i do believe it's unnecessary.

I said unnecessary because you install LT once, it's not like we're uninstalling and installing the app everyday.

Last time i did a fresh install of LT was few months ago, so if i had to set a custom instance (suppose this feature existed from before) i would have used it only once in 4-5 months (+ small portion of LT users rely on custom instances) . So it's unnecessary in my opinion.

In case it's to be implemented, then i suggest a full walkthrough when opening the app where you can set (in addition to custom instance) the region and theme as those 2 settings are changed by like 90% of the users on first app launch.

app/src/main/java/com/github/libretube/ui/viewholders/InstancesViewHolder.kt
Bnyro: We currently have all the view holders for the recycler views in a separate file, so I don't really think we should do it different here? 🤔

Isira-Seneviratne: The Kotlin documentation [recommends](https://kotlinlang.org/docs/coding-conventions.html#class-layout) putting related declarations in the same file.

Bnyro:
>     1. adjust the spacing between `screen border - radio button - label` to look more like the [Guidelines](https://m3.material.io/components/radio-button/guidelines) (make the margins bigger)
> 
Good point, I'll adjust that!
>     2. add an option to sort it by country (if it's not to complex)
> 
I don't see a way to implement that unfortunately, we only have the flags as information but not the names of the countries. So there's no real way to add that, also because some instances are on multiple countries.
>     3. the Guidelines also recommend using a dropdown menu (so you can search your country or server) instead of a list if it's to long
> 
I don't think that would look good in a welcome activity, because there would be too much free space left then.
>     4. maybe add a random server button
>
I fail to see the purpose of this. Selecting a random instance doesn't make sense because you want to pick the one that's the closest to you likely. If you want to select a random instance, just click something random.

app/src/main/java/com/github/libretube/ui/viewholders/InstancesViewHolder.kt
Bnyro: You're right that, however I think that should be done in a separate PR for all RecyclerView adapters.
I guess it makes sense to do it for all view holders in a separate file until then to make it be the same everywhere.

app/src/main/java/com/github/libretube/ui/adapters/InstancesAdapter.kt
Bnyro: I don't think this makes sense here since we don't actually use any method like `insertItems()` or any diff comparisons for the instances list. The instances list is only initialized once and never modified after.

luckkmaxx:
Point 2: sort or filter by country.
Probably adding a database from where instances info would be fetched. Obvoisly filled here by devs or contributors

Bnyro:
> The instances list could be filtered by countries. Geographically near servers have a lower ping. Related to that, instances with a CDN should probably be displayed with a world icon (as a globe).
>
I agree that it could be useful, however I don't really see a way to add a filter menu to this welcome page since it should stay as simple as possible. Also if there are connection issues with `piped-instances.kavin.rocks`, we fallback to an in-app hard-coded list of instances which doesn't contain info about the instance's location, so it would cause issues then.
> To align the flags of the countries in a row, it would maybe look good if the flags would be shown before the selection circle.
>
I thought about that too, but some instances (like the official one) have so many locations, that the instance name might not be on the page anymore on small screen devices.

Bnyro:
Changes with the latest commit:
* Added scrollbars
* Increased padding between label and radio button
* Indicator about CDN as a globe
* Note about the flags and that closer instances to where you live might be faster
* Increased width of radio buttons so that users can click on the right of the screen to select an instance too
* Shortened the message when pressing `Okay` with no instance selected

app/src/main/res/values/strings.xml
```diff
@@ -412,6 +412,10 @@
     <string name="creation_date_reversed">Creation date (reversed)</string>
     <string name="alphabetic">Alphabetic</string>
     <string name="alphabetic_reversed">Alphabetic (reversed)</string>
+    <string name="welcome">Welcome to LibreTube</string>
+    <string name="choose_instance">Please choose an instance first!</string>
+    <string name="choose_instance_long">Please choose a Piped instance to use below. The Piped instance will act as a middle man between you and YouTube. These servers are located at different physical places (indicated with flags). Instances that are close to where you live, are likely faster than others. You can still change it later in the instance settings.</string>
```
Bnyro, line 417: If we have a native English speaker here, it'd be great if they could quickly take a look here!

luckkmaxx:
Nice. this is the new look after commit.
Maybe you need to explain or provide external info about what a CDN is (many users would be confused with this)
Screenshot_20230604-232612_LibreTube Debug

app/src/main/res/values/strings.xml
```diff
@@ -412,6 +412,10 @@
...
+    <string name="registration_disabled">Registration disabled</string>
```
FireMasterK, line 418: Potential improvement.
```suggestion
    <string name="choose_instance_long">Please choose a Piped instance to use from below. The Piped instance will act as the middleman between you and YouTube. These instances are located at different physical locations - indicated by their country flag(s). Instances physically closer to you are likely faster than instances that are not. You can change the instance you use in the settings at any time.</string>
    <string name="registration_disabled">Registration disabled</string>
```
```suggestion
    <string name="choose_instance_long">Please choose a Piped instance to use from below. The Piped instance will act as the middleman between you and YouTube. These instances are located at different physical locations - indicated by their country flag(s). Instances physically closer to you are likely faster than instances that are not. You can change the instance you use in the settings at any time.</string>
    <string name="registration_disabled">Registration disabled</string>
```
```suggestion
    <string name="choose_instance_long">Please choose a Piped instance to use from below. The Piped instance will act as the middleman between you and YouTube. These instances are located at different physical locations - indicated by their country flag(s). Instances physically closer to you are likely faster than instances that are not. You can change the instance you use in the settings at any time.</string>
    <string name="registration_disabled">Registration disabled</string>
```

luckkmaxx: Sounds better.
'- indicated by their country flag(s) -.'
(I dont know in english, but maybe you missed the ending hyphen)

PR #3957, opened 2023-06-08
Significantly increase download speed using range requests

Bnyro:
* This increases the download speed to around 100 times the one we had earlier, that's significant!


Bnyro:
cc @AudricV, finally applying your suggestion after half a year :wink:

PR #4021, opened 2023-06-17
Added ability to choose what categories are skipped automatically and…

### LINKED ISSUE #4020: Allow both automatic and manual skipping with sponsor block

It should allow the user to choose which categories are skipped automatically and which can be manually skipped.

### Other details

_No response_

### Acknowledgements

general-a:
This is the PR for #4020
closes #4020


app/src/main/java/com/github/libretube/helpers/PlayerHelper.kt
```diff
@@ -476,7 +422,7 @@ object PlayerHelper {
     fun ExoPlayer.checkForSegments(
         context: Context,
         segments: List<Segment>,
-        skipManually: Boolean = false,
+        category_config: MutableMap<String, String>,
```
Bnyro, line 425: Please name that var `categoryConfig`

app/src/main/java/com/github/libretube/services/OnlinePlayerService.kt
```diff
@@ -76,6 +76,7 @@ class OnlinePlayerService : LifecycleService() {
      * SponsorBlock Segment data
      */
     private var segments: List<Segment> = listOf()
+    private var catConfig: MutableMap<String, String> = PlayerHelper.getSponsorBlockCategories()
```
Bnyro, line 79: Please name that var `sponsorBlockConfig`

app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
```diff
@@ -181,6 +181,7 @@ class PlayerFragment : Fragment(), OnlinePlayerOptions {
      * SponsorBlock
      */
     private var segments = listOf<Segment>()
+    private var catConfig = PlayerHelper.getSponsorBlockCategories()
```
Bnyro, line 184: Please name that var `sponsorBlockConfig` and init it after `sponsorBlockEnabled` for better readability

app/src/main/java/com/github/libretube/helpers/PlayerHelper.kt
```diff
@@ -69,65 +68,18 @@ object PlayerHelper {
...
+        val categories: MutableMap<String, String> = mutableMapOf()
+
+        for (cat in SPONSOR_CATEGORIES){
+            val state = PreferenceHelper.getString(cat + "_category_key", "Off")
+            if (state != "Off"){
+                categories[cat] = state
+            }
         }
         return categories
+
```
Bnyro, line 82: Please remove that new line

app/src/main/res/values/array.xml
```diff
@@ -422,4 +422,11 @@
         <item>playlists</item>
     </string-array>
 
+    <string-array name="category_sponsor_skip_choices">
+        <item>Off</item>
+        <item>Manual</item>
+        <item>Automatic</item>
+    </string-array>
+
+
```
Bnyro, line 431: The strings are not currently translatable. Please create one string array `sb_skip_options_values`, that contains `off`, `manual` and `automatic` as items (so the current one basically, but lowercased), and a second string array that's called `sb_skip_options` with `@string/all`, `@string/manually` and `@string/off`, and use them respectively in the preferences `entryValues` and `entries`.

Bnyro:
Looks very promising, thanks so far! Just some nitpicks from my side. You don't have to worry about the failing CI at the moment, it's not your PRs fault.

Bnyro:
Please run `git add .` in the projects root directory ;)

app/src/main/res/values/strings.xml
```diff
@@ -85,21 +85,21 @@
     <string name="segment_skipped">Skipped segment</string>
     <string name="category_segments">Segments</string>
     <string name="category_sponsor">Sponsor</string>
-    <string name="category_sponsor_description">Paid promotion, paid referrals and direct advertisements. Not for self-promotion or free genuine shoutouts to causes, creators, websites, and products.</string>
+    <string name="category_sponsor_description">Paid promotion, paid referrals and direct advertisements. Not for self-promotion or free genuine shoutouts to causes, creators, websites, and products: %s</string>
```
general-a, line 88: Is there a better way to do this? Adding %s to all languages seems tedious and inefficient but I wanted a way to indicate what setting was currently selected.

Bnyro: Yes, we can subclass `DropDownPreference` and override `getSummary()`. That's probably way better because using this way, each translator needs to update all the strings too.

Bnyro:
```kotlin
override fun getSummary(): CharSequence {
val currentIndex = entryValues.indexOf(value)
return "${super.getSummary()} • ${entries[currentIndex]}"
}
```

general-a: Where would this code go?

Bnyro:
```kotlin
class SponsorBlockDropDownPreference(context: Context, attrs: AttributeSet): DropDownPreference(context, attrs) {
// insert the cod here
}`
```

Bnyro: And then just use that class in the preference XML files instead of `DropDownPreference`. I didn't test it though, might be that the getSummary function needs some minor modifications

general-a: Do you think we could implement a spinner instead? Kind of like how it was before but instead of the slider we have the spinner with the options. Not entirely sure how to implement it (What I'm not sure about is how to get the xml file to read our custom class but I think thats as simple as adding the class to a certain folder).

Bnyro: I don't really see the purpose of it, wouldn't a spinner preference be very similar to a dropdown preference. And it doesn't matter which directory we add the preference class to, we just have to reference it in the XML with `com.github.libretube.ui.views.SponsorBlockPrefernece` for example.

general-a: So currently the setting looks like this:
Screen Shot 2023-06-17 at 2 31 41 PM
I think something like this would look cleaner:
Screen Shot 2023-06-17 at 2 33 58 PM

What do you think would be best?

Bnyro: Oh yes, I see! The second approach probably looks way better. I think we can just override `onCreate` in our custom preference class and set a different layout via `setContentView` and view bindings, and then just set the texts of the layout (title, summary, current value) manually.

general-a: Alright sounds good. I'll get that going

general-a: So I'm not the best with front end stuff. So from my understanding the way the settings are drawn needs to be changed for our custom preference. Is this done in the custom preference file itself or in the file that processes the xml file?

Bnyro: In the class/file that extends `DropDoenPreference`.

general-a: Do you know of any good resources/ examples on how to modify these functions?

Bnyro: https://overflow.lunar.icu/questions/16108609/android-creating-custom-preference

This looks pretty much what we're doing, just that we override `DropDownPreference` instead of `Preference`. The code in the linked issue is in java, but it works basically the same way in Kotlin.

app/src/main/res/xml/sponsorblock_settings.xml
```diff
@@ -27,7 +27,7 @@
 
     <PreferenceCategory app:title="@string/category_segments">
 
-        <DropDownPreference
+        <com.github.libretube.ui.views.SpinnerPreference
```
general-a, line 30: This doesn't seem to do anything different atm.

app/src/main/java/com/github/libretube/ui/views/SpinnerPreference.kt
```diff
@@ -0,0 +1,17 @@
...
+import android.view.View
+import android.view.ViewGroup
+import androidx.preference.DropDownPreference
+import com.github.libretube.R
+
+class SpinnerPreference(context: Context, attrs: AttributeSet): DropDownPreference(context, attrs){
+
+    fun onCreate(parent: ViewGroup): View {
+        val inflater = context.getSystemService(Context.LAYOUT_INFLATER_SERVICE) as LayoutInflater
+        return inflater.inflate(R.layout.spinner_preference, parent, false)
```
general-a, line 14: Adding this preference in the settings doesn't seem to do anything. Also for some reason the onCreate function do not exist thus does not need to be overridden..

Bnyro: Looks like we need to override `onCreateView` instead. And please inflate it like so:
```kotlin
val binding = SpinnerPreferenceBinding.inflate(LayoutInflater.from(context)
return binding.root
```

general-a: `onCreateView` doesn't seem to exist either :/ Is this something unique to the preference classes? I would assume something like creating the view would be universal across UI elements

Bnyro: I got it working like so:
```kotlin

package com.github.libretube.ui.views
import android.content.Context
import android.util.AttributeSet
import android.view.LayoutInflater
import android.view.View
import android.view.ViewGroup
import androidx.preference.Preference
import com.github.libretube.R

class SpinnerPreference(context: Context, attrs: AttributeSet): Preference(context, attrs) {
    init {
        layoutResource = R.layout.spinner_preference
    }
}
```
So we need to subclass `Preference` now, otherwise it crashes in my tests.

The text has to be set in
```kotlin
override fun onBindViewHolder(holder: PreferenceViewHolder) {}
```
then.

general-a: I have this so far (used chatgpt as a starter) Does this look correct?
`package com.github.libretube.ui.views

import android.content.Context
import android.content.res.TypedArray
import android.util.AttributeSet
import android.view.View
import android.widget.AdapterView
import androidx.preference.Preference
import androidx.preference.PreferenceViewHolder
import android.widget.ArrayAdapter
import android.widget.Spinner
import android.widget.TextView
import com.github.libretube.R

class SpinnerPreference(context: Context, attrs: AttributeSet) : Preference(context, attrs) {
    private lateinit var spinner: Spinner
    private lateinit var adapter: ArrayAdapter<CharSequence>
    private var selectedItem: CharSequence? = null

    init {
        layoutResource = R.layout.spinner_preference
    }

    override fun onBindViewHolder(holder: PreferenceViewHolder) {
        super.onBindViewHolder(holder)
        spinner = holder.itemView.findViewById(R.id.spinner)

        val titleView = holder.itemView.findViewById<TextView>(android.R.id.title) as? TextView
        titleView?.text = super.getTitle()

        val summaryView = holder.itemView.findViewById<TextView>(android.R.id.summary) as? TextView
        summaryView?.text = super.getSummary()

        // Set the spinner adapter
        adapter = ArrayAdapter.createFromResource(
            context,
            R.array.sb_skip_options,
            android.R.layout.simple_spinner_item
        )
        adapter.setDropDownViewResource(android.R.layout.simple_spinner_dropdown_item)
        spinner.adapter = adapter

        // Set the initial selected item
        if (selectedItem != null) {
            val position = adapter.getPosition(selectedItem!!)
            spinner.setSelection(position)
        }

        // Set a listener to handle item selection
        spinner.onItemSelectedListener = object : AdapterView.OnItemSelectedListener {
            override fun onItemSelected(
                parent: AdapterView<*>,
                view: View?,
                position: Int,
                id: Long
            ) {
                selectedItem = adapter.getItem(position)
                persistString(selectedItem?.toString())
            }

            override fun onNothingSelected(parent: AdapterView<*>) {}
        }
    }

    override fun onGetDefaultValue(a: TypedArray, index: Int): Any {
        // Get the default value from the XML attribute, if specified
        return a.getString(index) ?: ""
    }

    override fun onSetInitialValue(defaultValue: Any?) {
        // Set the initial selected item from the persisted value, if available
        selectedItem = getPersistedString(defaultValue?.toString())
    }
}
`

Bnyro: Seems fine

app/src/main/res/layout/spinner_preference.xml
```diff
@@ -1,29 +1,52 @@
...
 
-
-    <TextView
-        android:id="@android:id/title"
-        style="@android:style/TextAppearance.DeviceDefault.SearchResult.Title"
+    <LinearLayout
+        android:id="@+id/left_container"
         android:layout_width="wrap_content"
         android:layout_height="wrap_content"
-        android:text="Title" />
```
general-a: is it possible to copy the section margin for spacing on the left?

app/src/main/res/xml/sponsorblock_settings.xml
```diff
@@ -33,64 +33,64 @@
...
-            app:defaultValue="off"/>
+            app:defaultValue="@string/off"/>
 
-        <DropDownPreference
+        <com.github.libretube.ui.views.SpinnerPreference
             app:key="preview_category_key"
             app:summary="@string/category_preview_description"
             app:title="@string/category_preview"
             app:entries="@array/sb_skip_options"
             app:entryValues="@array/sb_skip_options_values"
```
Bnyro, line 92: `entries` and `entryValues` can be removed everywhere here

app/src/main/res/xml/sponsorblock_settings.xml
```diff
@@ -33,64 +33,64 @@
...
-            app:defaultValue="off"/>
+            app:defaultValue="@string/off"/>
```
Bnyro, line 93: Please don't use translatable strings to store preference values. If the translation is getting changed, the preference value would also change and not work anymore.

general-a: Hmm I see, when I put the string in directly it breaks the code. Do you know why this might be happening?
Screen Shot 2023-06-18 at 4 30 27 PM

Bnyro:
Thank you!

PR #4023, opened 2023-06-17
Update README.md

ghost:
Most people do not know that they can use LibreTube exactly like NewPipe if anything happened to the instances, especially these days when the instances go down more frequently, also a lot of them are using the app just for the UI they don't even know or care about the Piped Proxy.

I hope you accept my humble change to make it clear so people with find a way if anything goes wrong.

Thanks.


Bnyro:
That's not 100% correct. The option only disables using the Piped proxy for loading images and playing videos. However, in order for the app to work (load video information, load subscriptions, search, ...) a Piped instance is still required.

Bnyro:
No worries

PR #4025, opened 2023-06-17
Add PlayerData class

Isira-Seneviratne:
Add `PlayerData` class for storing player data as a parcelable object.


Bnyro (approved):
Nice one!

PR #4053, opened 2023-06-19
Option to fall back to Piped proxy when Piped proxy disabled and video unreachable

### LINKED ISSUE #4038: Fallback to Piped Proxy Option

An option under Audio and Video > Advanced that ties in with the "Disable Piped proxy" option. When "Fallback to Piped proxy" is enabled, LibreTube will first attempt direct connection to YouTube. Then if the playback fails, attempt connecting using the selected Piped instance.

This should allow YouTube Music content to load successfully without needing to manually switch back on the Piped proxy and would provide a good mix of speed to load content and less videos that fail to play back.

### Other details

The option can be greyed out if "Disable Piped proxy" is not enabled and checked off by default after enabling "Disable Piped proxy"

### Acknowledgements

Bnyro:
closes  #4038


app/src/main/java/com/github/libretube/helpers/ProxyHelper.kt
```diff
@@ -10,6 +10,8 @@ import okhttp3.OkHttpClient
 import okhttp3.Request
 
 object ProxyHelper {
+    private val client by lazy { OkHttpClient.Builder().build() }
```
FireMasterK, line 13: Why not use the cronet client here? :thinking: 
We have an adaptor for Cronet's engine for OkHttp

Bnyro:
Is there a video to test this with? For me it appears that YT music videos work as expected even without the Proxy somehow?

ehea617:
I'm currently getting an error at the bottom of the screen when opening any song in the "YT Music Songs" search filter, "Initial player response is not valid"

This error message appears regardless of the song selected or combination of options of piped proxy enabled/disabled and fallback enabled/disabled.

Bnyro:
Try a different instance, see https://github.com/TeamPiped/Piped/issues/2487

Bnyro:
That's not related to the changes in this PR.

ehea617:
Thanks for the info - I can confirm this works great after I switched the instance over to projectsegfau.lt and do not notice any issues specific to this now.

Bnyro:
Alright, thanks for testing!

PR #4055, opened 2023-06-19
Color segments depending on what type they are

### LINKED ISSUE #4054: Sponsor block segments should be colored by type

The desktop extension allows the coloring of segments by type. We should add this to the app.

### Other details

_No response_

### Acknowledgements

general-a:
closes #4054

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/views/MarkableTimeBar.kt
+++ b/app/src/main/java/com/github/libretube/ui/views/MarkableTimeBar.kt
@@ -2,6 +2,7 @@ package com.github.libretube.ui.views
 
 import android.content.Context
 import android.graphics.Canvas
+import android.graphics.Color
 import android.graphics.Paint
 import android.graphics.Rect
 import android.util.AttributeSet
@@ -39,6 +40,17 @@ class MarkableTimeBar(
 
     private fun drawSegments(canvas: Canvas) {
         val markersEnabled = PreferenceHelper.getBoolean(PreferenceKeys.SB_SHOW_MARKERS, true)
+        //TODO Add ability to change these colors via settings
+        val segmentColors = mapOf(
+            "intro" to "#00ffff",
+            "selfpromo" to "#ffff00",
+            "interaction" to "#cc00ff",
+            "sponsor" to "#00d400",
+            "outro" to "#0202ED",
+            "filler" to "#7300ff",
+            "music_offtopic" to "#ff9900",
+            "preview" to "#008fd6"
+        )
         if (player == null || !markersEnabled) return
 
         canvas.save()
@@ -55,10 +67,7 @@ class MarkableTimeBar(
                     canvas.height - marginY,
                 ),
                 Paint().apply {
-                    color = ThemeHelper.getThemeColor(
-                        context,
-                        R.attr.colorOnSecondary,
-                    )
+                    color =  Color.parseColor(segmentColors[it.category])
                 },
             )
         }


Bnyro:
Could you please turn this into an option? I'd prefer using the same color (which fits the theme) for all of them personally.

app/src/main/java/com/github/libretube/ui/views/ColorPreference.kt
```diff
@@ -0,0 +1,102 @@
...
+            circleView.setBackgroundColor(
+                it
+            )
+        }
+    }
+
+    private fun showColorPickerDialog() {
+        val colorEditText = EditText(context)
+        val dialog = android.app.AlertDialog.Builder(context)
+            .setTitle("Enter Color Hex Value")
```
Bnyro, line 65: Please make that string translatable, (add it to strings.xml) and use `.setTitle(R.string.enter_color_hex)` or something like that`

app/src/main/java/com/github/libretube/ui/views/ColorPreference.kt
```diff
@@ -0,0 +1,102 @@
...
+                dialog.dismiss()
+            }
+            .create()
+
+        dialog.show()
+    }
+
+    private fun showInvalidColorMessage() {
+        val invalidColorMessage = "Invalid color value entered!"
+        Toast.makeText(context, invalidColorMessage, Toast.LENGTH_SHORT).show()
```
Bnyro, line 94: Please use `Toast.makeText(context, R.string.invalid_color_message, Toast.LENGTH_SHORT).show()` and that string to `strings.xml`

app/src/main/java/com/github/libretube/ui/views/ColorPreference.kt
```diff
@@ -0,0 +1,102 @@
...
+            )
+        }
+    }
+
+    private fun showColorPickerDialog() {
+        val colorEditText = EditText(context)
+        val dialog = android.app.AlertDialog.Builder(context)
+            .setTitle("Enter Color Hex Value")
+            .setView(colorEditText)
+            .setPositiveButton("OK") { _, _ ->
```
Bnyro, line 67: `.setPositiveButton(R.string.okay) { _, _ ->`

app/src/main/java/com/github/libretube/ui/views/ColorPreference.kt
```diff
@@ -0,0 +1,102 @@
...
+                        val color = Color.parseColor(hexValue)
+                        setColor(color)
+                    }
+                    catch (e: IllegalArgumentException){
+                        showInvalidColorMessage()
+                    }
+
+                }
+            }
+            .setNegativeButton("Cancel") { dialog, _ ->
```
Bnyro, line 84: `.setNegativeButton(R.string.cancel) { _, _ ->`

Bnyro (changes requested):

Bnyro:
Please remove the blank lines everywhere where reviewdog complains [here](https://github.com/libre-tube/LibreTube/pull/4055/files), it's not common to add two blank lines between functions or blank lines after a closed curly bracket.

app/src/main/java/com/github/libretube/ui/dialogs/ColorPickerDialog.kt
```diff
@@ -0,0 +1,169 @@
...
+        // Initialize UI elements
+        colorPreview = findViewById(R.id.colorPreview)
+        redSeekBar = findViewById(R.id.redSeekBar)
+        greenSeekBar = findViewById(R.id.greenSeekBar)
+        blueSeekBar = findViewById(R.id.blueSeekBar)
+        alphaSeekBar = findViewById(R.id.alphaSeekBar)
+        colorHexInput = findViewById(R.id.colorHexInput)
+        okay = findViewById(R.id.okay)
+        cancel = findViewById(R.id.cancel)
+
```
Bnyro, line 45: That's all obsolete when using view bindings, see above.

app/src/main/java/com/github/libretube/ui/dialogs/ColorPickerDialog.kt
```diff
@@ -0,0 +1,169 @@
...
+        super.onStart()
+        val dialogWidth = ViewGroup.LayoutParams.MATCH_PARENT
+        val dialogHeight = ViewGroup.LayoutParams.WRAP_CONTENT
+        window?.setLayout(dialogWidth, dialogHeight)
+    }
+
+    override fun onProgressChanged(seekBar: SeekBar?, progress: Int, fromUser: Boolean) {
+        // Update color preview when SeekBar progress changes
+        val color = getColor()
+        setColorPreview(color)
```
Bnyro, line 96:
```suggestion
        setColorPreview(getColor())
```

app/src/main/java/com/github/libretube/ui/dialogs/ColorPickerDialog.kt
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    override fun onStartTrackingTouch(seekBar: SeekBar?) {
+    }
```
Bnyro, line 100:
```suggestion
    override fun onStartTrackingTouch(seekBar: SeekBar?) = Unit
```

app/src/main/java/com/github/libretube/ui/dialogs/ColorPickerDialog.kt
```diff
@@ -0,0 +1,169 @@
...
+        val invalidColorMessage = R.string.invalid_color
+        Toast.makeText(context, invalidColorMessage, Toast.LENGTH_SHORT).show()
+    }
+
+
+    override fun onClick(v: View?) {
+        if (v?.id == R.id.okay) {
+            // Notify the selected color
+            val color = getColor()
+            onColorSelectedListener.onColorSelected(color)
```
Bnyro, line 121:
```suggestion
            onColorSelectedListener.onColorSelected(getColor())
```

app/src/main/java/com/github/libretube/ui/dialogs/ColorPickerDialog.kt
```diff
@@ -0,0 +1,169 @@
...
+    private fun setColor(color: Int, textUpdate: Boolean = false) {
+        // Set the SeekBar progress values based on the color
+        val red = Color.red(color)
+        val green = Color.green(color)
+        val blue = Color.blue(color)
+        val alpha = Color.alpha(color)
+        redSeekBar.progress = red
+        greenSeekBar.progress = green
+        blueSeekBar.progress = blue
+        alphaSeekBar.progress = alpha
```
Bnyro, line 146:
```suggestion
        redSeekBar.progress = Color.red(color)
        greenSeekBar.progress = Color.green(color)
        blueSeekBar.progress = Color.blue(color)
        alphaSeekBar.progress = Color.alpha(color)
```

app/src/main/java/com/github/libretube/ui/views/ColorPreference.kt
```diff
@@ -0,0 +1,80 @@
...
+    private fun showColorPickerDialog() {
+        (if (currentColor is Int) currentColor else Color.BLACK)?.let {
+            val dialog = ColorPickerDialog(
+                context,
+                it,
+                object : ColorPickerDialog.OnColorSelectedListener {
+                    override fun onColorSelected(color: Int) {
+                        setColor(color)
+                    }
+                })
```
Bnyro, line 68:
```suggestion
            val dialog = ColorPickerDialog(context, it) { color ->
                setColor(color)
            }
```

Bnyro (changes requested):

app/src/main/java/com/github/libretube/ui/dialogs/ColorPickerDialog.kt
```diff
@@ -0,0 +1,169 @@
...
+import android.widget.EditText
+import android.widget.SeekBar
+import android.widget.Toast
+import com.github.libretube.R
+
+class ColorPickerDialog(
+    context: Context,
+    private val initialColor: Int,
+    private val onColorSelectedListener: OnColorSelectedListener
+) : Dialog(context), SeekBar.OnSeekBarChangeListener, View.OnClickListener {
```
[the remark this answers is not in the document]
general-a: I can't seem to get this to work any suggestions? (I've tried a lot of SO solutions :/)

Screen Shot 2023-06-21 at 12 48 21 PM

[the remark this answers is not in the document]
Bnyro: And for the tag, you can use `this::class.java.name` instead of "acknowledgements"

app/src/main/java/com/github/libretube/ui/views/ColorPreference.kt
general-a: This seems to break the colorpickerdialog

app/src/main/java/com/github/libretube/ui/dialogs/ColorPickerDialog.kt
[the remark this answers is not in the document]
general-a: That did it! Why does that work and not something like dialog.childFragmentManager?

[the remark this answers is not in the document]
Bnyro: Preferences just don't have fragment managers and don't include an inbuilt function to get the activity. The activity is at the top of the view hierarchy (so the most important one), so it basically has the control to display fragments, dialogs, ...
Activity inherits from Context, and since we use a `SettingsActivity` we can just cast the type.

Bnyro:
Code looks good to me now (didn't yet test the changes though)

NeeRaj-2401:
>  (didn't yet test the changes though)

i did a clean install to this debug > toggled segment colors ON in settings > restarted the app > played [this video](https://piped.kavin.rocks/watch?v=doUB1hRzLVQ)

here is the results :(


but later i went through 
sponsor settings > clicked the green color (assigned for sponsor sections) > clicked 'OK' > started same video again

here is the results :)


as u can see the last section is still white, cause it wasn't touched by me in settings
if i would repeat the above step again with all the colors and categories then



hence, the problem is its not taking the default value.

general-a:
Odd, I'll do some debugging later to see what the issue is

general-a:
So it does seem that the preference help is not able to get the initial color. Do I need to tell it to get this value? If so where would I do that?

app/src/main/java/com/github/libretube/ui/dialogs/ColorPickerDialog.kt
```diff
@@ -0,0 +1,133 @@
...
+import com.github.libretube.databinding.DialogColorPickerBinding
+import com.google.android.material.dialog.MaterialAlertDialogBuilder
+
+class ColorPickerDialog(
+    private val context: Context,
+    private val initialColor: Int,
+    private val onColorSelectedListener: OnColorSelectedListener
+) : DialogFragment(), SeekBar.OnSeekBarChangeListener, View.OnClickListener {
+
+    private lateinit var binding: DialogColorPickerBinding
```
Isira-Seneviratne, line 23: The view binding has to be freed in the `onDestroyView` method: https://developer.android.com/topic/libraries/view-binding#fragments

```suggestion
    private var _binding: DialogColorPickerBinding? = null
    private val binding get() = _binding!!
```

app/src/main/java/com/github/libretube/ui/dialogs/ColorPickerDialog.kt
```diff
@@ -0,0 +1,133 @@
...
+
+    override fun onCreateDialog(savedInstanceState: Bundle?): Dialog {
+        binding = DialogColorPickerBinding.inflate(layoutInflater)
```
Isira-Seneviratne, line 26:
```suggestion
        _binding = DialogColorPickerBinding.inflate(layoutInflater)
```

app/src/main/java/com/github/libretube/ui/dialogs/ColorPickerDialog.kt
```diff
@@ -0,0 +1,133 @@
...
+                    }
+                }
+            }
+        })
+
+        return MaterialAlertDialogBuilder(requireContext())
+            .setView(binding.root)
+            .show()
+    }
+
```
Isira-Seneviratne, line 72:
```suggestion
    override fun onDestroyView() {
        super.onDestroyView()
        _binding = null
    }
```

app/src/main/java/com/github/libretube/ui/dialogs/ColorPickerDialog.kt
```diff
@@ -0,0 +1,133 @@
...
+    private fun setColorPreview(color: Int) {
+        // Set the color preview
+        binding.colorPreview.setBackgroundColor(color)
+    }
+
+    private fun colorToString(color: Int): String {
+        return String.format("#%08X", color)
+    }
+
+    interface OnColorSelectedListener {
```
Isira-Seneviratne, line 130:
```suggestion
    fun interface OnColorSelectedListener {
```

app/src/main/java/com/github/libretube/ui/views/ColorPreference.kt
Isira-Seneviratne: The interface needs to be marked as a functional interface for the suggestion.

app/src/main/java/com/github/libretube/ui/dialogs/ColorPickerDialog.kt
general-a: Why do we have two binding vars here(still new to kotlin)?

general-a:
> this video

Patch this bug

general-a:
> i did a clean install to this debug > toggled segment colors ON in settings > restarted the app > played [this video](https://piped.kavin.rocks/watch?v=doUB1hRzLVQ)

Patched this

PR #4097, opened 2023-06-25
Simplify toDownloadItems() method

Isira-Seneviratne:


app/src/main/java/com/github/libretube/api/obj/Streams.kt
```diff
@@ -35,40 +36,33 @@ data class Streams(
...
-        videoId: String,
-        fileName: String,
-        videoFormat: String?,
-        videoQuality: String?,
-        audioFormat: String?,
-        audioQuality: String?,
-        subtitleCode: String?
-    ): List<DownloadItem> {
+    fun toDownloadItems(downloadData: DownloadData): List<DownloadItem> {
+        val (id, name, videoFormat, videoQuality, audioFormat, audioQuality, subCode) = downloadData
```
Bnyro, line 40: Why are we calling the language of the subtitle `subCode` instead of `subtitleCode`? Wouldn't the latter be more self-explaining?

Isira-Seneviratne: That was to avoid lint issues due to the line length.

Bnyro (approved):

PR #4189, opened 2023-07-08
refactor: Use ViewModel extension in the main activity

Isira-Seneviratne:


Bnyro (approved):

Bnyro:
From now on, we'll append the category before the PR title, following the conventional commit types: https://github.com/commitizen/conventional-commit-types/blob/master/index.json.
This just makes it easier to generate the changelog in the end.
I'll add that to our README too later.

PR #4228, opened 2023-07-14
Replace MathUtils.clamp with Kotlin coerceIn

TacoTheDank:
Just so you can see it's the same:

MathUtils.clamp function:
```
public static int clamp(int value, int min, int max) {
    if (value < min) {
        return min;
    } else if (value > max) {
        return max;
    }
    return value;
}
```

Kotlin's coerceIn extension function:
```
public fun Int.coerceIn(minimumValue: Int, maximumValue: Int): Int {
    if (minimumValue > maximumValue) throw IllegalArgumentException("Cannot coerce value to an empty range: maximum $maximumValue is less than minimum $minimumValue.")
    if (this < minimumValue) return minimumValue
    if (this > maximumValue) return maximumValue
    return this
}
```

### DIFF
--- a/app/src/main/java/com/github/libretube/helpers/AudioHelper.kt
+++ b/app/src/main/java/com/github/libretube/helpers/AudioHelper.kt
@@ -3,7 +3,6 @@ package com.github.libretube.helpers
 import android.content.Context
 import android.media.AudioManager
 import androidx.core.content.getSystemService
-import androidx.core.math.MathUtils
 import androidx.media.AudioManagerCompat
 import com.github.libretube.extensions.normalize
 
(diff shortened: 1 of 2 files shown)


TacoTheDank:
@Isira-Seneviratne You might be interested in this, seeing as this changes a line that git blames you as the last person to touch it. I think you'll like this change :)

Isira-Seneviratne:
> @Isira-Seneviratne You might be interested in this, seeing as this changes a line that git blames you as the last person to touch it. I think you'll like this change :)

The Kotlin function throws an exception of the minimum is greater than the maximum.

Bnyro:
At least its syntax looks way cleaner to Mr than using `MathUtils`. I'm not sure if we have any edge cases that could cause the `IllegalArgumentException` to be thrown, so I'd like to test this a bit before merging however.
Thanks for your PR!

Bnyro (approved):

PR #4276, opened 2023-07-20
feat: support for DeArrow

Bnyro:


app/src/main/res/values/strings.xml
```diff
@@ -433,6 +433,9 @@
     <string name="visible">Show in seek bar</string>
     <string name="fallback_piped_proxy">Fallback to Piped proxy</string>
     <string name="fallback_piped_proxy_desc">Load videos via the proxy if connecting to YouTube directly doesn\'t work for the current video (increases the initial loading times). If disabled, YouTube music content likely won\'t play due to YT restrictions.</string>
+    <string name="dearrow">Enable DeArrow</string>
+    <string name="dearrow_summary">Show more accurate and less sensationalism titles and thumbnails. Increases loading times.</string>
```
Isira-Seneviratne, line 437:
```suggestion
    <string name="dearrow_summary">Show more accurate and less sensationalist titles and thumbnails. Increases loading times.</string>
```

app/src/main/java/com/github/libretube/util/DeArrowUtil.kt
```diff
@@ -0,0 +1,89 @@
...
+import com.github.libretube.helpers.PreferenceHelper
+import kotlinx.serialization.json.JsonElement
+import kotlinx.serialization.json.decodeFromJsonElement
+
+object DeArrowUtil {
+    private fun extractTitleAndThumbnail(data: JsonElement): Pair<String?, String?> {
+        val content = try {
+            JsonHelper.json.decodeFromJsonElement<DeArrowContent>(data)
+        } catch (e: Exception) {
+            return Pair(null, null)
```
Isira-Seneviratne, line 20:
```suggestion
            return null to null
```

app/src/main/java/com/github/libretube/util/DeArrowUtil.kt
```diff
@@ -0,0 +1,89 @@
...
+        }
+        val newTitle = content.titles.maxByOrNull { it.votes }?.title
+        val newThumbnail =
+            content.thumbnails.filter { it.thumbnail != null }.maxByOrNull { it.votes }
+                ?.takeIf { !it.original }?.thumbnail
+        return Pair(newTitle, newThumbnail)
```
Isira-Seneviratne, line 26:
```suggestion
        return newTitle to newThumbnail
```

app/src/main/java/com/github/libretube/util/DeArrowUtil.kt
```diff
@@ -0,0 +1,89 @@
...
+            JsonHelper.json.decodeFromJsonElement<DeArrowContent>(data)
+        } catch (e: Exception) {
+            return null to null
+        }
+        val newTitle = content.titles.maxByOrNull { it.votes }?.title
+        val newThumbnail =
+            content.thumbnails.filter { it.thumbnail != null }.maxByOrNull { it.votes }
+                ?.takeIf { !it.original }?.thumbnail
+        return newTitle to newThumbnail
+    }
```
ajayyy, line 27: Array is presorted optimally (manually approved submissions above others even if less votes). Just take the first element.

However, you still need to make sure the top value is valid, as it will return submissions at -1 votes.

```kt
        val newTitle = content.titles.firstOrNull?.takeIf { it.votes >= 0 || it.locked }.title
        val newThumbnail =
            content.thumbnails.filter { it.thumbnail != null }.firstOrNull?.takeIf { (it.votes >= 0 || it.locked) && !it.original }?.thumbnail
```

Bnyro: I see, thanks!

PR #4290, opened 2023-07-22
fix: Sort videoIds before performing DeArrow fetch

FireMasterK:
This should both improve privacy slightly and improve cache hits on the server.


app/src/main/java/com/github/libretube/util/DeArrowUtil.kt
```diff
@@ -55,7 +59,11 @@ object DeArrowUtil {
         if (!PreferenceHelper.getBoolean(PreferenceKeys.DEARROW, false)) return contentItems
 
         val videoIds = contentItems.filter { it.type == "stream" }
-            .joinToString(",") { it.url.toID() }
+            .mapNotNull { it.url.toID() }
```
Bnyro, line 62: We should use `map` here, since url is non null

Bnyro (approved):
Lgtm

Issue #4325, opened 2023-07-29
Crash when opening a video after tags alpha update!


IndusAryan:
### Steps to reproduce

play video which dont have these tags,licenses etc on latest alpha

### Expected behavior

video should play

### Actual behavior

C.R.A.S.H

### LibreTube version

Latest Alpha

### Android version

A13

### Other details

kotlinx.serialization.MissingFieldException: Fields [license, visibility, tags] are required for type with serial name 'com.github.libretube.api.obj.Streams', but they were missing at path: $
 at kotlinx.serialization.json.internal.StreamingJsonDecoder.decodeSerializableValue(StreamingJsonDecoder.kt:93)
 at kotlinx.serialization.json.Json.decodeFromString(Json.kt:107)
 at com.jakewharton.retrofit2.converter.kotlinx.serialization.Serializer$FromString.fromResponseBody(Serializer.kt:26)
 at com.jakewharton.retrofit2.converter.kotlinx.serialization.DeserializationStrategyConverter.convert(DeserializationStrategyConverter.kt:11)
 at com.jakewharton.retrofit2.converter.kotlinx.serialization.DeserializationStrategyConverter.convert(DeserializationStrategyConverter.kt:7)
 at retrofit2.OkHttpCall.parseResponse(OkHttpCall.java:243)
 at retrofit2.OkHttpCall$1.onResponse(OkHttpCall.java:153)
 at com.google.net.cronet.okhttptransport.CronetCallFactory$CronetCall$2.onSuccess(CronetCallFactory.java:152)
 at com.google.net.cronet.okhttptransport.CronetCallFactory$CronetCall$2.onSuccess(CronetCallFactory.java:148)
 at com.google.common.util.concurrent.Futures$CallbackListener.run(Futures.java:1105)
 at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1137)
 at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:637)
 at java.lang.Thread.run(Thread.java:1012)
 Suppressed: kotlinx.coroutines.internal.DiagnosticCoroutineContextException: [StandaloneCoroutine{Cancelling}@ddd71e7, Dispatchers.IO]
Caused by: kotlinx.serialization.MissingFieldException: Fields [license, visibility, tags] are required for type with serial name 'com.github.libretube.api.obj.Streams', but they were missing
[...]

PR #4326, opened 2023-07-29
NullSafe Tags,licenses and Visibility.(Fix to #4325)

IndusAryan:
Screenshot_20230729_213646
Screenshot_20230729_213615


Bnyro:
Thank you, however this is not necessary since all instances should be auto-updated within a day, and all the params are non-null at server side.

PR #4328, opened 2023-07-29
fix: Sort local playlists based on selected preference

0x24D:
Since #3774 was merged, when adding a new video to a local playlist all playlists are sorted in newest to oldest order, even when the "Playlists order" preference is set otherwise. This PR makes the playlist sorting respect the preference.


Bnyro:
Like that, playlists used with an account wouldn't be sorted anymore?
I don't really understand how that change should fix the issue.

Bnyro:
I think it makes sense to move the sorting to the `PlaylistsHelper` so that they're sorted the same way in the `AddToPlaylistDialog`, but both, online and local playlists should be sorted the same way.

0x24D:
I didn't think about online playlists as I didn't have an account. I have just created a test account on an instance and this seems to work for both.

Bnyro:
Line 131 in `LibraryFragment.kt` can now be `val` instead of `var`, otherwise lgtm.

Bnyro (approved):
Thank you!

PR #4422, opened 2023-08-11
fix: duplicate comments

### LINKED ISSUE #4291: Comment section sometimes shows the same comments from previous loads
### Steps to reproduce

1. After opening the app, load a page of any video that has a lot of comments
2. Open the comment section
3. Scroll the comment section a few times causing comments to load
4. Issue: comment section sometimes shows the same comments from previous loads

### Expected behavior

Comments that were already shown should be excluded from the next loads of comments

### Actual behavior

Comment section sometimes shows the same comments from previous loads

### LibreTube version

0.16.1

### Android version

13

### Other details

_No response_

### Acknowledgements

salahmak:
Fix #4291

Sometimes, the Piped API's response contains comments that were already given in previous pages, so I have added a filter to remove them.

This might not be the best way to do it, but this is my first time working with Kotlin/Android because I found this bug annoying and wanted to do whatever it takes to fix it, so if there is a better way, please do it.

:)

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/models/CommentsViewModel.kt
+++ b/app/src/main/java/com/github/libretube/ui/models/CommentsViewModel.kt
@@ -56,9 +56,16 @@ class CommentsViewModel : ViewModel() {
                 Log.e(TAG(), e.toString())
                 return@launch
             }
+
             val updatedPage = commentsPage.value?.apply {
-                comments += response.comments.filterNonEmptyComments()
+                val combinedComments = this.comments + response.comments.filterNonEmptyComments()
+
+                // filtering out the comments that have the same id (duplicated)
+                this.comments = combinedComments.groupBy { it.commentId }
+                    .filter { it.value.size == 1 }
+                    .flatMap { it.value }
             }
+
             nextPage = response.nextpage
             commentsPage.postValue(updatedPage)
             isLoading = false
--- a/app/src/main/java/com/github/libretube/ui/models/CommentsViewModel.kt
+++ b/app/src/main/java/com/github/libretube/ui/models/CommentsViewModel.kt
@@ -58,12 +58,9 @@ class CommentsViewModel : ViewModel() {
             }
 
             val updatedPage = commentsPage.value?.apply {
-                val combinedComments = this.comments + response.comments.filterNonEmptyComments()
-
-                // filtering out the comments that have the same id (duplicated)
-                this.comments = combinedComments.groupBy { it.commentId }
-                    .filter { it.value.size == 1 }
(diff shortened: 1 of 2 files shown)


FireMasterK:
Can you give an example video where this happens?

salahmak:
@FireMasterK I don't currently have a video of it happening, but it usually happens when doing these steps:

1. Open comments section
2. Scroll down (load 1 or 2 more pages)
3. Open a comment's replies
4. Close the replies
5. Scroll down more comments (load more comments pages)
6. Now you should start noticing it happen

I have confirmed this by checking for duplicates in the `commentPage.comments` List (I didn't include the code in the commit as it was for my debugging) and it confirmed it for me after doing the above steps

salahmak:
What I think is causing the problem is the process of loading the replies of a comment and then going back, perhaps the wrong parameters are being set or overridden. I couldn't get my head around how that transition is happening (as I'm a beginner in Kotlin)

app/src/main/java/com/github/libretube/ui/models/CommentsViewModel.kt
```diff
@@ -56,9 +56,16 @@ class CommentsViewModel : ViewModel() {
                 Log.e(TAG(), e.toString())
                 return@launch
             }
+
             val updatedPage = commentsPage.value?.apply {
-                comments += response.comments.filterNonEmptyComments()
+                val combinedComments = this.comments + response.comments.filterNonEmptyComments()
```
Bnyro, line 61:
```kotlin
comments += response.comments.filterNonEmptyComments().filter { comment -> comments.none { it.commentId == comment.commented }
```

Bnyro: That's more performant and doesn't impac or modify the already visible comments.

salahmak: I'll try it and see

salahmak: It works as intended, and I have pushed the commit that uses this. Thanks

Bnyro:
Thank you!

salahmak:
You're welcome @Bnyro, although I still think that this is just a workaround to make it not happen, I'm kinda sure that the cause of the bug is something else related to moving from the comments view to a replies view and then going back.

Opening replies is buggy as well, as sometimes it opens the wrong comment (I'll open a different issue for this).

Bnyro:
Yes, I agree, but it's better to have a temporary workaround/fix instead of keeping the issue present :)

PR #4423, opened 2023-08-11
feat: Show channel icons for stream notifications

Isira-Seneviratne:
* Download and display channel icons for the stream summary notifications. This only occurs if thumbnails are enabled.
* Use the [social category](https://developer.android.com/reference/android/app/Notification#CATEGORY_SOCIAL) for stream notifications.

**Screenshots**
Screenshot_20230811_065654


Bnyro (approved):

Bnyro:
I think it's better if only the notification group/summary shows the channel avatar and not each individual stream notification, because all individual streams grouped together have the same uploaded anyways.

PR #4430, opened 2023-08-12
refactor: Use LocalDate.ofInstant()

Isira-Seneviratne:
* Bump compile SDK to 34 (sources are now available for download).
* Use [`LocalDate.ofInstant()`](https://developer.android.com/reference/java/time/LocalDate#ofInstant(java.time.Instant,%20java.time.ZoneId)).

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -9,7 +9,7 @@ plugins {
 }
 
 android {
-    compileSdk 33
+    compileSdk 34
 
     defaultConfig {
         applicationId 'com.github.libretube'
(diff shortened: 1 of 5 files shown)


Bnyro (approved):

Bnyro:
Haven't found any breaking changes coming with SDK 34 that are relevant for LibreTube, so we should be fine.

Isira-Seneviratne:
> Haven't found any breaking changes coming with SDK 34 that are relevant for LibreTube, so we should be fine.

I only updated the compile SDK, not the target SDK.

Bnyro:
Oh, I see.

PR #4472, opened 2023-08-14
feat: add tooltips

FineFindus:
Adds [tooltips](https://developer.android.com/develop/ui/views/components/tooltips) to various icon buttons. This not only helps with accessibility, but is also useful for new users, who may not immediately associate an icon with a certain action.
I most likely missed some, in that case, I will try to do to a follow-up pr once I notice them.


Bnyro:
Looks like a great enhancement! Please run `ktlint --android -F` to fix the linter.

FineFindus:
Thanks! I was about to investigate the CI failure, when Android Studio decided to crash and take my Wayland session with it. Guess it is enough android dev for today (once the linter is finally happy :) )

Bnyro:
I'm planning to do a bug fix release soon, so I want to merge new features after that release has published. Hope you're fine with waiting a bit until we're merging this :)

Bnyro:
Since it's not supported for some Android versions, I would prefer an extension function `ImageView#trySetTooltip` which does the version check any set the tooltip if supported by the Android version.

Bnyro:
Thank you!

PR #4481, opened 2023-08-15
fix(import): correctly import playlist csv

FineFindus:
Same issue as in https://github.com/libre-tube/LibreTube/pull/4466.


Bnyro (approved):
Thanks!

PR #4482, opened 2023-08-15
feat(stats): use bottom sheet

FineFindus:
Updated the Stats for Nerds dialog to use a bottom sheet instead of a dialog. This is more consistent with the rest of the application design, which mostly uses bottom sheets instead of dialogs.
Using a bottom sheet also improves the display of the stats in landscape mode.

| Before                                                                                                                          	| After                                                                                                                                    	|
|---------------------------------------------------------------------------------------------------------------------------------	|------------------------------------------------------------------------------------------------------------------------------------------	|
| Dialog showing stats for nerds 	| Stats for nerds shown in a bottom sheet 	|


Bnyro:
I think it's fine to move it to a bottom sheet, but
1. We should keep the title present in the bottom sheet
2. I'd prefer keeping the text fields as before

FineFindus:
Stats for Nerds sheet with title and text fields

> I'd prefer keeping the text fields as before

I'm not sure about this one. Having disabled text fields without any context to the user can be pretty disorienting. The [M3 Design Guidelines](https://m3.material.io/components/text-fields/guidelines#f572b53d-1a15-4ede-881a-80adbbdc24a5) suggest marking them as read-only, which would inform the user that the text fields cannot be edited.

Bnyro:
Imo, the title should not be centered.

The text fields make the dialog feel much better and less empty, I still think that it's the best choice.

FineFindus:
Start-aligned title

I agree that the dialog looks a bit empty otherwise, maybe a larger font/adjusted layout would solve that? I'm not a designer, so I'm not really the right person to solve it, I was just super confused, why the stats were text fields, as they cannot be changed by the user.

Bnyro:
I think it's just fine like as it is now, I don't think users would be confused by this in general. Other popular Android music players, for example Auxio, do something similar for displaying stats.

Bnyro (approved):

PR #4498, opened 2023-08-15
feat(player): allow pausing when loading

FineFindus:
Allows to pause the player, while the video is buffering. This can be helpful if the video is constantly buffering, to load a longer timeframe. It is also helpful when simply wanting to pause the video while it is buffering.

To test:
1. Start watching a video
2. Turn off the network
3. Wait until the video starts to buffer
4. Pause the video and turn the network back on
5. Notice how the video does not start playing again once it finished loading


FineFindus:
Maybe? I thought of this as adding new behavior to the player rather than fixing a bug, not allowing pausing when buffering is acceptable behavior and not really a bug.

Bnyro:
In my opinion, it should work like that by default, so it's rather a fix I think.

Bnyro:
Thanks though!

PR #4538, opened 2023-08-18
chore: update dutch translations (keyword Weblate)

BabyBenefactor:
Translated the entire strings.xml to Dutch.

Weblate was not being used as discussed with Bnyro.


Bnyro:
Unfortunately, builds fail because something is not formatted properly, seems like a `</string>` misses somewhere. From my quick look I couldn't figure out where the issue is exactly.

Issue #3558, opened 2023-04-15
Improve channel group UI


SameenAhnaf:
### Describe your suggested feature

- [x] 1. Place ✏️ icon (group manager menu accessible at the very beginning, not channel addition menu) on the left of all channel groups, not right.
- [x] 2. Within group manager menu, allow reordering groups similar to navigation tabs by replacing the pencil column with sliders' column within green box. See the screenshot for reference.
- [x] 3. Edit a channel group. Making changes to the name field and tapping "Cancel" button will not save the changes. But if you add or remove channels from the group, the changes are saved anyway after cancelling.
- ~[ rejected ] 4. For group manager menu,  there's no need for an entire column of pencil icons. Instead, long press on any group to open channel addition menu.~

- [x] 5. Place switcher between channel list and feed under groups, not above. Browse both channels and feed within any specific group. Only show unselected channels in channel addition menu. While showing channel list from any specific group, long press on the channel to show `Remove from group` option.
- ~[ rejected ] 6. Show `Feed` and `Channels` instead of kinda ambiguous *Subscriptions* headline. Show number of channels as per selected group.~
- [x] 7. Show a warning if no group name has been set for or no channel is added to any group while creation.
- ~[ rejected ] 8. `Play latest videos` should only load new videos found after each reload, not the entire list of videos.~

- ~[ rejected ] 9. For `All` channel group, filter channels by number of added groups. The options should be available in radio buttons for `0`, `1`, `2` and `More than 2` number of groups.~
- ~[ rejected ] 10. Instead of `Unsubscribe` button after subscribing channels, show `SUBSCRIBED` button to show an all-in-one menu to unsubscribe, add to groups and enable notifications for the channel. This new menu should be available for legacy subscriptions layout by long press on the channel.~
- [x] 11. [BUG] Although notification for new streams is disabled, notification bell is shown after unsubscribing from subscriptions' list and subscribing again.
- [x] 12. Allow setting a default group that shows initially when opening the subscriptions tab or remember the last used channel group and restore it when opening the app the next time.

### Other details

![Screenshot_20230416-004456](https://user-images.githubusercontent.com/79245641/232247891-6d8038de-5db0-4a29-aebe-a3bf590f77d7.jpg)

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

[...]

PR #4550, opened 2023-08-19
feat(channelGroups): disable confirm button if creation is not possible

FineFindus:
Disables the confirm button when creating or modifying a channel group if the group cannot be created. This can either be the case if the group name is empty, or (newly added) no channels are selected.

Disabled confirm button with error text
Enabled confirm button

addresses #3558

### DIFF
--- a/app/src/main/java/com/github/libretube/db/dao/SubscriptionGroupsDao.kt
+++ b/app/src/main/java/com/github/libretube/db/dao/SubscriptionGroupsDao.kt
@@ -11,6 +11,9 @@ interface SubscriptionGroupsDao {
     @Query("SELECT * FROM subscriptionGroups")
     suspend fun getAll(): List<SubscriptionGroup>
 
+    @Query("SELECT EXISTS(SELECT * FROM subscriptionGroups WHERE name = :name)")
+    suspend fun exists(name: String): Boolean
+
     @Insert(onConflict = OnConflictStrategy.REPLACE)
     suspend fun createGroup(subscriptionGroup: SubscriptionGroup)
 
--- a/app/src/main/java/com/github/libretube/ui/sheets/EditChannelGroupSheet.kt
+++ b/app/src/main/java/com/github/libretube/ui/sheets/EditChannelGroupSheet.kt
@@ -9,6 +9,7 @@ import androidx.core.widget.addTextChangedListener
 import androidx.fragment.app.activityViewModels
 import androidx.lifecycle.lifecycleScope
 import androidx.recyclerview.widget.LinearLayoutManager
+import com.github.libretube.R
 import com.github.libretube.api.SubscriptionHelper
 import com.github.libretube.api.obj.Subscription
 import com.github.libretube.databinding.DialogEditChannelGroupBinding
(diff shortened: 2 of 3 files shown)


app/src/main/java/com/github/libretube/ui/sheets/EditChannelGroupSheet.kt
```diff
@@ -78,8 +84,23 @@ class EditChannelGroupSheet(
...
+        with(binding) {
+            val isGroupNameBlank = groupName.text?.isBlank() == true
+
+            groupName.error = if (isGroupNameBlank) {
+                requireContext().getString(R.string.group_name_error)
+            } else {
+                null
+            }
+
+            confirm.isEnabled = !isGroupNameBlank && group.channels.isNotEmpty()
```
Bnyro, line 103: The confirm button is activates when the group name is not empty - whereas here, we additionally require that there's at least one channel selected. Both should behave the same way. I think we should also check if no existing group has the same name - if it's a new group - so that users don't accidentally override the old group.

FineFindus: > whereas here, we additionally require that there's at least one channel selected

Yes, I added that requirement, because I don't think it should be possible to create empty groups, as they don't serve any purpose. I can revert the change if you want.

>  I think we should also check if no existing group has the same name - if it's a new group - so that users don't accidentally override the old group.

Done in 86854813e20ed71f9f52acfec7daa79fcc79aadd

app/src/main/java/com/github/libretube/db/dao/SubscriptionGroupsDao.kt
```diff
@@ -11,6 +11,9 @@ interface SubscriptionGroupsDao {
     @Query("SELECT * FROM subscriptionGroups")
     suspend fun getAll(): List<SubscriptionGroup>
 
+    @Query("SELECT * FROM subscriptionGroups WHERE name = :name")
+    suspend fun get(name: String): SubscriptionGroup?
```
Bnyro, line 15:
```suggestion
    @Query("SELECT EXISTS(SELECT * FROM subscriptionGroups WHERE name = :name)")
    suspend fun exists(name: String): Boolean
```

app/src/main/java/com/github/libretube/ui/sheets/EditChannelGroupSheet.kt
```diff
@@ -78,8 +86,34 @@ class EditChannelGroupSheet(
...
+        if (groupNameExists(name)) {
+            return requireContext().getString(R.string.group_name_error_exists)
+        }
+
+        return null
+    }
+
+    private fun groupNameExists(name: String): Boolean = runBlocking(Dispatchers.IO) {
+        DatabaseHolder.Database.subscriptionGroupsDao().get(name) != null
+    }
```
Bnyro, line 118: This should not be a separate function - it's only one line and only used once.

app/src/main/java/com/github/libretube/db/dao/SubscriptionGroupsDao.kt
FineFindus: Interesting, didn't know this existed (pun intended), thanks!

Bnyro (approved):
Thank you!

PR #4551, opened 2023-08-19
feat(channelGroups): allow re-ordering of groups

FineFindus:
Addresses https://github.com/libre-tube/LibreTube/issues/3558 by allowing channel groups to be rearranged using DnD.

https://github.com/libre-tube/LibreTube/assets/63370021/18d7fbe8-08ff-4feb-a13b-6f18ffeebf1a

This will currently update all groups in the database, this may cause problems with larger sets of groups, in which case it may be worth checking which groups need updating.
Due to the other issues listed in https://github.com/libre-tube/LibreTube/issues/3558 with the Ui, I did not add drag handles.

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/fragments/SubscriptionsFragment.kt
+++ b/app/src/main/java/com/github/libretube/ui/fragments/SubscriptionsFragment.kt
@@ -187,6 +187,7 @@ class SubscriptionsFragment : Fragment() {
         binding.channelGroups.removeAllViews()
 
         binding.channelGroups.addView(binding.chipAll)
+        channelGroups = channelGroups.sortedBy { subscriptionGroup -> subscriptionGroup.index }
         channelGroups.forEach { group ->
             val chip = layoutInflater.inflate(R.layout.filter_chip, null) as Chip
             chip.apply {
--- a/app/src/main/java/com/github/libretube/ui/fragments/SubscriptionsFragment.kt
+++ b/app/src/main/java/com/github/libretube/ui/fragments/SubscriptionsFragment.kt
@@ -187,7 +187,7 @@ class SubscriptionsFragment : Fragment() {
         binding.channelGroups.removeAllViews()
 
         binding.channelGroups.addView(binding.chipAll)
-        channelGroups = channelGroups.sortedBy { subscriptionGroup -> subscriptionGroup.index }
+        channelGroups = channelGroups.sortedBy { it.index }
         channelGroups.forEach { group ->
             val chip = layoutInflater.inflate(R.layout.filter_chip, null) as Chip
             chip.apply {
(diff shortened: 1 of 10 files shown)


Bnyro:
It's not yet self-explaining to users that it's possible to reorder them - I think we should add an indicator for that like it's present in the queue bottom sheet or in the navigation bar order options dialog.

FineFindus:
Channel group bottom sheet with drag handle
I initially didn't add due to the already listed UI issues in https://github.com/libre-tube/LibreTube/issues/3558, as I didn't want to add another icon to the already crowded list of icons.

Bnyro:
> Channel group bottom sheet with drag handle I initially didn't add due to the already listed UI issues in #3558, as I didn't want to add another icon to the already crowded list of icons.
>
I get that, but there's no other way to tell users that they can re-order these.

The drag handle should be at the left though, so that it can be clearly distinguished from the edit and delete icon.

app/src/main/java/com/github/libretube/ui/sheets/ChannelGroupsSheet.kt
```diff
@@ -46,6 +49,35 @@ class ChannelGroupsSheet(
...
+                viewHolder: RecyclerView.ViewHolder,
+                target: RecyclerView.ViewHolder
+            ): Boolean {
+                val from = viewHolder.absoluteAdapterPosition
+                val to = target.absoluteAdapterPosition
+
+                groups.move(from, to)
+                adapter.notifyItemMoved(from, to)
+
+                groups.mapIndexed { index, subscriptionGroup -> subscriptionGroup.index = index }
```
Isira-Seneviratne, line 67:
```suggestion
                groups.forEachIndexed { index, subscriptionGroup -> subscriptionGroup.index = index }
```

app/src/main/java/com/github/libretube/ui/fragments/SubscriptionsFragment.kt
```diff
@@ -187,6 +187,7 @@ class SubscriptionsFragment : Fragment() {
         binding.channelGroups.removeAllViews()
 
         binding.channelGroups.addView(binding.chipAll)
+        channelGroups = channelGroups.sortedBy { subscriptionGroup -> subscriptionGroup.index }
```
Isira-Seneviratne, line 190:
```suggestion
        channelGroups = channelGroups.sortedBy { it.index }
```

Isira-Seneviratne (approved):
Looks good to me overall, added a couple of small suggestions.

app/src/main/java/com/github/libretube/ui/sheets/ChannelGroupsSheet.kt
FineFindus: I'm not familiar with Kotlin, what's the difference between the two? The documentation wasn't much help.

Isira-Seneviratne: `mapIndexed` creates a new list with the content specified in the lambda, while `forEachIndexed` just iterates over the items in the list.

Bnyro (approved):

PR #4555, opened 2023-08-20
chore: Migrate to KSP

Isira-Seneviratne:
It is recommended to switch from kapt to KSP: https://developer.android.com/build/migrate-to-ksp


FireMasterK (approved):
Looks fine according to the documentation!

Bnyro (approved):

Bnyro:
`up to 2 times faster ` according to the docs 😋

PR #4559, opened 2023-08-20
feat(settings): move proxy settings to instance page

FineFindus:
Moves the option to disable the piped proxy to the instance settings page. The proxy preferences align better with the instance settings, given they refer more to video loading than to audio and video configurations. A newcomer might equate proxy settings to "Avoid using a piped instance."

Instance settings with proxy settings


Bnyro (approved):
Thanks!

PR #4584, opened 2023-08-22
feat: add sponsorblock option to skip only once

FineFindus:
Adds a new SponsorBlock skip option, which only skips the segment when watching the first time. This allows the user to go back and rewatch a segment if they wish to do so. 
When rewatching an already skipped segment, the manual skip option is shown instead.

https://github.com/libre-tube/LibreTube/assets/63370021/d80e73e2-9096-4dae-89d1-3d3d7c32d4b9

### DIFF
--- a/app/src/main/java/com/github/libretube/api/obj/Segment.kt
+++ b/app/src/main/java/com/github/libretube/api/obj/Segment.kt
@@ -12,7 +12,8 @@ data class Segment(
     val segment: List<Double> = listOf(),
     val userID: String? = null,
     val videoDuration: Double? = null,
-    val votes: Int? = null
+    val votes: Int? = null,
+    var skipped: Boolean? = false
 ) {
     val segmentStartAndEnd = segment[0] to segment[1]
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/github/libretube/api/obj/Segment.kt
```diff
@@ -12,7 +12,8 @@ data class Segment(
     val segment: List<Double> = listOf(),
     val userID: String? = null,
     val videoDuration: Double? = null,
-    val votes: Int? = null
+    val votes: Int? = null,
+    var skipped: Boolean? = false
```
Bnyro, line 16: Why's `skipped` nullable here? Shouldn't it always either be true or false?

FineFindus: I originally set it as nullable, so it could be created from the API, but forgot to remove it later on

Bnyro (approved):
Thanks!

PR #4590, opened 2023-08-22
feat: support for submitting SponsorBlock segments

### LINKED ISSUE #132: Support for submitting sponsor segments
**Describe the feature you'd like to see added**
I would like there to be a way to submit sponsorblock segments through the app, so we can contribute to the segment database
**Additional context**
#18 is obviously a prerequisite to this. I also think this would make more sense to be inside this app rather than through the piped web client, as you would have to either trust the instance hoster to not do anything malicious with your private key or snatch it from localstorage. If kavin wants to have submitting sponsor segments directly into piped though, we can move this issue over there instead.

Bnyro:
closes #132


Bnyro:
@ajayyy If you have a second, it would be great if you could look at he changes and tell me if anything should be done different generally, since I'm not really familiar with the SponsorBlock API.

ajayyy:
Is there any way to move +- 100 ms like in the vanced submission menu?

Roki100:
nice, big fan

app/src/main/java/com/github/libretube/helpers/PreferenceHelper.kt
```diff
@@ -143,6 +144,16 @@ object PreferenceHelper {
         ).apply()
     }
 
+    fun getSponsorBlockUUID(): String {
+        var uuid = getString(PreferenceKeys.SPONSORBLOCK_UUID, "")
+        if (uuid.isEmpty()) {
+            // generate a new uuid
+            uuid = UUID.randomUUID().toString()
```
ajayyy, line 151: Can you change the generation to use something like this https://stackoverflow.com/a/20536819

Bnyro: Why is it called UUID on the SponsorBlock api docs if it's not a UUID then? For reference, https://en.m.wikipedia.org/wiki/Universally_unique_identifier.

ajayyy: It used to be a UUID but it turns out that they are not guarenteed to not be predictable, which means they are not good for authentication

FireMasterK: > It used to be a UUID but it turns out that they are not guarenteed to not be predictable, which means they are not good for authentication. Will update docs

Actually, this is false. Most UUID implementation (including Java's) are generated with cryptographically secure random number generator.

mini-bomba: UUIDs can hold at most 128 bits, but sha256 is used to derive the public userID, which can hold 256 bits. So if UUIDs were used for private userIDs, you'd get much less combinations for the private part than the public part, making it easier to guess the private part.
Also in most of the places in the API docs hosted on the wiki, the term "userID" is used instead of submitter UUID. (it probably should be called "private userID" to clearly separate it from the derived public userID)

app/src/main/java/com/github/libretube/ui/dialogs/SubmitSegmentDialog.kt
```diff
@@ -0,0 +1,86 @@
...
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
+
+class SubmitSegmentDialog(
+    private val videoId: String,
+    private val currentPosition: Long,
+    private val duration: Long?
+) : DialogFragment() {
+    private lateinit var binding: DialogSubmitSegmentBinding
```
Isira-Seneviratne, line 28:
```suggestion
    private var _binding: DialogSubmitSegmentBinding? = null
    private val binding get() = _binding!!
```

app/src/main/java/com/github/libretube/ui/dialogs/SubmitSegmentDialog.kt
```diff
@@ -0,0 +1,86 @@
...
+
+                    lifecycleScope.launch {
+                        submitSegment()
+                        dismiss()
+                    }
+                }
+            }
+    }
+
+    private suspend fun submitSegment() {
```
Isira-Seneviratne, line 60:
```suggestion
    override fun onDestroyView() {
        super.onDestroyView()
        _binding = null
    }

    private suspend fun submitSegment() {
```

app/src/main/java/com/github/libretube/ui/dialogs/SubmitSegmentDialog.kt
```diff
@@ -0,0 +1,86 @@
...
+            .apply {
+                getButton(DialogInterface.BUTTON_POSITIVE).setOnClickListener {
+                    requireDialog().hide()
+
+                    lifecycleScope.launch {
+                        submitSegment()
+                        dismiss()
+                    }
+                }
+            }
```
Isira-Seneviratne, line 57:
```suggestion
            .setPositiveButton(R.string.okay) { _, _ ->
                requireDialog().hide()

                lifecycleScope.launch {
                    submitSegment()
                    dismiss()
                }
            }
            .setNegativeButton(R.string.cancel, null)
            .show()
```

Isira-Seneviratne (changes requested):
Aside from the requested changes, this looks good overall.

app/src/main/java/com/github/libretube/ui/dialogs/SubmitSegmentDialog.kt
```diff
@@ -0,0 +1,86 @@
...
+
+class SubmitSegmentDialog(
+    private val videoId: String,
+    private val currentPosition: Long,
+    private val duration: Long?
+) : DialogFragment() {
+    private lateinit var binding: DialogSubmitSegmentBinding
+
+    override fun onCreateDialog(savedInstanceState: Bundle?): Dialog {
+        binding = DialogSubmitSegmentBinding.inflate(layoutInflater)
```
Isira-Seneviratne, line 31:
```suggestion
        _binding = DialogSubmitSegmentBinding.inflate(layoutInflater)
```

app/src/main/java/com/github/libretube/ui/dialogs/SubmitSegmentDialog.kt
Bnyro: This doesn't work, because we get a `CoroutineCancelationException` then, as the life cycle ends automatically after a short time when clicking the dialog button. We already do the same work around as here at other places, e.g. in https://github.com/libre-tube/LibreTube/blob/b1cccdd935c3de3ad3136b9e30b809f82b22363f/app/src/main/java/com/github/libretube/ui/dialogs/RenamePlaylistDialog.kt#L40.

app/src/main/java/com/github/libretube/helpers/PreferenceHelper.kt
Bnyro: Which length should the userID be of then?

rozari0:
LT is crashing when I click that new sponsor block icon in player.

PR #4616, opened 2023-08-25
fix: Free view bindings in dialogs

Isira-Seneviratne:
Allow dialog view bindings to be garbage collected when the dialog is destroyed.

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/sheets/IntentChooserSheet.kt
+++ b/app/src/main/java/com/github/libretube/ui/sheets/IntentChooserSheet.kt
@@ -26,6 +26,7 @@ class IntentChooserSheet(
     }
 
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        val binding = binding
         binding.optionsRecycler.layoutManager = GridLayoutManager(context, 3)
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/github/libretube/ui/sheets/IntentChooserSheet.kt
```diff
@@ -26,6 +26,7 @@ class IntentChooserSheet(
     }
 
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        val binding = binding
```
Bnyro, line 29: What exactly is this doing?

Isira-Seneviratne: It's calling the non-null getter method and holding the value.

Bnyro: Interesting, never seen that before.

Bnyro (approved):

app/src/main/java/com/github/libretube/ui/sheets/IntentChooserSheet.kt
Bnyro: But makes sense, yeah.

PR #4634, opened 2023-08-26
feat(player): play music at 1x speed

FineFindus:
Changes the default video speed for Music videos to 1x. The user can still manually change the speed.
Please note that this can have the side effect of some non-music videos, which are categorized as music videos, being played at the wrong speed.

### DIFF
--- a/app/src/main/java/com/github/libretube/api/obj/Streams.kt
+++ b/app/src/main/java/com/github/libretube/api/obj/Streams.kt
@@ -93,6 +93,6 @@ data class Streams(
     }
 
     companion object {
-        var categoryMusic = "Music"
+        const val categoryMusic = "Music"
     }
 }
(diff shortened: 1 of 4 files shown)


Bnyro:
Please save "Music" to a constant. Otherwise lgtm.

FineFindus:
Sure. Should the constant be in the `PlayerFragment` or in the `Streams` file?

Bnyro:
I'd prefer having it in the `Streams` file as a companion object of `Streams`

app/src/main/java/com/github/libretube/api/obj/Streams.kt
```diff
@@ -91,4 +91,8 @@ data class Streams(
             shortDescription = description
         )
     }
+
+    companion object {
+        var categoryMusic = "Music"
```
Bnyro, line 96:
```suggestion
        const val categoryMusic = "Music"
```

Bnyro (approved):

PR #4651, opened 2023-08-28
feat: hide mark as watched when video is watched

FineFindus:
Hides the `Marked as watched` bottom sheet option, if the video is over 90% watched.

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/sheets/VideoOptionsBottomSheet.kt
+++ b/app/src/main/java/com/github/libretube/ui/sheets/VideoOptionsBottomSheet.kt
@@ -3,12 +3,13 @@ package com.github.libretube.ui.sheets
 import android.os.Bundle
 import androidx.navigation.fragment.NavHostFragment
 import com.github.libretube.R
-import com.github.libretube.api.RetrofitInstance
+import com.github.libretube.api.obj.StreamItem
 import com.github.libretube.constants.PreferenceKeys
 import com.github.libretube.db.DatabaseHelper
 import com.github.libretube.db.DatabaseHolder
 import com.github.libretube.db.obj.WatchPosition
 import com.github.libretube.enums.ShareObjectType
+import com.github.libretube.extensions.toID
 import com.github.libretube.helpers.BackgroundHelper
 import com.github.libretube.helpers.NavigationHelper
 import com.github.libretube.helpers.PlayerHelper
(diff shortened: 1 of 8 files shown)


Bnyro:
Instead of only passing the duration, we should pass the whole `StreamItem` to the `VideoOptionsBottomSheet`.

We should then also use it for saving the item to the watch history when marking as watched as well, instead of making an API request there (not exactly related to your PR, but I think this PR is a good reason to do these changes now).

FineFindus:
uhm, I actually did that first, but wasn't able to get it to work (probably just because of my (non-existing) Kotlin knowledge), so I switched to only passing the duration.

app/src/main/java/com/github/libretube/ui/sheets/VideoOptionsBottomSheet.kt
```diff
@@ -120,7 +119,9 @@ class VideoOptionsBottomSheet(
                         // add video to watch history
                         runCatching {
                             RetrofitInstance.api.getStreams(videoId)
-                        }.getOrNull()?.let { DatabaseHelper.addToWatchHistory(videoId, it) }
+                        }.getOrNull()?.let {
+                            DatabaseHelper.addToWatchHistory(videoId, it)
+                        }
```
FineFindus, line 124: This could potentially also use `StreamItem`, but it only saves a localized version of the `uploadDate` with no way of reversing it to a `LocalDate`.

FineFindus:
Updated it (seems like I only missed a null check).

app/src/main/java/com/github/libretube/ui/sheets/VideoOptionsBottomSheet.kt
Bnyro: It's not localized, it's just a string that is used for storing it. We can use the `uploaded` parameter of `StreamItem` and convert it to a `LocalDate`, since `uploaded` represents the milliseconds of the date.
```kotlin
val uploadDate = Instant.ofEpochMilli(epoch).atZone(ZoneId.systemDefault()).toLocalDate()
```

FineFindus: Hmm, when I tried to use parsed it to a `LocalDate`, it failed, because 23 hours ago is not a valid date.

Bnyro (approved):
Thanks for applying the changes, lgtm now!

Would be great if you could rebase the PR into two commits:
1. feat: hide mark as watched when video is watched
2. refactor: simplify VideoOptionsBottomSheet logic and improve its performance

Bnyro:
You can use `git rebase -i HEAD~7` and just squash some of the commits.

FineFindus:
That was the plan :)

Bnyro (approved):

PR #4661, opened 2023-08-29
feat: add option to save playback speed

FineFindus:
Adds a new setting for saving the playback speed when changing it. This can be useful when wanting to change the playback speed only for a shorter duration (a couple of videos), as it eliminates the need to manually change it for each video, or to go to the settings each time the user wants to change the default speed.

Screenshot of the settings page with a new 'Save playback speed' setting

### DIFF
--- a/app/src/main/java/com/github/libretube/constants/PreferenceKeys.kt
+++ b/app/src/main/java/com/github/libretube/constants/PreferenceKeys.kt
@@ -62,6 +62,7 @@ object PreferenceKeys {
     const val RELATED_STREAMS = "related_streams_toggle"
     const val CUSTOM_PLAYBACK_SPEED = "custom_playback_speed"
     const val PLAYBACK_SPEED = "playback_speed"
+    const val SAVE_PLAYBACK_SPEED = "save_playback_speed"
     const val BACKGROUND_PLAYBACK_SPEED = "background_playback_speed"
     const val FULLSCREEN_ORIENTATION = "fullscreen_orientation"
     const val PAUSE_ON_SCREEN_OFF = "pause_screen_off"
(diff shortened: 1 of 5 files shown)


FineFindus:
I agree, but I didn't want to just change the current behavior without warning. That's why I made it optional when enabling the setting.

Bnyro:
I'd like to keep this open for about a week - if nobody complains about the changes it's good to merge.

Bnyro:
(Please also remove the option from the preferences). The option for the background player playback speed should be kept nevertheless.

FineFindus:
> (Please also remove the option from the preferences). The option for the background player playback speed should be kept nevertheless.

I'm a bit confused by this. I already removed the option from the preferences and didn't touch the background speed playback?

Bnyro:
I'm referring to the playback speed option, which is the top one in your screenshot.

FineFindus:
I'm not sure whether that is a good idea. A new user may be confused a by the lack of an option to change the default playback speed. In that case, it would be fine to have a duplicated setting.
In the same vein, maybe it would be a good idea to add a toast, that the playback speed has been saved?

Bnyro:
I don't think we should show a toast for such trivial things. And neither I think  that new users would be confused by this.
If anyone is confused, then those are the people already using the app for a while and not finding the option anymore -> hence we mark it as a breaking change in the changelog.

Bnyro (approved):

CactiChameleon9:
Shame to see this change, but I understand that this is more user friendly

I liked how I could have a faster default speed that works for most video (x1.8) but then temporally adjust it up or down for one video if the pacing wasn't good enough

PR #4707, opened 2023-09-07
chore: Update androidx core

Isira-Seneviratne:


Bnyro (approved):
LGTM!

(note: I've been changing the PR title to the commit type `chore`, since I think that fits dependency updates better. `refactor` would be possible too)

PR #4748, opened 2023-09-10
fix(settings/sponsorblock): allow toggling of contribute setting

FineFindus:
Removes the dependence of the `sb_contribute_key` on itself, which blocked enabling the setting, as it required itself to be enabled first.

Disable contribute setting


Bnyro (approved):
Thanks!

PR #4749, opened 2023-09-10
feat(settings/sponsorblock): allow changing sb user id

FineFindus:
Adds a setting to change the used SponsorBlock user ID. Users may want to change (or view) it, as SponsorBlock keeps track of the amount of submitted segments.

Setting with SponsorBlock username setting

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -472,6 +472,7 @@
     <string name="segment_type">Segment type</string>
     <string name="sb_invalid_segment">Invalid segment start or end</string>
     <string name="contribute_to_sponsorblock">Contribute to SponsorBlock</string>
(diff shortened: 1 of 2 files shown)


Bnyro (approved):
Thanks!

app/src/main/res/values/strings.xml
```diff
@@ -472,6 +472,7 @@
     <string name="segment_type">Segment type</string>
     <string name="sb_invalid_segment">Invalid segment start or end</string>
     <string name="contribute_to_sponsorblock">Contribute to SponsorBlock</string>
+    <string name="sponsorblock_user_id">SponsorBlock Username</string>
```
ajayyy, line 475: Usernames are a seperate thing in SponsorBlock, can this be changed to "SponsorBlock UserID"?

PR #4827, opened 2023-09-21
fix: allow pausing when buffer is less than min buffer

FineFindus:
Fixes a regression in 42d081f04e132bc4ff5dbf408c596bd1f6927992.
Currently, the player only allows pausing when loading, when the current buffer is 0, which is only the case when no network connection is available. The player should pause as long as the current buffered time is less than the minimum buffer needed for playing.


Bnyro (approved):
Good catch, thanks!

PR #4829, opened 2023-09-21
feat: option to autoplay videos in playlist

FineFindus:
Currently, playing a playlist is only possible when autoplay is enabled.
This pr changes this and autoplays all videos in a playlist without autoplay being enabled.


FineFindus:
Autoplay plays the next video after the current one finishes. This is done regardless if the watched video was just a single video or part of a playlist.
Enabling autoplay will still continue to play the next video, this only affects the behavior if the user is watching a playlist.

Bnyro:
The same behavior can be achieved by disabling "Insert related items to queue" and enabling autoplay.

I don't see why we should automatically play the next videos even though the user explicitly disabled it (Autoplay is enabled by default).

FineFindus:
I would consider myself a fairly experienced user (I both use the app a lot and have contributed to it), and yet I had no idea that this combination of settings achieves the desired behavior. How is a new and inexperienced user supposed to discover this?
They might just disable autoplay because they don't like it, and later try to play a playlist, but it only plays one video at a time. From their experience, the application is obviously broken because it breaks with already established concepts in other applications (e.g. playlists in YouTube play all videos even if autoplay is disabled).

Also, I think the user group that clicks on a playlist and expects it to play only one video at a time is pretty small or nonexistent.

Bnyro:
> I would consider myself a fairly experienced user (I both use the app a lot and have contributed to it), and yet I had no idea that this combination of settings achieves the desired behavior. How is a new and inexperienced user supposed to discover this?
>
I don't expect any new user to discover this option, however:
- we have an open issue that focuses on improving the organization of the settings
- the user can temporarily enable autoplay

I'd be okay with adding an option to autoplay playlists.

Nevertheless this should also affect the background mode / audio only player too.

FineFindus:
Added a setting for it and changed the code in `OnlinePlayerService` (I this one is responsible for background play?).
Not sure how to support the audio only player, as it doesn't even support autoplay.

app/src/main/java/com/github/libretube/helpers/PlayerHelper.kt
```diff
@@ -323,6 +323,10 @@ object PlayerHelper {
                 true
             )
 
+    fun shouldPlayNextVideo(playlistId: String?): Boolean {
```
Bnyro, line 326:
```suggestion
    fun isAutoPlayEnabled(isPlaylist: Boolean = false): Boolean {
```

Bnyro:
> Added a setting for it and changed the code in `OnlinePlayerService` (I this one is responsible for background play?). Not sure how to support the audio only player, as it doesn't even support autoplay.
>
Just make `PlayerHelper.autoPlay` a private var and replace it where the compilation fails.

The background / audio only mode does support autoplay though.

Bnyro:
> Bnyro you added thousand lines for this 😂
> 
> photo_2023-09-25_16-14-08
>
That's unrelated to this issue, and in fact it's mainly useful for the banner of a channel, not its avatar.

FineFindus:
I already replaced it everywhere, except for the autoplay switches themselves.

> The background / audio only mode does support autoplay though.

Hmm, odd, I can't get it to work

app/src/main/res/values/strings.xml
```diff
@@ -340,6 +340,8 @@
     <string name="bookmark">Bookmark</string>
     <string name="clear_bookmarks">Clear bookmarks</string>
     <string name="queue_insert_related_videos">Insert related videos</string>
+    <string name="autoplay_playlists">Autoplay playlists</string>
+    <string name="autoplay_playlists_summary">Play next video in playlists regardless of autoplay settings</string>
```
Bnyro, line 344:
```suggestion
    <string name="autoplay_playlists_summary">Automatically play next video in playlists regardless of selected autoplay setting</string>
```

app/src/main/res/xml/player_settings.xml
```diff
@@ -133,6 +133,12 @@
             android:title="@string/queue_insert_related_videos"
             app:key="queue_insert_related_videos" />
 
+        <SwitchPreferenceCompat
+            android:defaultValue="true"
```
Bnyro, line 137: This should default to false, since most users expect autoplay to impact the normal queue and PLAYLISTS. This way, if users disable autoplay via the player UI, it might appear like a bug that playlists still autoplay to new users.

app/src/main/java/com/github/libretube/helpers/PlayerHelper.kt
```diff
@@ -323,6 +323,10 @@ object PlayerHelper {
                 true
             )
 
+    fun shouldPlayNextVideo(isPlaylist: Boolean = false): Boolean {
+        return autoPlayEnabled || (PreferenceHelper.getBoolean(PreferenceKeys.AUTOPLAY_PLAYLISTS, true) && isPlaylist)
```
Bnyro, line 327:
```suggestion
        return autoPlayEnabled || (PreferenceHelper.getBoolean(PreferenceKeys.AUTOPLAY_PLAYLISTS, false) && isPlaylist)
```
see below

Bnyro (approved):

PR #4924, opened 2023-10-06
fix: Properly parse date strings

Isira-Seneviratne:
Properly parse the new date format, which contains a time offset.

### DIFF
--- a/app/src/main/java/com/github/libretube/api/obj/Streams.kt
+++ b/app/src/main/java/com/github/libretube/api/obj/Streams.kt
@@ -2,18 +2,19 @@ package com.github.libretube.api.obj
 
 import com.github.libretube.db.obj.DownloadItem
 import com.github.libretube.enums.FileType
-import com.github.libretube.extensions.toLocalDateSafe
 import com.github.libretube.extensions.toMillis
 import com.github.libretube.helpers.ProxyHelper
+import com.github.libretube.json.SafeLocalDateSerializer
 import com.github.libretube.parcelable.DownloadData
-import kotlin.io.path.Path
+import kotlinx.datetime.LocalDate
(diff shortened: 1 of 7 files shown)


Isira-Seneviratne:
> today when i created a branch to work on the instance button issue, i merged this and app is crashing with same datetimeparse errors, but the current alpha with new dates is working fine!

What's the date format displayed in the error message?

Bnyro:
We currently still only use the date without time. Wouldn't it make more sense to use `Instant` for everything (so that hours, minutes and seconds are taking into consideration too) and parse `LocalDate` instances to an `Instant` at `00:00` instead?

Isira-Seneviratne:
> We currently still only use the date without time. Wouldn't it make more sense to use `Instant` for everything (so that hours, minutes and seconds are taking into consideration too) and parse `LocalDate` instances to an `Instant` at `00:00` instead?

Yeah, it would.

app/src/main/java/com/github/libretube/db/obj/WatchHistoryItem.kt
```diff
@@ -10,7 +10,7 @@ import kotlinx.serialization.Serializable
 
 @Serializable
 @Entity(tableName = "watchHistoryItem")
-data class WatchHistoryItem(
+data class WatchHistoryItem constructor(
```
Bnyro, line 13: What is this change needed for?

Bnyro (approved):

app/src/main/java/com/github/libretube/db/obj/WatchHistoryItem.kt
Isira-Seneviratne: I accidentally committed that change.

PR #4951, opened 2023-10-11
fix: wrong comment opening bug fixed

### LINKED ISSUE #4428: Sometimes clicking on comments opens the wrong replies
### Steps to reproduce

1. Open the comments section of a video
2. Scroll down few pages
3. Open few comments' replies and notice the content of the parent comment and its author
4. After opening the replies, the wrong comment/author is shown
5. After clicking the "back" button, it will show that the comment you originally clicked on doesn't have any replies

Here is a recorded video to show the issue:

https://github.com/libre-tube/LibreTube/assets/49586511/bcb36ca5-348a-4587-bc7a-f9338d71f45a

### Expected behavior

Show the replies of the comment I clicked

### Actual behavior

Shows the replies of another comment

### LibreTube version

0.14.1, also in Master branch

### Android version

Android 13

### Other details

I'm still not familiar enough with Kotlin and the codebase of the project to figure out why this is happening. I also suspect this has to do with #3676 and #4291 (I did fix #4291 using some workaround but I still suspect that this is the cause somehow).

### Acknowledgements

malik0011:
**About fix:** When we are fast scrolling through comments and suddenly click on any comment, sometimes it opens the wrong comment's reply.

**Reason:**  This is only happening on fast scrolling as per my view. The main reason is the position we getting in onViewBinding is wrong. As per my knowledge, it happens when the recycle view recycleView can't find unique layout(on fast scroll). 

**Solution:** This approach means that each item will have a unique view type based on its position in the RecyclerView. So that, also on fast scroll rcv can identify unique views(not the optimised approach but using this approach we can solve this bug). 

Closes #4428

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/adapters/ChaptersAdapter.kt
+++ b/app/src/main/java/com/github/libretube/ui/adapters/ChaptersAdapter.kt
@@ -26,7 +26,7 @@ class ChaptersAdapter(
     }
 
     override fun onBindViewHolder(holder: ChaptersViewHolder, position: Int) {
-        val chapter = chapters[position]
+        val chapter = chapters[holder.absoluteAdapterPosition]
         holder.binding.apply {
             if (chapter.highlightDrawable != null) {
                 chapterImage.setImageDrawable(chapter.highlightDrawable)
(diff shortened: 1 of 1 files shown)


salahmak:
@malik0011 I'll try it when I have some free time, thanks!

Bnyro (approved):
Looks good from my side, thank you :+1:

Isira-Seneviratne (approved):
Looks good to me as well.

PR #4955, opened 2023-10-11
feat(comments): show if creator replied

FineFindus:
Shows the uploader's channel thumbnail next to a comment's reply count if the creator has replied.

Requires a server that has merged https://github.com/TeamPiped/Piped-Backend/pull/718, as the official one doesn't yet, I haven't been able to test if everything works, so I've marked it as a draft for now.

Comment with creator thumbnail juxtopposed to reply count


Bnyro (approved):
Awesome, thank you!

Issue #5025, opened 2023-10-22
[FR]: Continue watching in library tab


Silther:
### Describe your suggested feature

Please add the list of videos you didn't finish watching (called `continue watching` in the home tab) to the library tab in the same style as `watch history` and `downloads` (show a but saying continue watching and if you click it, it should open a new page with all those videos).

I have hidden the home tab in my app, because I don't want to see the trending videos and everything else (except the `continue watching`) is already in the library tab.

### Other details

![signal-2023-10-22-145435](https://github.com/libre-tube/LibreTube/assets/97352673/0c5c5dfc-1f74-41f2-a2ec-fb3b266e9b32)

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

Bnyro:
In my opinion we should integrate this with a filter in the watch history, so that there's e.g. a checkbox that is called "only show unfinished videos". A dedicated screen doesn't make much sense.

You can work on this if you want to (should be easy), but I don't really see the need to keep the issue open because it wouldn't be addressed without your contribution.

PR #5032, opened 2023-10-23
feat: filter options for Watch History Tab

Silther:
(one day) implements #5025


Bnyro:
That's just a lot of duplicated logic, it would he way better to add a boolean argument `showWatched` to the `WatchHistoryFragment` instead.

Bnyro:
And as I said, I prefer to add the functionality to the existing watch history tab and not add a dedicated button for "continue watching" in the library.

Bnyro:
I would place a checkbox below the "play all" and 'clear history" button.

It's okay if it's not remembered and needs to be enabled manually each time you open the watch history. It still could be changed later if we ever decided to.

Bnyro:
I just wanted to say that we should avoid duplicated code by that.

Bnyro:
We need database migrations for that, I can push the necessary changes (because it's kinda hard to explain). I'm not at home until Monday however, so you gotta be patient.

app/src/main/java/com/github/libretube/constants/PreferenceKeys.kt
```diff
@@ -130,7 +131,7 @@ object PreferenceKeys {
     const val LAST_WATCHED_FEED_TIME = "last_watched_feed_time"
     const val HIDE_WATCHED_FROM_FEED = "hide_watched_from_feed"
     const val SELECTED_FEED_FILTER = "filer_feed"
-    const val FEED_SORT_ORDER = "sort_oder_feed"
+    const val FEED_SORT_ORDER = "sort_order_feed"
```
Bnyro, line 134: I'll undo this because this will cause all users to lose the previously saved sort order when updating. It's not nice to have, but we can live with that typo.

Bnyro (approved):
Thank you!

I've placed three commits on top of yours:
1. did add the necessary database migrations
2. made some code cleanups to remove some duplicated code
3. automated formatting to make the linter happy

Issue #5141, opened 2023-11-12
Videos stop playing when locked, because Libretube is put to sleep


lu4p:
### Steps to reproduce

1. Play any video
2. Lock your screen
3. Wait 5 minutes (This may depend on how aggressive the device is with power management)

### Expected behavior

Video should continue playing.

### Actual behavior

Video stops playing.

### LibreTube version

0.19.0

### Android version

Android 13

### Other details

This is because Libretube needs to acquire a wakelock https://developer.android.com/training/scheduling/wakelock
in order for Android to not put Libretube to sleep after a few minutes.

Exoplayer can acquire this wakelock for us when it makes sense (i.e. video playing while the screen is off), so I thought I had a 10 line fix for this.
```kotlin
// for online usage
player.setWakeMode(WAKE_MODE_NETWORK)
// for offline usage
player.setWakeMode(WAKE_MODE_LOCAL)
```
https://developer.android.com/reference/androidx/media3/exoplayer/ExoPlayer#setWakeMode(int)

However when debugging afterwards I learned that exoplayer can only do this when it runs in a service like is already done in the audio player.

This should probably be implemented by extending the Offline and OnlinePlayerServices to also support video, and offloading all PlayerFragement player interactions to it.

The officially recommended solution is a [mediasessionservice](https://developer.android.com/guide/topics/media/session/mediasessionservice), however that would require a major refactor which is probably not something that can be easily done (I tried and failed miserably, but I also know little about android development).

I will send a PR to fix the audio player, it is less affected because it is already running as a foreground service, but will also be put to sleep after a longer interval or when Battery Saver is active.

I'm also trying to fix the video player but I'm not sure if I will succeed.

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

Bnyro:
PRs welcome, as I already said I don't have interest in working on that since the audio only mode is good enough for the purpose of background playback.

Pamilg8:
I can say that there are the following issues with playback
1.Sometimes when opening a video, the player is locked and the play button does not work. Reopening the video does not fix it
2.The player and related videos section is not loading
3.↻ appears and the player does not open
4.The video stops in the middle and the play does not work.
5.Sometimes the 4th problem occurs when playing video forward and backward.
6.The player restarted,when I came back from the background.Or the playback stopped.

[...]

PR #5142, opened 2023-11-12
fix: background service audio playback when screen locked

lu4p:
updates #5141


Bnyro:
You've only modified the audio only player services which are being used when using the "play in background" mode. As far as I know, these services don't have any issues with being put to sleep automatically.

The changes don't have an impact on the linked issue. Is that intended? 

(The AndroidManifest changes are necessary for Android 14 anyways, so these are fine)

Bnyro (approved):
Merging anyways because this won't have any negative impact anyways, thanks!

johnaweiss:
Hi, i have LibreTube set to "Audio Only" in Settings, but it's stopping if i lock the phone. Any fix? - Thx

PR #5159, opened 2023-11-15
fix: properly apply creator highlight

FineFindus:
Fixes a problem where, due to the way recycler views work, the show highlight was not reset properly, causing the highlight to be displayed on the wrong users.


Bnyro (approved):
ty!

PR #5171, opened 2023-11-16
fix: restoring nav after process death

Pittvandewitt:
This pr fixes restoring from process death. LibreTube navigates to the start fragment after process death, which is not necessary as fragments are restored automatically. On top of that, the bottom navigation bar has its own mechanism of saving and restoring. This sometimes leads to an inconsistent UI state if the last selected navigation item before process death does not have the start fragment as root. The selected menu item does not reflect the displayed tab in this case.

This change actually sets the newly created navigation graph and removes the workaround of navigating to the start fragment to display it. The main activity xml was edited to prevent inflating the navigation graph twice.


Bnyro (approved):
Good catch, thank you for the PR!

PR #5174, opened 2023-11-17
fix: auto fullscreen stops working after first rotation

Zoo-M0:
Fix orientation bug where screen get stuck after first orientation change.

Bug was in an if that where removing `SCREEN_ORIENTATION_SENSOR` from `mainActivity.requestedOrientation` ignoring autoFullscreenEnabled prefs.

I'm not sure if I've done some regression on other things as I don't know yet all the app behaviours, every suggestion is welcome.


Zoo-M0:
I'm having problems with ktlint and I can't run it locally to be sure it's ok, I suggest to use detect with maybe git hooks to run it in pre-commit or pre-push

Bnyro:
Please run `ktlint --android -F` next time locally before creating the PR

Bnyro (approved):
Thanks!

Issue #5095, opened 2023-11-02
Increase subtitle font size in fullscreen


kamlendras:
### Steps to reproduce

1. Play a video
2. Turn on subtitles 
3. Everything looks fine
4. 
![Screenshot_20231103-011938~2](https://github.com/libre-tube/LibreTube/assets/96082996/3ad7066f-0989-467b-aee0-b988e6e4f16f)
5. Switch to full screen mode
6. 
![Screenshot_20231103-012005](https://github.com/libre-tube/LibreTube/assets/96082996/036e50ad-6792-479c-8596-b0504611d2f3)
As you can see here subtitles does not expand .

### Expected behavior

Subtitles should expand on full screen mode .

### Actual behavior

Subtitles does not expand on full screen mode.

### LibreTube version

Latest Stable

### Android version

AOSP

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

kamlendras:
Thanks! Issue solved 😀

PR #5200, opened 2023-11-19
fix: Increase subtitle font size in fullscreen

saymanrifat:
#5095 fix: Increase subtitle font size in fullscreen

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/views/CustomExoPlayerView.kt
+++ b/app/src/main/java/com/github/libretube/ui/views/CustomExoPlayerView.kt
@@ -692,16 +692,26 @@ open class CustomExoPlayerView(
     }
 
     override fun onFullscreenChange(isFullscreen: Boolean) {
-        if (PlayerHelper.swipeGestureEnabled && this::brightnessHelper.isInitialized) {
-            if (isFullscreen) {
+        if (isFullscreen) {
+            if (PlayerHelper.swipeGestureEnabled && this::brightnessHelper.isInitialized) {
                 brightnessHelper.restoreSavedBrightness()
-                if (resizeMode == AspectRatioFrameLayout.RESIZE_MODE_ZOOM) {
-                    subtitleView?.setBottomPaddingFraction(SUBTITLE_BOTTOM_PADDING_FRACTION)
-                }
-            } else {
-                brightnessHelper.resetToSystemBrightness(false)
-                subtitleView?.setBottomPaddingFraction(SubtitleView.DEFAULT_BOTTOM_PADDING_FRACTION)
             }
+            subtitleView?.setFixedTextSize(
+                Cue.TEXT_SIZE_TYPE_ABSOLUTE,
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/github/libretube/helpers/PlayerHelper.kt
```diff
@@ -289,6 +289,12 @@ object PlayerHelper {
             "18"
         ).toFloat()
 
+    val fullscreenCaptionsTextSize: Float
+        get() = PreferenceHelper.getString(
+            PreferenceKeys.CAPTIONS_SIZE,
+            "27"
+        ).toFloat()
```
Bnyro, line 296: You didn't add a preference for that yet, so the user can't change it. And  I think we should just make the font size 1.5 times bigger instead of adding a new preference.

Bnyro:
I think we should just make the font size 1.5 times bigger in landscape than usually instead of adding a new preference.

app/src/main/java/com/github/libretube/helpers/PlayerHelper.kt
saymanrifat: Sure.

app/src/main/java/com/github/libretube/ui/views/CustomExoPlayerView.kt
```diff
@@ -695,11 +695,19 @@ open class CustomExoPlayerView(
         if (PlayerHelper.swipeGestureEnabled && this::brightnessHelper.isInitialized) {
```
Bnyro, line 695: Due to this line this will only be executed when swipe gestures are enabled.
Please move this condition around the `brightnessHelper.restoreSavedBrightness()` calls below to fix this.

Otherwise lgtm.

saymanrifat: Like this?
Screenshot 2023-11-19 at 3 17 02 PM

Bnyro: Yes, exactly. Please do the same in the `else` block where `brightnessHelper.resetToSystemBrightness()` is called, then we're good to merge this :+1:

app/src/main/java/com/github/libretube/ui/views/CustomExoPlayerView.kt
```diff
@@ -692,9 +692,10 @@ open class CustomExoPlayerView(
     }
 
     override fun onFullscreenChange(isFullscreen: Boolean) {
-        if (PlayerHelper.swipeGestureEnabled && this::brightnessHelper.isInitialized) {
             if (isFullscreen) {
+                if (PlayerHelper.swipeGestureEnabled && this::brightnessHelper.isInitialized) {
                 brightnessHelper.restoreSavedBrightness()
```
Bnyro, line 697: nit: indentation missing

Bnyro (approved):
Thank you!

PR #5234, opened 2023-11-28
chore: add abort on lint function

IndusAryan:
looks like there are too many string issues, but this can be merged for future aspects as it prevents many lint issues locally


IndusAryan:
looks like there are too many string issues, but this can be merged for future aspects as it prevents many lint issues locally

Bnyro (approved):

PR #5354, opened 2023-12-19
feat(ui): add playback speed shortcut

### LINKED ISSUE #5349: make easier to switch playing speed

#5346 is out of scope. I wonder if its possible to modify the interface a lil so they are not 7 taps away as switching speed is a very common use case

Ideas:
* add 1.75x button as is very common speed
* letting you select the speed after clicking the video. This will make every video 2 taps instead of 1, and can be added as a option setting
* show default speed buttons directly after you tap the show options on the top right corner
* show a speed bar you can drag as direct option of the video

I can probably come up with more ideas to this issue if none of this are suitable :)

Thank you for your consideration and thanks for this great program!

### Other details

_No response_

### Acknowledgements

IndusAryan:
Closes #5349 

youtube also uses this as this is most useful among options, and about the cluttered experience, i think the lock button in portrait mode and the sponssorblock toggle shortcut are utter useless .
close if you dont want, if merged i can remove useless sponsor toggle and lock button when player is not in fullscreen.
youtube uses more options like shuffle controls etc. but we have queue sheet for that
this can be moved also beside timebar beside total video length.

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
+++ b/app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
@@ -514,13 +514,13 @@ class PlayerFragment : Fragment(), OnlinePlayerOptions {
         }
 
         playerBinding.speedShortcut.setOnClickListener {
-            PlaybackOptionsSheet(exoPlayer).show(childFragmentManager,null)
+            PlaybackOptionsSheet(exoPlayer).show(childFragmentManager, null)
         }
 
         playerBinding.speedShortcut.setOnLongClickListener {
             val defaultSpeed = 1f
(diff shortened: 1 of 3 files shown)


Bnyro:
2 things that could/should be improved:

1. The button should only be visible when there's a different playback speed set, not when you're using the default one.
2. The button shouldn't always set the speed to 1.0: Instead it should set the speed the "Custom playback speed" preference value if it is enabled, and otherwise, if it's not enabled, to 1.0f.

IndusAryan:
the 1st point is against the main motive of this feature, this feature is all about handy settings that immediately change speed if default one feels slow or fast

Bnyro:
Then this is not planned due to the reasons provided in https://github.com/libre-tube/LibreTube/issues/5349#issuecomment-1860944660.

Issue #5355, opened 2023-12-19
Error on trying to change Playback speed on Settings.


RafaelsRamos:
### Steps to reproduce

1. Ensure the setting **Player** -> **Defaults** -> **Remember playback speed** is enabled;
2. Go to a video and change the Playback speed to an odd number (e.g. 1.05);
3. Go to **Player** -> **Defaults** -> **Remember playback speed** and disable it;
4. Try manually setting a new Playback speed;

https://github.com/libre-tube/LibreTube/assets/40279132/d5361853-cb89-4297-9cd6-af8d70273592

### Expected behavior

The error does not occur.

### Actual behavior

The error dialog shows up, and in occasion the app crashes.

### LibreTube version

0.20.1 Debug

### Android version

Android 12

### Other details

I will open a PR with a possible fix.

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

PR #5356, opened 2023-12-19
fix: Error on Settings > Player > Playback speed

RafaelsRamos:
When a SliderPreference is processing a preference with a value outside the supported values - **valueFrom + x * stepSize** an error is shown. This is happening on the issue #5355 

We could go for 3 approaches:
1. Change the step on the settings (the issue might persist on other SliderPreference settings);
2. Change the step on the video settings (users currently set on an odd number can be impacted);
3. Add a safeguard on the SliderPreference component itself;

I went for the 3. but I understand if you think we should go for 1. or 2. - Let me know which approach is more suitable 😊

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/views/SliderPreference.kt
+++ b/app/src/main/java/com/github/libretube/ui/views/SliderPreference.kt
@@ -34,8 +34,10 @@ class SliderPreference(
     }
 
     private var prefValue: Float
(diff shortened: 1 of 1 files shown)


Bnyro:
I would prefer solution 1, the `stepSize` for the preference is defined in `player_settings.xml`.

Thanks for your investigations and your PR!

RafaelsRamos:
Great! I'll close this one and open a new open with that tweak.

PR #5364, opened 2023-12-20
feat: add copy video ID button in stats sheet

FineFindus:
Adds a copy button for the video ID to the stats sheet. This can be useful for debugging, e.g. to quickly check the response returned by the API. The current alternative is to share the video and extract the ID, which is quite cumbersome.

Stats sheet showing the video ID with a copy icon after it


app/src/main/java/com/github/libretube/ui/sheets/StatsSheet.kt
```diff
@@ -33,7 +36,12 @@ class StatsSheet : ExpandedBottomSheet() {
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         val binding = binding
 
+        val clipboard = context?.let { getSystemService(it, ClipboardManager::class.java) }
```
Bnyro, line 39:
```suggestion
        val clipboardManager = context?.getSystemService(ClipboardManager::class.java)!!
```
The `context` can't be null here since we're in `onViewCreated` (at the beginning of the lifecycle), so it can't yet be destroyed.

FineFindus: Yeah, I kind of suspected that, but I wasn't quite sure, so I figured better safe than sorry :)

Bnyro (approved):
Thanks!

PR #5366, opened 2023-12-20
fix: improve constrast between background and buttons

### LINKED ISSUE #5363: [master] Buttons in player page almost have no background
### Steps to reproduce

1. Open a video
2. Take a look on the buttons below the player (share, download, save, audio, comments)

### Expected behavior

They should be noticeable as buttons with visible outlines.

### Actual behavior

The colour of buttons is the same as the color of surrounding interface.

Screenshot

### LibreTube version

Nightly 0.12.1 commit a9b58d9

### Android version

Android 13

### Other details

I suggest that the buttons should have the same look as the "Subscriptions" button at the top of the "Subscriptions" tab (sorry for the pun).

### Acknowledgements

RafaelsRamos:
I tried many possible colours but ultimately decided to go for `android:background`. These were the reasons:
- With `android:background`, the colour matches the status bar;
- I think it is one of the most aesthetically appealing combinations;
- YouTube follows a similar approach of a darker background with lighter buttons.

Here are the possible colours - [Link to material-components colours](https://github.com/material-components/material-components-android/blob/master/docs/theming/Color.md)

Below are all the combinations I checked. Please let me know if we should choose another colour.

### android:background

closes #5363

### DIFF
--- a/app/src/main/res/layout-land/fragment_player.xml
+++ b/app/src/main/res/layout-land/fragment_player.xml
@@ -11,7 +11,7 @@
         android:id="@+id/player_scrollView"
         android:layout_width="0dp"
         android:layout_height="0dp"
-        android:background="?colorSurfaceContainerLow"
+        android:background="?android:colorBackground"
         app:layout_constraintBottom_toBottomOf="parent"
         app:layout_constraintEnd_toStartOf="@id/related_container"
         app:layout_constraintHorizontal_bias="0.5"
--- a/app/src/main/res/layout/fragment_player.xml
+++ b/app/src/main/res/layout/fragment_player.xml
@@ -11,7 +11,7 @@
         android:id="@+id/player_scrollView"
         android:layout_width="match_parent"
         android:layout_height="0dp"
-        android:background="?colorSurfaceContainerLow"
+        android:background="?android:colorBackground"
         app:layout_constraintBottom_toBottomOf="parent"
         app:layout_constraintEnd_toEndOf="parent"
         app:layout_constraintHorizontal_bias="0.5"


IndusAryan:
what is the sense of having pure amoled mode if the main fragment of player isnt pure black

Bnyro:
> what is the sense of having pure amoled mode if the main fragment of player isnt pure black
>
We will override the color we decide for in the end to become black in the amoled theme too ...
It's just about the normal theme here.

Bnyro:
The reason for the background color changes were the design ideas in https://github.com/libre-tube/LibreTube/discussions/5247.

I think that generally we should choose a dark background color, but I feel like `android:colorBackground` is still too bright compared to the draft in the linked discussion.

It seems like `colorSurfaceDim` is quite a good solution from what I've tried so far, but that's just my POV :thinking:

IndusAryan:
the design is very good, but when the pure mode is on, the colors should amoled black everywhere, irrespective of the color theme, because it is meant for it. plus i am unable to find the thread and recall what was your response to removal of vertical related items , the horizontal one looks ugly af and doesnt justify the guidelines of an ideal video streaming app.

DanielProg39:
`colorSurfaceDim` and `android:background` have identical colour schemes in dark mode, which I like best, but the light mode of `colorSurfaceDim` is more appealing to me, so I prefer this option.

RafaelsRamos:
`colorSurfaceDim` was my second favourite.

If you guys think that approach is best let me know 👍

DanielProg39:
As I said, I like the light mode of `colorSurfaceDim` a bit more, but I only use the dark mode, so my thoughts on the light mode are not really appropriate, so it doesn't make much difference to me, I would be happy with either of these options :)

So in general, I think it's more up to the users of the light mode to decide which option is better. At the very least we can open a discussion about it to get more attention and feedback.

M00NJ:
I'm the person who created the design ideas. Thanks for the ping @RafaelsRamos :)
Like it was mentioned `android:background` and

M00NJ:
I'm the person who created the mock-ups for the new design. Thanks for the ping @RafaelsRamos . 
Like it it was said `android:colorBackground` and `colorSurfaceDim` are the same in dark mode but the later will keep it's relative brightness in light mode, so it won't be the brightest thing on screen. This can make sense for a streaming app, though with `colorSurfaceDim` the status bar color won't be the same in light mode but with `android:colorBackground` it will. So for that reason I would stick with the default background color. However just like @DanielProg39 I don't use light mode, so my opinion should be weighed to much here.

RafaelsRamos:
Thank you for your input, M00NJ ! 🙌 

After re-reading all comments, we have a tie between the 2. Because M00NJ created the mockups, I propose that he gets the highest weight in the discussion.

Please, let me know if you guys disagree.

Bnyro (approved):

DanielProg39:
I also thought about `android:colorBackground` as an equal option because of the status bar color, but in the end I don't think that this really matters and has much impact (I think a brighter status bar won't be an issue at all for the vast majority of users), so I'd rather look at this from an aesthetic perspective.

That said, although I (aesthetically) like `colorSurfaceDim` a bit more, I'm fine with both options and I don't mind either of them being taken, so I'm OK with @RafaelsRamos' suggestion

PR #5367, opened 2023-12-21
refactor(fix): fix black screen and delay when unsetting fullscreen

IndusAryan:
also migrate buildconfig to dsl from gradle properties as it will get removed in gralde 9.0 release

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -32,7 +32,7 @@
         android:roundIcon="@mipmap/ic_launcher_round"
         android:supportsRtl="true"
         android:theme="@style/StartupTheme"
-        tools:targetApi="tiramisu">
+        tools:targetApi="n">
 
         <activity
             android:name=".ui.activities.WelcomeActivity"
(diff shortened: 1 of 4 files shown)


IndusAryan:
fu*k you reviewdog ktlint

app/src/main/AndroidManifest.xml
```diff
@@ -32,7 +32,7 @@
         android:roundIcon="@mipmap/ic_launcher_round"
         android:supportsRtl="true"
         android:theme="@style/StartupTheme"
-        tools:targetApi="n">
```
Bnyro, line 35: That's just telling the IDE to not complain if some of the used attributes don't work for all SDKs we support. Hence we should only change that if required to suppress incompatibility warnings, it's a bad idea to do so now.

Bnyro:
Can confirm that fixes the bug :+1:

Bnyro (changes requested):
See comment above.

Bnyro (approved):

IndusAryan:
wdym, i dont get you, what will happen if its set to tiramisu @Bnyro

Bnyro:
> wdym, i dont get you, what will happen if its set to tiramisu @Bnyro
>
Nothing, but it's a bad practice.

PR #5378, opened 2023-12-22
fix: Make setting's descriptions more consistent

### LINKED ISSUE #5377: Inconsistency in settings descriptions
### Steps to reproduce

Some settings' descriptions with a period while others do not. It would be great for all settings to be consistent.

The Settings on Android devices' OS and Youtube App all end without a period:

### Propostal

I propose we follow the same approach.

### Expected behavior

Consistency between all settings.

### Actual behavior

Some end up with "." others do not.

### LibreTube version

0.20.1

### Android version

Android 13

### Other details

_No response_

### Acknowledgements

RafaelsRamos:
This PR:
- Removes a few unused translations;
- Removes final periods from settings descriptions (in all languages);
- Changes **only English** translation of `disable_proxy_summary` from "**Load videos and images directly from YouTube\'s servers. Only enable the option if you use a VPN anyways!**" to "**Load videos and images directly from YouTube\'s servers. Only enable the option if you always use a VPN**".

### Sample

https://github.com/libre-tube/LibreTube/assets/40279132/162c190d-1526-45a9-8bbf-45bfd81c0a46

### Notes:
- Pay special attention to the Sample.mp4 at the 11th second mark. As you can see, the descriptions are very large and contain a period in the middle - making it a complex sentence. Usually, these descriptions are simple, so I suggest we take this into consideration for future settings.
- I removed all periods in the end due to the points mentioned at #5377. Please let me know if you disagree with this approach.

closes #5377


Bnyro (approved):
Thank you for taking a look into that.

That's really one of the things I've already wanted to look into some time ago but I've just been too lazy to spend time with it :)

PR #5380, opened 2023-12-23
refactor: Use Navigation Safe Args

Isira-Seneviratne:


IndusAryan:
@Isira-Seneviratne what does this do actually ?

Isira-Seneviratne:
> @Isira-Seneviratne what does this do actually ?

It allows navigating between fragments to be performed in a type-safe manner: https://developer.android.com/guide/navigation/use-graph/pass-data

RafaelsRamos:
My 2 cents. From my experience, SafeArgs brings to the table Safety at the expense of complexity.

By adding SafeArgs to a single navigation, we increase entropy and have 2 ways of handling arguments between fragments. If we decide to follow this approach, I suggest that we consider applying it to all other navigations.

Isira-Seneviratne:
> My 2 cents. From my experience, SafeArgs brings to the table Safety at the expense of complexity.
> 
> By adding SafeArgs to a single navigation, we increase entropy and have 2 ways of handling arguments between fragments. If we decide to follow this approach, I suggest that we consider applying it to all other navigations.

I've made the changes for existing navigations that require arguments, e.g. the channel fragment.

Bnyro (approved):

PR #5392, opened 2023-12-26
feat: support close video on MiniPlayer swipe down

### LINKED ISSUE #2020: Swipe down to close miniplayer

When the video player is minimised on the app, one can close the video by further swiping down the player just like YouTube and NewPipe.

Screenshot_20221124-114602~2

### Other details

_No response_

### Acknowledgements

RafaelsRamos:
Adds support for closing the video being played when the user downswipes the MiniPlayer.

https://github.com/libre-tube/LibreTube/assets/40279132/4d0a4a0a-ab84-4ec3-9c19-c51852dedd41

closes #2020

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/extensions/ViewAnimations.kt
+++ b/app/src/main/java/com/github/libretube/ui/extensions/ViewAnimations.kt
@@ -1,0 +1,23 @@
+package com.github.libretube.ui.extensions
+
+import android.view.View
+
+/**
+ * This function animates a view in a downward direction by a specified amount.
+ *
+ * @param duration The duration of the animation in milliseconds.
+ * @param dy The distance to move the view along the Y-axis.
+ * @param onEnd An optional lambda function that is invoked when the animation ends.
+ */
+fun View.animateDown(
+    duration: Long,
+    dy: Float,
+    onEnd: () -> Unit = { }
+) {
+    this
+        .animate()
+        .withEndAction { onEnd.invoke() }
+        .y(dy)
+        .setDuration(duration)
+        .start()
+}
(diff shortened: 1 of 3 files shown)


Pamilg8:
Wouldn't it be nice to add this option as on/off in Settings?

NeeRaj-2401:
> Thank you👍 Wouldn't it be nice to add this option as on/off in Settings?

Bruhh 💀, there is already plenty of settings that are customisable, this small behaviour isn't worth to add one

akano12:
> > Thank you👍 Wouldn't it be nice to add this option as on/off in Settings?
> 
> Bruhh 💀, there is already plenty of settings that are customisable, this small behaviour isn't worth to add one 

Nah let's add a new one, there aren't enough options/settings to choose from.


And jokes aside, if you don't want to use then don't swipe down tk close the mini player maybe 🙄🙄🙄🙄

IndusAryan:
> Thank you👍 Wouldn't it be nice to add this option as on/off in Settings?

yes, i think it needs a preference , but i hate miniplayer i think there should be a setting to direclty close video when it is gestured down to miniplayer. would be awesome to have it.

Pamilg8:
> > Thank you👍 Wouldn't it be nice to add this option as on/off in Settings?
> 
> yes, i think it needs a preference , but i hate miniplayer i think there should be a setting to direclty close video when it is gestured down to miniplayer. would be awesome to have it.

Why the 👎emoji?

RafaelsRamos:
My 2 cents, I agree with NeeRaj-2401 and akano12, in the sense that this is such a small behaviour, it might not be worth adding one more setting to enable/disable it.
That being said, if we deem it the best approach, I am willing to add it 👍

IndusAryan:
> My 2 cents, I agree with NeeRaj-2401 and akano12, in the sense that this is such a small behaviour, it might not be worth adding one more setting to enable/disable it.
> That being said, if we deem it the best approach, I am willing to add it 👍 

my 10 cents, i think this should be a setting , just like many ultra useful settings like different qualities for fullscreen and normal screen i think i also need another quality setting for miniplayer, i hope devs add it, maybe i should add it or open issue.

RafaelsRamos:
What do you think Bnyro, should we go ahead with the setting? If so, maybe should be done in another PR?

akano12:
> What do you think Bnyro, should we go ahead with the setting? If so, maybe should be done in another PR?

Pretty sure the reply above yours is satire (quality option for mini player...)

RafaelsRamos:
IndusAryan always looks serious to me 🥹

Bnyro (approved):
Works very well, thank you!

PR #5414, opened 2024-01-01
Add baseline profile generation

Isira-Seneviratne:
Add module for generating a [baseline profile](https://developer.android.com/topic/performance/baselineprofiles/overview) for the app.


IndusAryan:
wow, pr so advanced that i couldn't understand what the code is doing (ik making app fast somehow) and the pixel props included, maybe @RafaelsRamos can drop his 2 cents here with good explanation.

Isira-Seneviratne:
> wow, pr so advanced that i couldn't understand what the code is doing (ik making app fast somehow) and the pixel props included (for what?), maybe @RafaelsRamos can drop his 2 cents here with good explanation.

It allows a profile of methods called during startup to be generated, so that they can be compiled ahead of time to improve startup performance on Android 7.0 and later.

baselineProfile/build.gradle.kts
```diff
@@ -0,0 +1,52 @@
+import com.android.build.api.dsl.ManagedVirtualDevice
+
+plugins {
+    alias(libs.plugins.androidTest)
+    alias(libs.plugins.jetbrainsKotlinAndroid)
+    alias(libs.plugins.baselineprofile)
+}
+
+android {
+    namespace = "com.example.baselineprofile"
```
RafaelsRamos, line 10: Should we update the module namespace?

```suggestion
    namespace = "com.github.libretube.baselineprofile"
```

IndusAryan:
i understood now, but why pixel virtual device prop is needed for tests and why source compat of baseline is not jdk 17 (LTS) but 1_8 ,different from Libretube

RafaelsRamos:
The source compact can be `17`, it defaulted to `1_8` because Android Studio generated it this way (it is probably a good idea to change it to `17` for consistency).

Regarding the virtual device, the way the baseline profile generates the profiles, is by pre-compiling classes and methods from flows we want him to optimize (startup on this PR). This must be done on rooted devices (rooted emulators or physical devices).

The reason why we want it on an emulator is because it should be ran before a new release, so adding it to the release pipeline would be the ultimate goal.

Bnyro (approved):

PR #5415, opened 2024-01-01
fix: correctly display transparent channelImages

FineFindus:
Fixes an issue where transparent profile pictures were not readable in dark mode because YouTube uses a light background for the image by default. This change has the side effect of making unloaded profile pictures appear white.

| Before                                                                                                                                    	| After                                                                                                                               	|
|-------------------------------------------------------------------------------------------------------------------------------------------	|-------------------------------------------------------------------------------------------------------------------------------------	|
| ChannelImage with transparent background 	| ChannelImage with white background 	|


Bnyro:
I might have an other idea to prevent these white placeholders, I'll take a look after my vacation.

PR #5429, opened 2024-01-03
feat: cache feed response

FineFindus:
Adds a light cache to the initial feed response. This improves the user experience by eliminating the need to load the feed a second time when switching from Home to the Subscriptions tab, as well as eliminating unnecessary requests. The response is cached for 5 minutes, using pull-to-refresh ignores the caching and always requests a new version from the server.

### DIFF
--- a/app/src/main/java/com/github/libretube/api/SubscriptionHelper.kt
+++ b/app/src/main/java/com/github/libretube/api/SubscriptionHelper.kt
@@ -25,6 +25,7 @@ object SubscriptionHelper {
 
     private var feed: List<StreamItem> = listOf()
     private var lastFeedFetchTime: Long = 0
+
     /*
      * Minimum time that has to pass until the feed is refreshed.
      * Currently set to 5 minutes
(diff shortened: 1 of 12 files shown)


FireMasterK:
I don't like this feature to be implemented in LibreTube. I think we can add ETag based caching on the server-side to be more efficient with feed updates. The only problem is with unauthenticated feeds, these will use the `POST` method if there are over 100 subscriptions, and POST requests may not be cached in the client side.

Bnyro:
We already have an option called "Load feed in background" which prevents the feed from being reloaded each time, so I don't think adding yet another way for the same thing is a good idea, sorry.

FineFindus:
OK, that seems to do most of what this implementation does. What is the reason for disallowing pull-to-refresh when it is enabled? Would you be open to allowing it as a compromise between the current implementation and this pr?

Bnyro:
Fine by me :+1:

FineFindus:
I tested the "Load feed in background" setting, but I'm unsure about the "prevent it from being auto-refreshed". I initially assumed it would block the pull-to-refresh, since the feed is being refreshed when opening the app, but that's also not the case. What does is actually prevent?

Bnyro:
Reloading when switching tabs.

Issue #5439, opened 2024-01-05
Swipe control don't work as expect


SAOS200X:
### Steps to reproduce

1. Turn on or off swipe controls
2. Play any video and swipe
3. The volume swipe alway work despite the swipe controls turn on or off.

### Expected behavior

It should able to adjust the brightness and volume when turn on swipe controls as it described.

### Actual behavior

Only the volume adjust work and alway work despite swipe controls turn off.

### LibreTube version

0.21

### Android version

Android 10

### Other details

One UI version: 2.5

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

RafaelsRamos:
Hey there @SAOS200X 👋 

Congratz on your first ticket in the repo 💪 

After analysing the issue, these were the results:

https://github.com/libre-tube/LibreTube/assets/40279132/333a92d1-51b8-481b-81b4-2512fb414d85

https://github.com/libre-tube/LibreTube/assets/40279132/eb190bfe-12a7-4c91-b94a-c9fbf146370d

Please let me know if this is the flow you reported.

After reviewing the setting you mentioned, I see the description might be misleading. This setting only affects Video-player, on Fullscreen mode. For that reason, I will go ahead and update the description, to ensure a more accurate message.

SAOS200X:
oh, look like I was downloaded the wrong version, I download the arm64-v8a that make the app don't work as expect, while the universal version work fine, sorry about this.

SAOS200X:
oh, look like I was downloaded the wrong version, I download the arm64-v8a that make the app stuck in audio only mode, while the universal version work fine. But in both version, the volume swipe alway work in audio only mode despite swipe controls turn off. I re-add the issue here: https://github.com/libre-tube/LibreTube/issues/5445

PR #5440, opened 2024-01-05
fix: clarify setting description

RafaelsRamos:
Tweak `swipe_controls_summary` to describe what it does more accurately.

see #5439


Bnyro (approved):
Thanks!

PR #5441, opened 2024-01-05
chore: improve save feed summary

FineFindus:
Improves the summary of the "Load feed in background" option to help users understand what exactly changes when this option is turned on, see https://github.com/libre-tube/LibreTube/pull/5429.
I also think that enabling this option by default is worthwhile, as it both improves the user experience and reduces the load on the server, with very little to no downside.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -297,7 +297,7 @@
     <string name="preferences">Preferences</string>
     <string name="backup_customInstances">Custom Instances</string>
     <string name="save_feed">Load feed in background</string>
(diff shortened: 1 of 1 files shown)


Bnyro (approved):
Thanks!

I don't agree with enabling it by default since users likely expect the feed to be refreshed when they switch tabs. Also, feed responses are being cached in most cases.

Bnyro:
> I also think that enabling this option by default is worthwhile, as it both improves the user experience and reduces the load on the server, with very little to no downside.
>
Okay, changed my mind, I have changed it now to not refresh the feed every time when switching tabs :+1:

PR #5443, opened 2024-01-05
fix: Improve home feed layout and fix issue where incorrect cache size was defaulted.

RafaelsRamos:
### Issue

The main issue being addressed in this PR is the fact that by default the **Max image cache size** setting was empty, instead of "128MB" (that is defaulted when we go to Settings > Advanced > Max image cache size).

For a sample of the issue please take a look at this video (on Google Drive - too large) - [Issue replication video](https://drive.google.com/file/d/100EXLmjwwdKpMxl-F8tkpWWbm--05C3h/view?usp=drive_link)
In the video, a `200` represents a new image loaded, and a `304` or nothing, represents images being retrieved from cache.

To test this on your own, here's a patch with the necessary changes - [analyse-current-state.patch](https://github.com/libre-tube/LibreTube/files/13845315/Analyse_image_load_-_currently_.patch).

### What does this PR do?

- Changes the default value from "" to "128";
- Sets recycler view to fixed size, since its size will not change without page reload;
- Adjusts __fragment_home.xml__, so that the horizontal recycler view extends to the end of the screen;
- Added `crossfade` - Please let me know if you want this removed, I added it because it look very much an improvement in my eyes 👀 .

Here's the same analysis with the changes contained in this PR (on Google Drive - too large) - [post-changes sample](https://drive.google.com/file/d/1BT3HVkWN2H7vUQZW9vSO9pBNKMb6Y3l_/view?usp=sharing).

To test this on your own, here's a patch with the necessary changes - 
[Post-changes-sample.patch](https://github.com/libre-tube/LibreTube/files/13845335/Analyse_image_load_-_Post_changes_.patch).

Lastly, here are the current and pos changes samples for visual comparison.

https://github.com/libre-tube/LibreTube/assets/40279132/6b92d025-9eac-4357-8e45-164759657582

https://github.com/libre-tube/LibreTube/assets/40279132/563e58f6-17ee-4a0c-b136-9e4bdeb96c77

Please let me know if this does not make sense, or if I may have missed some business rule(s).

### DIFF
--- a/app/src/main/java/com/github/libretube/helpers/ImageHelper.kt
+++ b/app/src/main/java/com/github/libretube/helpers/ImageHelper.kt
@@ -19,35 +19,43 @@ import com.github.libretube.util.DataSaverMode
 import java.nio.file.Path
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.withContext
+import java.io.File
 
 object ImageHelper {
     lateinit var imageLoader: ImageLoader
+
+    private val Context.coilFile get() = cacheDir.resolve("coil")
 
     /**
      * Initialize the image loader
      */
     fun initializeImageLoader(context: Context) {
-        val maxImageCacheSize = PreferenceHelper.getString(
-            PreferenceKeys.MAX_IMAGE_CACHE,
-            ""
-        )
+        val maxCacheSize = PreferenceHelper.getString(PreferenceKeys.MAX_IMAGE_CACHE, "128")
 
         imageLoader = ImageLoader.Builder(context)
             .callFactory(CronetHelper.callFactory)
+            .crossfade(true)
             .apply {
-                when (maxImageCacheSize) {
-                    "" -> {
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/github/libretube/helpers/ImageHelper.kt
```diff
@@ -19,44 +19,53 @@ import com.github.libretube.util.DataSaverMode
...
 
         imageLoader = ImageLoader.Builder(context)
             .callFactory(CronetHelper.callFactory)
+            .crossfade(true)
             .apply {
-                when (maxImageCacheSize) {
-                    "" -> {
-                        diskCachePolicy(CachePolicy.DISABLED)
-                    }
+                if (!allowsCache) {
```
Bnyro, line 40: As we only use this condition once, I don't think it's reasonable to save `allowCache` in a variable.

Bnyro (approved):
Thanks!

PR #5475, opened 2024-01-11
Update baseline profile configuration

Isira-Seneviratne:
* Updated the generated baseline and startup profiles with the latest changes.
* Updated the baseline profile generation configuration.

### DIFF
--- a/baselineProfile/build.gradle.kts
+++ b/baselineProfile/build.gradle.kts
@@ -28,21 +28,21 @@ android {
 
     targetProjectPath = ":app"
 
-//    testOptions.managedDevices.devices {
-//        create<ManagedVirtualDevice>("pixel6Api34") {
-//            device = "Pixel 6"
-//            apiLevel = 34
-//            systemImageSource = "google"
-//        }
-//    }
+    testOptions.managedDevices.devices {
+        create<ManagedVirtualDevice>("pixel6Api34") {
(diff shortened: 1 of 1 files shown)


Bnyro:
> Updated the generated baseline and startup profiles with the latest changes.
>
We need to do this after each change now? Wouldn't it be better to just do it once before the release?

Isira-Seneviratne:
> > Updated the generated baseline and startup profiles with the latest changes.
> >
> We need to do this after each change now? Wouldn't it be better to just do it once before the release?

You're right. I'll just include the config change and remove the rest.

Bnyro (approved):

Bnyro:
@Isira-Seneviratne I've been trying to update the baseline profile, but I can't seem to get it work (tried with Android Studio and `./gradlew generateBaselineProfile`). What's the right command/way to generate these?

Isira-Seneviratne:
Yeah, there seems to be an issue when including the native libraries. To generate the profile, I commented out some lines in app/build.gradle (I've added a comment indicating as such).

Bnyro:
Hmm, it still gives me some errors when trying to generate the profile :/

Could you please create them and make a PR when you have time (no need to hurry, just when you're free :) )?

Isira-Seneviratne:
Sure.

PR #5478, opened 2024-01-11
feat: Support multiple filter selection

### LINKED ISSUE #5453: Adding multiple selection to the subscription feed filter

Currently in the subscription feed you can filter by selecting only one of the following options: All, Videos, Shorts & Livestreams. As I am only interested in videos and livestreams. I have to switch between those options and sometimes can be annoying. I don't use de "All" option because shorts are included and there are to many of them that it takes time to get to the videos and livestreams.

My suggestion is add a checkbox to each option in the filters, so the user can check the options that he wants to see.

### Other details

_No response_

### Acknowledgements

RafaelsRamos:
## What does this PR do?
1. Fixed typo on filter preference key;
2. Create a bottom sheet for filtering & sorting;
3. Ensures Home & Subscription screens filter their content accordingly;
4. closes #5453 

### Transition
Fixing the typo on the filter preference key results in resetting the users' current filter preference. This is intended since it greatly simplifies the logic to support multiple filters. If we decide that we want to transition to multiple filter support without a reset, that can be achieved by reverting the typo fix + adjusting the following property on _ContentFilter.kt_:
```kotlin
private val enabledFiltersSet get() = PreferenceHelper
    .getString(SELECTED_FEED_FILTERS, "1,2,3")
    // Line below supports a transition from the currently selected filter
    .let { filtersPref -> if (filtersPref == "0") "1,2,3" else filtersPref } 
    .split(',')
    .toMutableSet()
```
If you think it is worth it, we can go ahead and add support for a transition from the currently selected filter.

## Showcase
https://github.com/libre-tube/LibreTube/assets/40279132/124730e4-31cf-4bb4-a099-3ae88a655f2d

## Closing Notes
Thank you for reviewing! Please let me know if you disagree with the approaches I chose.


app/src/main/res/drawable/ic_filter_sort.xml
```diff
@@ -0,0 +1,5 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android" android:height="24dp" android:tint="#000000" android:viewportHeight="24" android:viewportWidth="24" android:width="24dp">
```
Bnyro, line 1:
```suggestion
<vector xmlns:android="http://schemas.android.com/apk/res/android" android:height="24dp" android:tint="?colorControlNormal" android:viewportHeight="24" android:viewportWidth="24" android:width="24dp">
```

Bnyro:
> Fixing the typo on the filter preference key results in resetting the users' current filter preference. This is intended since it greatly simplifies the logic to support multiple filters. If we decide that we want to transition to multiple filter support without a reset, that can be achieved by reverting the typo fix + adjusting the following property on ContentFilter.kt:
>
I think it's fine to change the preference key for the transition, I already did the same thing multiple times before when refactoring a preference.

app/src/main/java/com/github/libretube/ui/sheets/FilterSortBottomSheet.kt
```diff
@@ -0,0 +1,99 @@
...
+
+        fun createWith(
+            sortOptions: Collection<SelectableOption>,
+            sortListener: (index: Int, isSelected: Boolean) -> Unit,
+            filtersListener: () -> Unit
+        ) = FilterSortBottomSheet().apply {
+            this.sortOptions = sortOptions
+            this.sortFilter = sortListener
+            this.filterListener = filtersListener
+        }
```
Bnyro, line 95: We're using `FragmentResultListener`'s for communicating between fragments in other places of the app currently, because when passing functions they are not kept when the fragment is recreated (e.g. because Android kills/recreates the fragment for memory reasons or when the user switches the device from dark to light theme or vice versa). I haven't tested this on your PR yet, but in my opinion it would be preferable to use the `FragmentResultListener` API too here to avoid any issues.

RafaelsRamos: Got it! I'll take a look at that API and update the changes 👍

app/src/main/java/com/github/libretube/ui/fragments/SubscriptionsFragment.kt
```diff
@@ -258,15 +246,18 @@ class SubscriptionsFragment : DynamicLayoutManagerFragment() {
     }
 
     private fun List<StreamItem>.filterByStatusAndWatchPosition(): List<StreamItem> {
+
         val streamItems = this.filter {
             val isLive = (it.duration ?: -1L) < 0L
```
Bnyro, line 251: We know have a `StreamItem#isLive` property, so this line is redundant and can be replaced with just using `it.isLive`.

app/src/main/java/com/github/libretube/enums/ContentFilter.kt
```diff
@@ -0,0 +1,26 @@
+package com.github.libretube.enums
+
+import com.github.libretube.constants.PreferenceKeys.SELECTED_FEED_FILTERS
+import com.github.libretube.helpers.PreferenceHelper
+
+private val enabledFiltersSet get() = PreferenceHelper
+    .getString(SELECTED_FEED_FILTERS, "1,2,3")
+    .split(',')
+    .toMutableSet()
```
Bnyro, line 9:
```suggestion
private val defaultFilters = arrayOf(ContentFilter.VIDEOS, ContentFilter.SHORTS, ContentFilter.LIVESTREAMS)

private val enabledFiltersSet get() = PreferenceHelper
   .getString(SELECTED_FEED_FILTERS, defaultFilters.joinToString(","))
    .split(',')
    .toMutableSet()
```

Additionally, please also move these two values into the companion object of the enum class `ContentFilter`, just since we're doing so too in other places to keep the things that belong together at one class.

app/src/main/java/com/github/libretube/enums/ContentFilter.kt
```diff
@@ -0,0 +1,26 @@
...
+
+enum class ContentFilter(private val id: Int) {
```
Bnyro, line 11: We don't actually need this "id" parameter here.
Instead we can use `ContentFilter#ordinal` which returns the index of the enum property, so:
VIDEOS -> 0
SHORTS -> 1
LIVESTREAMS -> 2

Bnyro:
Looks pretty good and well overthought from the UI/UX perspective, only left some recommendations for the code implementation, which would make things a bit cleaner overall :+1: 

Thank you for looking into this!

M00NJ:
Awsome!
Since the player screen design ideas where well recieved I wanted to look at some other screens in the app and propose improvements. This was one of the ideas to make the subscription tab ui a bit cleaner. Cool that it's already implemented :)

RafaelsRamos:
> Awsome! Since the player screen design ideas where well recieved I wanted to look at some other screens in the app and propose improvements. This was one of the ideas to make the subscription tab ui a bit cleaner. Cool that it's already implemented :)

That's on @SameenAhnaf , he's the one that proposed that UI 💪

app/src/main/java/com/github/libretube/obj/SelectableOption.kt
```diff
@@ -0,0 +1,6 @@
+package com.github.libretube.obj
+
+data class SelectableOption(
+    val isSelected: Boolean,
+    val name: String
+)
```
Bnyro, line 6:
```suggestion

@Parcelize
data class SelectableOption(
    val isSelected: Boolean,
    val name: String
): Parcelable
```

app/src/main/java/com/github/libretube/ui/fragments/SubscriptionsFragment.kt
```diff
@@ -196,6 +170,27 @@ class SubscriptionsFragment : DynamicLayoutManagerFragment() {
...
+            ) { _, resultBundle ->
+                selectedSortOrder = resultBundle.getInt(SELECTED_SORT_OPTION_KEY)
+                showFeed()
+            }
+
+            FilterSortBottomSheet.createWith(
+                sortOptions = sortOptions.mapIndexed { index, option ->
+                    SelectableOption(isSelected = index == selectedSortOrder, name = option)
+                }
+            ).show(fragManager)
```
Bnyro, line 190:
```suggestion
               val sortOptions = sortOptions.mapIndexed { index, option ->
                    SelectableOption(isSelected = index == selectedSortOrder, name = option)
                }
            FilterSortBottomSheet.apply {
                arguments = bundleOf(IntentDate.selectableOptions, sortOptions)
            }.show(fragManager)
```

app/src/main/java/com/github/libretube/ui/sheets/FilterSortBottomSheet.kt
```diff
@@ -0,0 +1,108 @@
...
+
+class FilterSortBottomSheet: ExpandedBottomSheet() {
+
+    private var _binding: FilterSortSheetBinding? = null
+    private val binding get() = _binding!!
+
+    private lateinit var sortOptions: Collection<SelectableOption>
+
+    private var selectedIndex: Int = 0
+
```
Bnyro, line 22:
```suggestion
    override fun onCreate() {
        super.onCreate()
        
        sortOptions = arguments.parcelable(IntentData.selectableOptions)
    }
```

app/src/main/java/com/github/libretube/ui/sheets/FilterSortBottomSheet.kt
```diff
@@ -0,0 +1,108 @@
...
+        const val FILTER_SORT_REQUEST_KEY = "filter_sort_request_key"
+        const val SELECTED_SORT_OPTION_KEY = "selected_sort_option_key"
+
+        fun createWith(
+            sortOptions: Collection<SelectableOption>
+        ) = FilterSortBottomSheet().apply {
+            this.sortOptions = sortOptions
+        }
+
```
Bnyro, line 105:
```suggestion
       
```

app/src/main/res/drawable/ic_filter_sort.xml
```diff
@@ -0,0 +1,5 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android" android:height="24dp" android:tint="?colorControlNormal" android:viewportHeight="24" android:viewportWidth="24" android:width="24dp">
```
Bnyro, line 1: Please also format that file using Android studio's builtin formatter. It's just for convenience.

Bnyro:
The above suggestions (except the SVG thing) are only a small refactor to avoid crashes when the bottom sheet is being recreated, since the `var sortOptions = ...` value won't be kept either when the sheet is killed. Hence we must pass it as an intent data argument in order to not become killed by Android lifecycle.

RafaelsRamos:
Hey Bnyro, thank you for catching this issue. It would lead to some unnecessary crashes...

Updated the icon's formatting + passed the parameters through arguments.

One point to note. I was not able to use the recommended `arguments.parcelable(...)` due to the fact that the argument being passed is a collection of items. For that reason, I used the `arguments.getParcelableArray`, I hope that's OK.

Bnyro:
> One point to note. I was not able to use the recommended `arguments.parcelable(...)` due to the fact that the argument being passed is a collection of items. For that reason, I used the `arguments.getParcelableArray`, I hope that's OK.
>
Sure :+1:

I didn't test the code at all before commenting, it was just more of a general hint how I'd imagine the changes to be :)

Bnyro (approved):
Lgtm now, thank you!

Would you mind rebasing squashing your commits into two or three so that we can make the history of commits cleaner. Or should I just squash the whole PR into one commit when merging?

RafaelsRamos:
Yeah go ahead! I just like committing multiple to keep track of what was done (it helps me with the PR description).

PR #5491, opened 2024-01-15
feat: Improve new user experience + adjust home load

### LINKED ISSUE #5479: Facilitate changing instance

When new users join, it is unclear what instances will work for them. I suggest we make clear to them that the videos might not be loaded due to the instance and make it easy for them to change it (e.g. through a Snackbar).

### Other details

_No response_

### Acknowledgements

RafaelsRamos:
## Feature - Improve UX for new users
Some users have the bad luck of not picking a working instance on the first try. This development aims to offer them support and guide them into selecting a working instance.

## Types of instances
For new users, instances can be divided into three types.
* **Type A** - An instance that we don't get a response from. Using such an instance results in having an infinite loading;
* **Type B** - An instance that immediately results in the "Nothing here" screen;
* **Type C** - A functional instance;

## How can we improve the UX to facilitate the selection of an appropriate instance?
Depending on the type of instance the user has selected, the way the user will be guided will be different.
For users with an instance of:
* **Type A** - After 10 seconds of trying to load content, the user will be prompted with a SnackBar, with an Action that takes him/her to the IntentSettings screen (video A) - This does not cancel the thread.;
* **Type B** - On the "Nothing here" screen, two buttons were added:
  * "Retry" - The first time the user joins the App, he/she does not know that below exists a swipeable view;
  * "Change instance" - Button to facilitate the user journey into a functional instance;
* **Type C** - Nothing! The App is good to go! 💪 

## Adjacent issues
While working on this feature, there were a couple of behaviours that could be improved:
* When on an instance of **Type A**, swapping between tabs was causing multiple threads to be created and never cancelled - [check this video for a showcase](https://drive.google.com/file/d/1sQcdCGsH-vUqAL1R4rRTiYAppTEq1kEd/view?usp=drive_link);
* When loading the home screen and before finishing, swapping to another tab was causing the load not to be used (due to `binding` being null when it was time to fill the views) - sample below ("current_ui_behaviour.mp4");
* Switching between tabs leads the home to reload every time - sample below ("before_tab_swap.mp4");

https://github.com/libre-tube/LibreTube/assets/40279132/919cd439-220d-4312-af6c-7747c3667682

https://github.com/libre-tube/LibreTube/assets/40279132/c0881f71-7a35-4b9d-bdc9-fe0538772876

## What does this PR do?
- Create `HomeViewModel` responsible for the calls that fill the screen and is responsible for holding the LiveData instances that have the data being shown - Subscribing (on _onResume_) and Unsubscribing (on _onPause_) to these LiveData instances leads to a healthy UI/IO management, and supports data being loaded, even if we are not in the home screen;
- Create two mechanisms that facilitate changing the instance;
- closes #5479;

## Notes
- Clarify the rationale behind the chosen approach.

## Showcase

https://github.com/libre-tube/LibreTube/assets/40279132/88fc1f05-0bc4-4e3e-a119-75447943fd20

Showcase of the current approach cancelling previous, ongoing threads when creating new ones - [video sample](https://drive.google.com/file/d/1VYCb1H8iEPZ5X2uxJfU2WK_0NdlXW896/view?usp=drive_link).

https://github.com/libre-tube/LibreTube/assets/40279132/90acb8ba-1a74-4ca7-96bf-d5d8dd018bb4

https://github.com/libre-tube/LibreTube/assets/40279132/bf47d258-d000-4832-909a-67183cb4cfe5

## Test cases

This was the main test case I used to test this development. Please let me know if there are other "must test" flows.

1. Fully wipe data;
2. Select instance that fails immediately;
3. Press Retry button a couple of times;
3a. Ensure the progress bar briefly shows up, indicating ongoing attempt;
4. Press change instance button;
4a. Ensure redirection to Settings > Instance;
5. Select a new instance to an instance that loads indefinetely;
6. Go back all the way to the Home;
6a. Ensure the progress bar is enabled, indicating an ongoing attempt;
7. Wait a couple of seconds and switch to another tab;
8. Wait around 20 seconds;
8a. Ensure the SnackBar DOES NOT shows up;
9. Go back to Home tab, and wait around 20 seconds;
10. Wait 10 seconds;
10a. Ensure the SnackBar DOES show up;
11. Press the "Change" on the SnacBar;
12. Change to a working instance;
13. Go back to Home;
13a. Ensure the Home screen is correctly loaded;

## Closing Notes
I sincerely appreciate the review; I hope this improves both new users' UX and the home load.
Please let me know if any of the points or approaches mentioned do not make sense or if we should opt for another approach.

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/activities/SettingsActivity.kt
+++ b/app/src/main/java/com/github/libretube/ui/activities/SettingsActivity.kt
@@ -2,11 +2,13 @@ package com.github.libretube.ui.activities
 
 import android.os.Bundle
 import androidx.activity.addCallback
+import androidx.fragment.app.Fragment
 import androidx.fragment.app.commit
 import androidx.fragment.app.replace
 import com.github.libretube.R
 import com.github.libretube.databinding.ActivitySettingsBinding
 import com.github.libretube.ui.base.BaseActivity
+import com.github.libretube.ui.preferences.InstanceSettings
 import com.github.libretube.ui.preferences.MainSettings
 
 class SettingsActivity : BaseActivity() {
@@ -24,9 +26,7 @@ class SettingsActivity : BaseActivity() {
         }
 
         if (savedInstanceState == null) {
-            supportFragmentManager.commit {
-                replace<MainSettings>(R.id.settings)
-            }
+            redirectTo<MainSettings>()
         }
 
         // new way of dealing with back presses instead of onBackPressed()
@@ -34,15 +34,35 @@ class SettingsActivity : BaseActivity() {
             if (supportFragmentManager.findFragmentById(R.id.settings) is MainSettings) {
                 finishAndRemoveTask()
             } else {
-                supportFragmentManager.commit {
-                    replace<MainSettings>(R.id.settings)
-                }
+                redirectTo<MainSettings>()
                 changeTopBarText(getString(R.string.settings))
             }
+        }
+
+        handleRedirect()
+    }
+
+    private fun handleRedirect() {
+        val redirectKey = intent.extras?.getString(REDIRECT_KEY)
+
+        when (redirectKey) {
+            REDIRECT_TO_INTENT_SETTINGS -> redirectTo<InstanceSettings>()
+            else                        -> { }
         }
     }
 
     fun changeTopBarText(text: String) {
         if (this::binding.isInitialized) binding.toolbar.title = text
+    }
+
+    private inline fun <reified T : Fragment> redirectTo() {
+        supportFragmentManager.commit {
+            replace<T>(R.id.settings)
+        }
+    }
+
+    companion object {
+        const val REDIRECT_KEY = "redirect"
+        const val REDIRECT_TO_INTENT_SETTINGS = "intent_settings"
     }
 }
--- a/app/src/main/java/com/github/libretube/ui/activities/SettingsActivity.kt
+++ b/app/src/main/java/com/github/libretube/ui/activities/SettingsActivity.kt
@@ -45,10 +45,7 @@ class SettingsActivity : BaseActivity() {
     private fun handleRedirect() {
         val redirectKey = intent.extras?.getString(REDIRECT_KEY)
 
-        when (redirectKey) {
-            REDIRECT_TO_INTENT_SETTINGS -> redirectTo<InstanceSettings>()
-            else                        -> { }
-        }
+        if (redirectKey == REDIRECT_TO_INTENT_SETTINGS) redirectTo<InstanceSettings>()
     }
 
     fun changeTopBarText(text: String) {
--- a/app/src/main/java/com/github/libretube/ui/fragments/HomeFragment.kt
+++ b/app/src/main/java/com/github/libretube/ui/fragments/HomeFragment.kt
@@ -23,6 +23,7 @@ import com.github.libretube.constants.PreferenceKeys
 import com.github.libretube.databinding.FragmentHomeBinding
 import com.github.libretube.db.DatabaseHelper
 import com.github.libretube.db.DatabaseHolder
+import com.github.libretube.enums.ContentFilter
 import com.github.libretube.helpers.LocaleHelper
 import com.github.libretube.helpers.PlayerHelper
 import com.github.libretube.helpers.PreferenceHelper
@@ -145,12 +146,13 @@ class HomeFragment : Fragment() {
                 }
             }.getOrNull()?.takeIf { it.isNotEmpty() } ?: return
         }
+
+        val allowShorts = ContentFilter.SHORTS.isEnabled()
+        val allowVideos = ContentFilter.VIDEOS.isEnabled()
+        val allowAll = (!allowShorts && !allowVideos)
+
         var filteredFeed = feed.filter {
-            when (PreferenceHelper.getInt(PreferenceKeys.SELECTED_FEED_FILTER, 0)) {
-                1 -> !it.isShort
-                2 -> it.isShort
-                else -> true
-            }
+            (allowShorts && it.isShort) || (allowVideos && !it.isShort) || allowAll
         }
         if (PreferenceHelper.getBoolean(PreferenceKeys.HIDE_WATCHED_FROM_FEED, false)) {
             filteredFeed = runBlocking { DatabaseHelper.filterUnwatched(filteredFeed) }
--- a/app/src/main/java/com/github/libretube/ui/fragments/HomeFragment.kt
+++ b/app/src/main/java/com/github/libretube/ui/fragments/HomeFragment.kt
@@ -1,5 +1,6 @@
 package com.github.libretube.ui.fragments
 
+import android.content.Intent
 import android.os.Bundle
 import android.view.LayoutInflater
 import android.view.View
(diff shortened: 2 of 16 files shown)


RafaelsRamos:
If you agree with the data binding approach, I would like to mention that for the future it would be very beneficial to have a repository layer.

At the moment view models make their calls directly on the Retrofit instances and the state is being saved either on LiveData at the view model level - e.g. _SubscriptionsViewModel_ (this is already a great step, however does not support exchange of information between screens), or directly in fragments (usually the data is re-fetched when we re-enter the screen).

Here's a very simplified example of what it would look like:

```kotlin
class HomeFragment: Fragment() {
    private val viewModel: HomeViewModel by activityViewModels()

    override fun onResume() {
        super.onResume()
        viewModel.feedContent.observe(viewLifecycleOwner, ::showFeed)
    }

    override fun onPause() {
        super.onPause()
        viewModel.feedContent.removeObserver(::showFeed)
    }
    ...
}

class HomeViewModel(
    private val contentRepository: IContentRepository
): ViewModel() {
    val feedContent get() = contentRepository.feedContent

    init {
        viewModelScope.launch {
            contentRepository.loadFeed()
        }
    }
}

class ContentRepository: IContentRepository {
    override val feedContent: LiveData<List<StreamItem>?> get() = _feedContent
    private val _feedContent: MutableLiveData<List<StreamItem>> = MutableListData(null)

    override fun loadFeed() {
        if (feedContent.value == null) {
            // Load from API and fill live data
        }
    }
}
```

With this approach, all data fetched in one screen would be available in other screens that need the same information. All this management would be done inside the repositories

app/src/main/java/com/github/libretube/ui/models/HomeViewModel.kt
```diff
@@ -0,0 +1,186 @@
...
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.withContext
+
+class HomeViewModel: ViewModel() {
+
+    private val useSavedFeed get() = PreferenceHelper.getBoolean(SAVE_FEED, false)
+    private val hideWatched get() = PreferenceHelper.getBoolean(HIDE_WATCHED_FROM_FEED, false)
+
+    val trending: LiveData<List<StreamItem>?> get() = _trending
```
Bnyro, line 37: Why aren't we directly accessing the `MutableLiveData` from below?

app/src/main/java/com/github/libretube/ui/models/HomeViewModel.kt
```diff
@@ -0,0 +1,186 @@
...
+        val allowVideos = ContentFilter.VIDEOS.isEnabled()
+        val allowAll = (!allowShorts && !allowVideos)
+
+        val filteredFeed = this.filter {
+            allowAll || (allowShorts && it.isShort) || (allowVideos && !it.isShort)
+        }
+        return runBlocking { DatabaseHelper.filterUnwatched(filteredFeed) }
+    }
+
+    private suspend fun <T> runSafely(
```
Bnyro, line 157: Please move that into the `extensions` package.

app/src/main/java/com/github/libretube/ui/fragments/HomeFragment.kt
```diff
@@ -91,118 +127,58 @@ class HomeFragment : Fragment() {
...
-    private suspend fun loadTrending() {
-        val region = LocaleHelper.getTrendingRegion(requireContext())
-        val trending = runCatching {
-            withContext(Dispatchers.IO) {
-                RetrofitInstance.api.getTrending(region).deArrow().take(10)
-            }
-        }.getOrNull()?.takeIf { it.isNotEmpty() } ?: return
-        val binding = _binding ?: return
+    private fun showTrending(streamItems: List<StreamItem>?) {
+        streamItems ?: return
```
Bnyro, line 143:
```suggestion
    if (streamItems == null) return
```
is more readable.

app/src/main/java/com/github/libretube/ui/fragments/HomeFragment.kt
```diff
@@ -91,118 +127,58 @@ class HomeFragment : Fragment() {
...
-
-        var filteredFeed = feed.filter {
-            (allowShorts && it.isShort) || (allowVideos && !it.isShort) || allowAll
-        }
-        if (PreferenceHelper.getBoolean(PreferenceKeys.HIDE_WATCHED_FROM_FEED, false)) {
-            filteredFeed = runBlocking { DatabaseHelper.filterUnwatched(filteredFeed) }
-        }
-        val binding = _binding ?: return
+    private fun showFeed(streamItems: List<StreamItem>?) {
+        streamItems ?: return
```
Bnyro, line 154:
```suggestion
    if (streamItems == null) return
```
is more readable.

app/src/main/java/com/github/libretube/ui/fragments/HomeFragment.kt
```diff
@@ -91,118 +127,58 @@ class HomeFragment : Fragment() {
...
 
-    private suspend fun loadPlaylists() {
-        val playlists = runCatching {
-            withContext(Dispatchers.IO) {
-                PlaylistsHelper.getPlaylists().take(20)
-            }
-        }.getOrNull()?.takeIf { it.isNotEmpty() } ?: return
-        val binding = _binding ?: return
+    private fun showPlaylists(playlists: List<Playlists>?) {
+        playlists ?: return
```
Bnyro, line 175:
```suggestion
    if (playlists == null) return
```
is more readable.

app/src/main/java/com/github/libretube/ui/fragments/HomeFragment.kt
```diff
@@ -216,46 +192,72 @@ class HomeFragment : Fragment() {
...
-        val videos = withContext(Dispatchers.IO) {
-            DatabaseHolder.Database.watchHistoryDao().getAll()
-        }
-        val unwatchedVideos = DatabaseHelper.filterUnwatched(videos.map { it.toStreamItem() })
-            .reversed()
-            .take(20)
-        if (unwatchedVideos.isEmpty()) return
-        val binding = _binding ?: return
+    private fun showContinueWatching(unwatchedVideos: List<StreamItem>?) {
+        unwatchedVideos ?: return
```
Bnyro, line 196:
```suggestion
    if (unwatchedVideos == null) return
```
is more readable.

app/src/main/java/com/github/libretube/ui/activities/SettingsActivity.kt
```diff
@@ -24,25 +26,43 @@ class SettingsActivity : BaseActivity() {
...
+
+        handleRedirect()
+    }
+
+    private fun handleRedirect() {
+        val redirectKey = intent.extras?.getString(REDIRECT_KEY)
+
+        when (redirectKey) {
+            REDIRECT_TO_INTENT_SETTINGS -> redirectTo<InstanceSettings>()
+            else                        -> { }
```
Bnyro, line 50:
```suggestion
            else                        -> Unit
```

app/src/main/java/com/github/libretube/ui/activities/SettingsActivity.kt
```diff
@@ -24,25 +26,43 @@ class SettingsActivity : BaseActivity() {
...
+        }
```
Bnyro, line 51:
```suggestion
        if (redirectKey == REDIRECT_TO_INTENT_SETTINGS) redirectTo<InstanceSettings>()
```

app/src/main/java/com/github/libretube/ui/models/HomeViewModel.kt
RafaelsRamos: I created the extra properties to prevent the users of `HomeViewModel` from posting data on the object. This way, we ensure only the view model is responsible for posting new values.

app/src/main/java/com/github/libretube/ui/models/HomeViewModel.kt
RafaelsRamos: Applied 💪

RafaelsRamos:
Thanks a lot for the review Bnyro 💪  I hope I have not forgotten any topic.

app/src/main/java/com/github/libretube/ui/models/HomeViewModel.kt
Bnyro: We're not a library, so nobody should ever get the idea to post data from anywhere else than the view model (except if it's really necessary), I think we should stick with using `MutableLiveData` only everywhere for consistency across different view models.

app/src/main/java/com/github/libretube/ui/models/HomeViewModel.kt
```diff
@@ -0,0 +1,170 @@
...
+
+        val filteredFeed = this.filter {
+            allowAll || (allowShorts && it.isShort) || (allowVideos && !it.isShort)
+        }
+        return runBlocking { DatabaseHelper.filterUnwatched(filteredFeed) }
+    }
+
+    private fun <T> MutableLiveData<T>.updateIfChanged(newValue: T) {
+        if (value != newValue) value = newValue
+    }
```
Bnyro, line 160: Please move that to that extensions package too, we certainly can re-use this function in other places of the app where view models are being used too :+1:

Bnyro:
> If you agree with the data binding approach, I would like to mention that for the future it would be very beneficial to have a repository layer.
> At the moment view models make their calls directly on the Retrofit instances and the state is being saved either on LiveData at the view model level - e.g. SubscriptionsViewModel (this is already a great step, however does not support exchange of information between screens), or directly in fragments (usually the data is re-fetched when we re-enter the screen).
>
Personally, I don't think we need that additional level of abstraction.

I agree that refactoring parts of the app to make more use of view models to persist data is the right direction, but these additional layers would just cause a lot of additional code to maintain. It might be a common practice in the industry (I can't tell, I don't work there), but I think for us it's much better to keep things as short and simple as possible to keep maintenance as easy as possible.

Hence we should only use such repository approaches if really needed, and I don't currently see a place in the app where it'd be really useful.

app/src/main/res/layout/fragment_home.xml
```diff
@@ -126,6 +127,26 @@
...
+
+        <com.google.android.material.button.MaterialButton
+            android:id="@+id/refresh_button"
+            android:layout_width="150dp"
+            android:layout_height="50dp"
+            android:layout_gravity="center"
+            android:layout_marginTop="30dp"
+            android:textSize="12sp"
+            app:backgroundTint="@color/exo_black_opacity_70"
+            android:text="@string/retry"/>
```
Bnyro, line 139: Please remove these lines, the buttons look quite strange like that.

Either give them a style of `@style/Widget.Material3.Button.TextButton` or use `style="@style/Widget.Material3.Button.ElevatedButton` (I personally prefer elevated buttons for that purpose)

RafaelsRamos:
Great 🙌  I hope I addressed all points!

Regarding the repositories, I agree that at the moment it might not make sense since it is a shift from what we currently use. Thus, it would also lead to some adjustment time 👍 

That being said, I would just like to point out a benefit of using repositories I forgot the first time.

On flows such as these:
1. User is on HomePage;
2. User presses on a Video and subscribes;
3. Goes to Miniplayer;
4. Home page is not refreshed with this subscription;

having a central point where information is fetched and observed from, ensures the UI is currently reflecting the most recent updates. Additionally, if the data (subscriptions in this case) is updated in a centralized place, there is no need to re-fetch information again when entering or re-entering screens that need it - resulting also in less busy instances (not sure about this part, please current me if I'm wrong).

That being said, I agree that it may not make sense such changes 👍

app/src/main/java/com/github/libretube/ui/models/HomeViewModel.kt
RafaelsRamos: Yeah, I understand your point, removed the `LiveData` properties.

Bnyro (approved):
Thank you!

PR #5515, opened 2024-01-19
feat(ui): save scroll position in trends

IndusAryan:
previously when we swtiched to other tab and came back to trends it reloaded, after switching to viewmodel it is good,
 but when we scroll and switch to other tab and come back to trends, it resets to the top position.
this enhances that.

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/fragments/TrendsFragment.kt
+++ b/app/src/main/java/com/github/libretube/ui/fragments/TrendsFragment.kt
@@ -7,6 +7,8 @@ import android.view.View
 import android.view.ViewGroup
 import androidx.core.view.isGone
 import androidx.fragment.app.activityViewModels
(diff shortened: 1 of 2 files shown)


IndusAryan:
bnyro what is the reason to not decapitate reviewdog ?

Bnyro:
> bnyro what is the reason to not decapitate reviewdog ?
>
Lazyness

IndusAryan:
> I personally don't like this behavior change and would prefer to have discussed this first.

i thought my viewmodel f*ked it up, but why ? all the majority apps whether foss or not, save scroll position when switching tabs or views. anyways discuss with the deciders team, i mean collaborator team.

Bnyro (approved):
There were no voices against it, so it's fine to accept that behavior I guess.

Thanks!

PR #5525, opened 2024-01-21
feat: support for downloading whole playlist at once

### LINKED ISSUE #3080: Download playlist and list whole playlist under download

Hi!

I also use libretube for bluetooth to listen to music. 

At the moment you can only download a video/audio one at a time. It would be great if you could click (hold) on a playlist and get the option to download the entire playlist. Once downloaded, it would be good to list the playlist in the download section and not each video at once. 

Via bluetooth, there is no skip function for the video/audio file at the moment. At least I can't skip to the next clip. I always have to go back and start the next one manually.

Thanks! It's a really great app! 

### Other details

_No response_

### Acknowledgements

Bnyro:
closes #3080

disclaimer: only tested for very small playlists so far

@RafaelsRamos if you have time, I'd appreciate a quick review!


app/src/main/java/com/github/libretube/ui/dialogs/DownloadPlaylistDialog.kt
```diff
@@ -0,0 +1,87 @@
...
+                }
+                val captionLanguage = binding.subtitleSpinner.selectedItemPosition.let { captionPosition ->
+                    if (captionPosition >= 1) availableLanguages[captionPosition - 1].code else null
+                }
+                val audioLanguage = binding.audioLanguageSpinner.selectedItemPosition.let { audioLanguage ->
+                    if (audioLanguage >= 1) availableLanguages[audioLanguage - 1].code else null
+                }
+
+                if (maxVideoQuality == null && maxAudioQuality == null) {
+                    return@setPositiveButton
```
RafaelsRamos, line 69: In this instance, we leave the Dialog without informing the user about what happened. What do you think about a toast informing him that the download didn't start? 🤔

app/src/main/java/com/github/libretube/ui/dialogs/DownloadPlaylistDialog.kt
```diff
@@ -0,0 +1,87 @@
...
+            listOf(getString(R.string.no_subtitle)) + availableLanguages.map { it.name }
+        binding.audioLanguageSpinner.items =
+            listOf(getString(R.string.default_language)) + availableLanguages.map { it.name }
+
+        return MaterialAlertDialogBuilder(requireContext())
+            .setTitle(getString(R.string.download_playlist) + ": " + playlistName)
+            .setMessage(R.string.download_playlist_note)
+            .setView(binding.root)
+            .setPositiveButton(R.string.download) { _, _ ->
+                val maxVideoQuality = binding.videoSpinner.selectedItemPosition.let { videoPosition ->
```
RafaelsRamos, line 53: Nitpicking (and debatable 😄): We might be able to make use of the `with` operator and List'`getOrNull` to possibly simplify this block:
```kotlin
    //...
    with (binding) {
        val videoQualityIndex = videoSpinner.selectedItemPosition
        //...

        if (videoQualityIndex == 0 && audioQualityIndex == 0) {
            return@setPositiveButton
        }

        val videoQuality = possibleVideoQualities.getOrNull(videoQualityIndex - 1)?.getWhileDigit()
        //...

        val downloadEnqueueIntent = Intent(requireContext(), PlaylistDownloadEnqueueService::class.java)
            .putExtra(IntentData.maxVideoQuality, videoQuality)
        //...
    }
```
Please, feel free to resolve.

RafaelsRamos (commented):
> @RafaelsRamos if you have time, I'd appreciate a quick review!

Sure, it is always a pleasure to help!

First of all, congratulations, great work 💪 🚀 

That being said, with the introduction of this incredible feature, it may be worth revisiting the Downloads screen.

There are a few points I think we should address in future tickets:
1. Adding a multiple downloads removal feature, since at the moment, it takes too much time to remove large amounts of downloads one by one;
2. The downloads recycler view has flaky behaviour when removing items from it (video below - 1st video);
3. When selecting to watch a download that has yet to start being downloaded (It frequently happens when we're trying to download many videos), we get into a state where the video will never play (sample below - 2nd video). I don't think this needs to be addressed ASAP, but I think at some point, we could prevent the user from advancing + show a toast with some message 🤔 ;
4. Lastly, with this new feature, we need to reload the downloads page a few times until we can see all the items that will be downloaded. This issue can be mitigated by adding that "Repository" layer we discussed the other day. With this layer, the DownloadsFragment can observe changes on a LiveData object that resides in the repository and is updated when new entries are added to the local database (also, this issue is not crucial, but it might be worth looking at in the future);

Of course, I would be very happy to address some of the issues pointed out above if you agree 👍 

https://github.com/libre-tube/LibreTube/assets/40279132/0ce5b2a6-4e28-4ce0-9327-53ac1c912bfe

https://github.com/libre-tube/LibreTube/assets/40279132/71aa8998-48d1-4735-9c2a-dc036f23c7f1

app/src/main/java/com/github/libretube/services/PlaylistDownloadEnqueueService.kt
```diff
@@ -0,0 +1,194 @@
...
+
+    override fun onDestroy() {
+        ServiceCompat.stopForeground(this, ServiceCompat.STOP_FOREGROUND_REMOVE)
+        stopSelf()
+
+        super.onDestroy()
+    }
+
+    companion object {
+        private const val ENQUEUE_PROGRESS_NOTIFICATION_ID = 3
```
FireMasterK, line 192: Rather than having this here as a magic number, would it make sense to create an enum somewhere else along with other notification IDs? It's difficult to understand why 3 is used otherwise. (I'm guessing 1 and 2 are used elsewhere in the app)

This could be another PR perhaps.

app/src/main/java/com/github/libretube/services/PlaylistDownloadEnqueueService.kt
```diff
@@ -0,0 +1,194 @@
...
+
+        // sort streams by their quality/bitrate
+        val sortedStreams = streams
+            .sortedBy { it.quality.getWhileDigit() }
+
+        // return the last item below the maximum quality - or if there's none - the stream with
+        // the lowest quality available
+        return sortedStreams
+            .lastOrNull { it.quality.getWhileDigit()!! <= maxStreamQuality }
+            ?: sortedStreams.firstOrNull()
```
FireMasterK, line 181: Since this is not a number, won't this have issues when sorting?
Eg: 1080p vs 720p.
I'm not sure if Kotlin converts it to a number.

Another thing to note is quality can have FPS, which can also cause issues.

FireMasterK (approved):
I tried reviewing with my limited Android knowledge :sweat_smile:

app/src/main/java/com/github/libretube/services/PlaylistDownloadEnqueueService.kt
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    override fun onDestroy() {
+        ServiceCompat.stopForeground(this, ServiceCompat.STOP_FOREGROUND_REMOVE)
+        stopSelf()
+
+        super.onDestroy()
```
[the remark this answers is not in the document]
RafaelsRamos: The same applies to the other places we call `onDestroy()` directly. I suggest we always go for `stopSelf()`, and let the system call the `onDestroy`

app/src/main/java/com/github/libretube/services/PlaylistDownloadEnqueueService.kt
Bnyro: The `getWhileDigit` function reads the string as long as the chars are digit, so "1080p" -> 1080 for example. That works fine, even if FPS are provided too.

app/src/main/java/com/github/libretube/services/PlaylistDownloadEnqueueService.kt
Bnyro: Yes, we should probably do thay in an other PR. 1 and 2 are used for the player and the normal download notification.

app/src/main/java/com/github/libretube/ui/dialogs/DownloadPlaylistDialog.kt
Bnyro: Yes, I thought about this too, but I ended up thinking that it's easier to read when we use an if/else block, because the `get(index-1)` is a bit like magic and it's probably not that easy to understand when looking at it again at some time.

Using `with` sounds useful though.

app/src/main/java/com/github/libretube/ui/dialogs/DownloadPlaylistDialog.kt
Bnyro: :+1:

app/src/main/java/com/github/libretube/ui/dialogs/DownloadPlaylistDialog.kt
```diff
@@ -0,0 +1,87 @@
...
+
+        val possibleVideoQualities = resources.getStringArray(R.array.defres).toList().let {
+            // remove the automatic quality entry
+            it.subList(1, it.size)
+        }
+        val possibleAudioQualities = resources.getStringArray(R.array.audioQualityBitrates)
+        val availableLanguages = LocaleHelper.getAvailableLocales()
+
+        binding.videoSpinner.items = listOf(getString(R.string.no_video)) + possibleVideoQualities
+        binding.audioSpinner.items = listOf(getString(R.string.no_audio)) + possibleAudioQualities
```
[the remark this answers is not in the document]
Bnyro: Yes, **in the future**. We already do this for the normal download dialog, but I wanted to keep this PR simple and basic for now.

app/src/main/java/com/github/libretube/services/PlaylistDownloadEnqueueService.kt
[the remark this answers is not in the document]
Bnyro: Oh right, I once had this crash too, but it didn't occur anymore when I tried again, so I didn't know what's happened there. That makes sense, thanks for investigating :+1:

app/src/main/java/com/github/libretube/ui/dialogs/DownloadPlaylistDialog.kt
RafaelsRamos: Yep, I agree that the `getOrNull(index-1)` complexifies the block

Bnyro:
>     1. Adding a multiple downloads removal feature, since at the moment, it takes too much time to remove large amounts of downloads one by one;
>
I had the same thought and opened https://github.com/libre-tube/LibreTube/issues/5526 yesterday. However, being able to select which ones to delete would be nice too. It's up to what you think works best and is doable to implement :+1:
>     2. The downloads recycler view has flaky behaviour when removing items from it (video below - 1st video);
> 
Yes, I noticed that too, but I guess it's rather low priority for now.
>     3. When selecting to watch a download that has yet to start being downloaded (It frequently happens when we're trying to download many videos), we get into a state where the video will never play (sample below - 2nd video). I don't think this needs to be addressed ASAP, but I think at some point, we could prevent the user from advancing + show a toast with some message 🤔 ;
> 
Yes, that's certainly a good idea (but as you said, not that important for now, but still reasonable).
>     4. Lastly, with this new feature, we need to reload the downloads page a few times until we can see all the items that will be downloaded. This issue can be mitigated by adding that "Repository" layer we discussed the other day. With this layer, the DownloadsFragment can observe changes on a LiveData object that resides in the repository and is updated when new entries are added to the local database (also, this issue is not crucial, but it might be worth looking at in the future);
> 
I agree with you on that, in this specific case it would make sense to use a repository and live data to make everything more responsive. Feel free to look into that if you want to, I'd appreciate it!

PR #5528, opened 2024-01-22
refactor: Rewrite search functionality using Paging

Isira-Seneviratne:
Improve the search functionality using the [Paging library](https://developer.android.com/topic/libraries/architecture/paging/v3-overview).

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/adapters/SearchChannelAdapter.kt
+++ b/app/src/main/java/com/github/libretube/ui/adapters/SearchChannelAdapter.kt
@@ -70,6 +70,7 @@ class SearchChannelAdapter : ListAdapter<ContentItem, SearchViewHolder>(SearchCa
 
     private fun bindVideo(item: ContentItem, binding: VideoRowBinding, position: Int) {
         binding.apply {
+            thumbnail.setImageDrawable(null)
             ImageHelper.loadImage(item.thumbnail, thumbnail)
             thumbnailDuration.setFormattedDuration(item.duration, item.isShort)
             videoTitle.text = item.title
--- a/app/src/main/java/com/github/libretube/ui/adapters/SearchChannelAdapter.kt
+++ b/app/src/main/java/com/github/libretube/ui/adapters/SearchChannelAdapter.kt
@@ -1,0 +1,182 @@
+package com.github.libretube.ui.adapters
+
+import android.view.LayoutInflater
+import android.view.ViewGroup
+import androidx.core.os.bundleOf
+import androidx.core.view.isGone
+import androidx.recyclerview.widget.ListAdapter
+import com.github.libretube.R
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/github/libretube/ui/models/sources/SearchPagingSource.kt
```diff
@@ -0,0 +1,28 @@
...
+import com.github.libretube.util.deArrow
+import retrofit2.HttpException
+
+class SearchPagingSource(
+    private val searchQuery: String,
+    private val searchFilter: String
+): PagingSource<String, ContentItem>() {
+    override fun getRefreshKey(state: PagingState<String, ContentItem>): String? {
+        return null
+    }
```
Bnyro, line 16:
```suggestion
    override fun getRefreshKey(state: PagingState<String, ContentItem>): String? = null
```

app/src/main/java/com/github/libretube/ui/adapters/callbacks/SearchCallback.kt
```diff
@@ -0,0 +1,14 @@
...
+import com.github.libretube.api.obj.ContentItem
+
+object SearchCallback : DiffUtil.ItemCallback<ContentItem>() {
+    override fun areItemsTheSame(oldItem: ContentItem, newItem: ContentItem): Boolean {
+        return oldItem.url == newItem.url
+    }
+
+    override fun areContentsTheSame(oldItem: ContentItem, newItem: ContentItem): Boolean {
+        return true
+    }
```
Bnyro, line 13:
```suggestion
    override fun areContentsTheSame(oldItem: ContentItem, newItem: ContentItem): Boolean = true
```

app/src/main/java/com/github/libretube/ui/adapters/SearchChannelAdapter.kt
```diff
@@ -0,0 +1,182 @@
...
+        return when (currentList[position].type) {
+            StreamItem.TYPE_STREAM -> 0
+            StreamItem.TYPE_CHANNEL -> 1
+            StreamItem.TYPE_PLAYLIST -> 2
+            else -> 3
+        }
+    }
+
+    private fun bindVideo(item: ContentItem, binding: VideoRowBinding, position: Int) {
+        binding.apply {
```
Bnyro, line 72:
```suggestion
        binding.apply {
            thumbnail.setImageDrawable(null)
```
This should fix that the thumbnails are changing during scrolling by showing the placeholder instead.

The same should be done for all other thumbnails and channel avatars, also for `bingPlaylist` and `bindChannel`.

Bnyro (changes requested):

Bnyro:
> The same should be done for all other thumbnails and channel avatars, also for bindPlaylist and bindChannel.
>

PR #5544, opened 2024-01-23
refactor: Simplify ContentFilter toggling

Isira-Seneviratne:


app/src/main/java/com/github/libretube/enums/ContentFilter.kt
```diff
@@ -8,26 +8,18 @@ enum class ContentFilter {
...
     companion object {
-
         private val enabledFiltersSet get() = PreferenceHelper
-            .getString(
-                key = SELECTED_FEED_FILTERS,
-                defValue = entries.joinToString(",") { it.ordinal.toString() }
-            )
-            .split(',')
+            .getStringSet(SELECTED_FEED_FILTERS, entries.mapTo(mutableSetOf()) { it.name })
             .toMutableSet()
```
Bnyro, line 23: That line is redundant now?

Bnyro:
Please also change the prederence key value of "SELECTED_FEED_FILTERS" to avoid crashes when updating the app.

Isira-Seneviratne:
> Please also change the prederence key value of "SELECTED_FEED_FILTERS" to avoid crashes when updating the app.

Wasn't the content filter functionality introduced after the 0.21.1 release? In that case, there should be no issue as it's new anyway.

app/src/main/java/com/github/libretube/enums/ContentFilter.kt
Isira-Seneviratne: The add and remove methods are still used in the code, so I left the `toMutableSet` call.

Bnyro:
Yes, but there are some people using the debug builds for testing, so it will cause issues for them.

Bnyro:
I would have just changed the preference key so that the pref is being reset for all users - but that's fine with me too :+1:

Maybe we should also add a "TODO" or some comment about removing this "migration" code some time in the future, because the migration is probably needed by a few people using debug builds and can be removed in some months.

Bnyro (approved):

PR #5546, opened 2024-01-24
.

ghost:
This PR fixes issue #5141.
I took the approach to switch to audio player.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -141,4 +141,7 @@ dependencies {
     /* Baseline profile generation */
     implementation(libs.androidx.profileinstaller)
     "baselineProfile"(project(":baselineprofile"))
+
+    /* AndroidX Paging */
+    implementation(libs.androidx.paging)
 }
--- a/app/src/main/java/com/github/libretube/enums/ContentFilter.kt
+++ b/app/src/main/java/com/github/libretube/enums/ContentFilter.kt
@@ -8,26 +8,32 @@ enum class ContentFilter {
     SHORTS,
     LIVESTREAMS;
 
-    fun isEnabled() = enabledFiltersSet.contains(ordinal.toString())
+    var isEnabled
+        get() = name in enabledFiltersSet
+        set(enabled) {
+            val newFilters = enabledFiltersSet
+                .apply { if (enabled) add(name) else remove(name) }
 
-    fun setState(enabled: Boolean) {
-        val newFilters = enabledFiltersSet
-            .apply {if (enabled) add(ordinal.toString()) else remove(ordinal.toString()) }
-            .joinToString(",")
-
-        PreferenceHelper.putString(SELECTED_FEED_FILTERS, newFilters)
-    }
+            PreferenceHelper.putStringSet(SELECTED_FEED_FILTERS, newFilters)
+        }
 
     companion object {
+        private val enabledFiltersSet: MutableSet<String> get() {
+            val entryNames = try {
+                PreferenceHelper
+                    .getStringSet(SELECTED_FEED_FILTERS, entries.mapTo(mutableSetOf()) { it.name })
+            } catch (e: ClassCastException) {
+                // TODO: Remove the conversion code below.
+                // Assume the old preference is present and convert it.
+                val string = PreferenceHelper.getString(SELECTED_FEED_FILTERS, "")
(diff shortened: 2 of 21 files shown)


Bnyro:
This also means that the normal player is no longer in sync, which should be discussed first or made optional.

Pamilg8:
Lately the audio player is not working in the background

M00NJ:
Sounds like a pretty good idea imo. Though I think if it switches automatically to adio mode when the screen is turned off, it should also automatically switch back to video, when the device gets unlocked.

Pamilg8:
Also, we click the home button and use other applications. It would be great if the audio player starts when you tap the home button. For continuous playback in the background, we need to press the audio + home buttons, then return to the application and press the button again to open the video player. Considering that this is difficult, it is better if we correct it.

IndusAryan:
> Note: An option similar to opening a popup when you click the home button.

another option??, you are right in first like , whenever app is in background wheter screen off or on, it should switch to audio, but it'll be annoying if playback stops for sometime while changing the state or you already got the fix @arkibax ?

Pamilg8:
I just expressed my opinion. The issue is tagged as a discussion

> > Note: An option similar to opening a popup when you click the home button.
> 
> another option??, you are right in first like , whenever app is in background whether screen off or on, it should switch to audio, but it'll be annoying if playback stops for sometime while changing the state or you already got the fix @arkibax ?

IndusAryan:
> I just expressed my opinion. The issue is tagged as a discussion

lol, do i really sound that bad, i just asked if you want another option??, idc if it gets another option

akano12:
We need an option called "Thanos" that when toggled would randomly delete half of the options available in the settings

IndusAryan:
> We need an option called "Thanos mode" that when toggled would randomly delete half of the options available in the settings

i can add that if bnyro is serious about it.

second: i am also looking to remove many settings, but i dont know which setting to remove . team libretube should make a list of them in discussion.

Pamilg8:
> > We need an option called "Thanos mode" that when toggled would randomly delete half of the options available in the settings
> 
> i can add that if bnyro is serious about it.
> 
> second: i am also looking to remove many settings, but i dont know which setting to remove . team libretube should make a list of them in discussion.

Why are you deleting? I didn't understand

FireMasterK:
I think the best behaviour would be to have it switch to the background player when the screen it turned off, and switch back to the video player, seeking to the background player's position when the device is unlocked.

Pamilg8:
> I think the best behaviour would be to have it switch to the background player when the screen it turned off, and switch back to the video player, seeking to the background player's position when the device is unlocked.

Yes, he did

0xrxL:
From what I understood, this PR is stalled, due to some issue encountered by some users, including the lack of sync between audio and video (and viceversa) players during foreground/background switching.

I'm a dev...but not an Android dev (so I don't know how everything exactly works), but can't be eventually possible to run two player instances (video and audio together) and mute one of them depending by the app focus (foreground or background), to keep the sync and bypass every eventual bug caused by the solution in this PR?

IndusAryan:
> From what I understood, this PR is stalled, due to some issue encountered by some users, including the lack of sync between audio and video (and viceversa) players during foreground/background switching.
> 
> I'm a dev...but not an Android dev (so I don't know how everything exactly works), but can't be eventually possible to run two player instances (video and audio together) and mute one of them depending by the app focus (foreground or background), to keep the sync and bypass every eventual bug caused by the solution in this PR?

together one will work, but it will require some higher memory as well as cause increased data usage, and syncing problem wont get solved there too, suppose the video loads but the audio track doesnt, and it will also increase some load on piped

0xrxL:
> > From what I understood, this PR is stalled, due to some issue encountered by some users, including the lack of sync between audio and video (and viceversa) players during foreground/background switching.
> > I'm a dev...but not an Android dev (so I don't know how everything exactly works), but can't be eventually possible to run two player instances (video and audio together) and mute one of them depending by the app focus (foreground or background), to keep the sync and bypass every eventual bug caused by the solution in this PR?
> 
> together one will work, but it will require some higher memory as well as cause increased data usage, and syncing problem wont get solved there too, suppose the video loads but the audio track doesnt, and it will also increase some load on piped

Not necessarily...

For example: video player could split audio and video tracks, and use its instance to buffer both for itself and audio player too.

Assuming that all this can be done, of course...

PR #5587, opened 2024-02-01
feat: stop loading after long unsuccessful buffering

### LINKED ISSUE #5474: Give up timer on long buffering

# Problem:
The player sometimes enters an endless loading mode, when the screen is black and the user has no control over the player, other than closing it.

Rarely, the piped backend fixes the issue and the playback resumes, but it usually takes > 30 minutes, by which point, the user most likely gave up.

# Possible solution
Have a timer from when the last frame was received from the Piped instance and played. If it exceeds a predefined value (maybe 10 minutes), it gives up for good and doesn't try to start loading the video again until the user explicitly requests it.
This state could be indicated with the play button appearing, the same as when the video is paused.

Maybe this would only work when the screen is turned off?
Since if the video doesn't start playing even after 10 minutes, it is safe to assume the user gave up.

# TLDR
Give up on playing the video after N seconds from the last frame received (from Piped instance).

# Priority
Low.

### Other details

_No response_

### Acknowledgements

manish99verma:
closes #5474


Bnyro (approved):
Thanks!

PR #5593, opened 2024-02-04
fix: Resolve homepage not loading when 'trending' is not selected among home tab content

Jt90-spec:


Bnyro:
Thanks. Please don't disable

PR #5568, opened 2024-01-28
fix: resolve stuttering when scrolling to the bottom of a video feed #4282

abGit9:
This pull request makes the following changes:

Fixes issue #4282
Resolves stuttering for ChannelFragment and PlaylistFragment. Could not fix SubscriptionsFragment because toggling function between Subscriptions and Channels did not work.  Will have to revisit.

Issue #4282, opened 2023-07-20
Intense stuttering when scrolling to the bottom of a video feed several times


xlash123:
### Steps to reproduce

When scrolling through a channel, subscription, playlist, etc. feed, after a few instances scrolling to the bottom of the video feed, the app's performance starts to get worse. This can be seen as stuttering effect as more of the feed is loaded. This also causes the video player to take much longer to go into fullscreen mode while the large feed is in the background. All of these problems go away once the video feed is unloaded (going back to the home page, leaving the channel page, etc.).

Reproduction:
1. Open up your subscription feed or a YouTube channel page with many videos in their library
2. Scroll down to the bottom of the feed
3. Wait for the load
4. Repeat 2-3 about 10 times
5. Select a video to play
6. Go to fullscreen

### Expected behavior

1. Hitting the bottom of the feed will trigger a load, and the app is responsive during the loading time
2. Going into fullscreen occurs quickly

### Actual behavior

1. Each additional feed load takes more time than the previous, and during the loading time, the app is unresponsive
2. Going into fullscreen takes much longer, and the video player appears to be in an odd position for a few seconds

I believe that issue 1 can be partial solved with processing the API request asynchronously or on a separate thread from that of the rendering. That would at least allow the user to keep scrolling through the already loaded feed while the request completes. Not sure why the request is taking as long as it is, but I can understand if there are limitations with Piped.

I have no clue what would cause the weird fullscreen bug, so good luck with that.

### LibreTube version

0.16.1

### Android version

Android 11 (/e/ 1.12.3)

### Other details

A video demonstrating the bug has been published here: https://youtu.be/5DVTMFe0AhI

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

Bnyro:
API requests are already done on a separate thread and non-blocking.

Bnyro:
Yes, it seems to be a general issue when rendering large lists.

xlash123:
That would make sense. I also notice that it happens when I pull up my list of subscriptions, which is around 200 channels long.

Bnyro:
[...]

PR #5607, opened 2024-02-07
fix: resolve stuttering when scrolling to the bottom of channel/playlist videos list

abGit9:
This is a revision of a previous pull request : #5568

This is the description copied from that PR. 
//////////////////////////////////////////////////////////////
This pull request makes the following changes:
Partially addresses issue [#4282](https://github.com/libre-tube/LibreTube/issues/4282)
Resolves stuttering for ChannelFragment and PlaylistFragment. Could not fix SubscriptionsFragment because toggling function between Subscriptions and Channels did not work. Will have to revisit.

### DIFF
--- a/app/src/main/res/layout/fragment_channel.xml
+++ b/app/src/main/res/layout/fragment_channel.xml
@@ -7,8 +7,10 @@
     android:layout_height="match_parent">
 
     <androidx.coordinatorlayout.widget.CoordinatorLayout
+        android:id="@+id/channel_coordinator"
         android:layout_width="match_parent"
-        android:layout_height="match_parent">
+        android:layout_height="match_parent"
+        android:visibility="gone">
 
         <com.google.android.material.appbar.AppBarLayout
             android:id="@+id/channel_app_bar"
--- a/app/src/main/res/layout/fragment_channel.xml
+++ b/app/src/main/res/layout/fragment_channel.xml
@@ -6,165 +6,173 @@
     android:layout_width="match_parent"
     android:layout_height="match_parent">
 
-    <ScrollView
-        android:id="@+id/channel_scrollView"
+    <androidx.coordinatorlayout.widget.CoordinatorLayout
         android:layout_width="match_parent"
-        android:layout_height="match_parent"
-        android:orientation="vertical"
-        android:visibility="invisible"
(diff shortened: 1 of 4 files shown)


Bnyro (changes requested):
Thanks for re-implementing the changes.

I noticed that when opening a channel, the loading animation is gone and the main layout is already visible before the channel information got fetched from the api and loaded in the views - which looks very odd

app/src/main/res/layout/fragment_channel.xml
```diff
@@ -6,165 +6,173 @@
...
+        <com.google.android.material.appbar.AppBarLayout
+            android:id="@+id/channel_app_bar"
             android:layout_width="match_parent"
-            android:layout_height="wrap_content"
-            android:orientation="vertical">
+            android:layout_height="wrap_content">
 
-            <ImageView
-                android:id="@+id/channel_banner"
+            <com.google.android.material.appbar.CollapsingToolbarLayout
```
Bnyro, line 18: Why do we introduce a collapsing toolbar here although we're not actually collapsing anything?

abGit9: -the content above the list of videos needs to be scrollable. The CT layout provides that behavior to its children in synchrony with other components like Recycler view. If content is placed inside app bar layout without being wrapped in the CT, it will remain fixed and not scroll up. Although in CT, it does not appear as if the content is collapsing it is possible that the container is and the content is remaining fixed as it is cut off in view as you scroll.
-an alternative to using CT would be to implement adapters that inflate different types of views so that the top views above the list can scroll off the screen in the same container as the list videos. 
-I feel that implementing CT in xml is easier and avoids adding more programmatic complexity to recyclerview adapters.

Bnyro (commented):
The same issue applies for the `PlaylistFragment` unfortunately. Apart from that the changes look good to me :)

abGit9:
I'll work on it in the coming days. Thanks.

abGit9:
Hi, any other issues that need to be addressed?

Bnyro (approved):
This really works great, thank you!

abGit9:
Thank you!

abGit9:
> This really works great, thank you!

Thank you!

Issue #5626, opened 2024-02-12
Double tap to seek causes Playback speed to return x1.0 [some videos]


Ehab-d:
### Steps to reproduce

1. Open the [video](https://youtu.be/tKrkdDp4OfU?si=Z9eC5HaahJejiv-8)
*This problem occurs with some videos like this... I can't determine why
2. Change Playback speed
3. Double tap to seek ↻

### Expected behavior

The video should continue at the selected Playback speed.

### Actual behavior

The video Playback speed returns to 1.0x.

### LibreTube version

0.21.1 [Latest]

### Android version

Android 14

### Other details

- If we open another video, the video will play at the specified Playback speed.
- The issue is possible to reproduce in all cases.
- I can record a video of it if necessary.

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

Ehab-d:
Also, I tried the same video on NewPipe and the issue did not occur.
*To point out that NewPipe Extractor is not the reason maybe.

Ehab-d:
Thanks! I tested it in LibreTube Debug and it was fixed.

PR #5634, opened 2024-02-18
fix: set music playback speed only once

FineFindus:
Fixes https://github.com/libre-tube/LibreTube/issues/5626. The playback speed for music videos was set to 1.0x on every event, instead of only once when starting the video.


Bnyro (approved):
lgtm, thanks!

PR #5668, opened 2024-02-27
feat(ui): option for automatic update checks

IndusAryan:
the blunders of previous pr are deeply regretted @Bnyro . those should not be existing in first place.

photo_1_2024-02-27_17-22-51

photo_2_2024-02-27_17-22-51


app/src/main/java/com/github/libretube/ui/dialogs/UpdateAvailableDialog.kt
```diff
@@ -1,33 +1,35 @@
...
+        changelog: String,
+        releaseUrl: String,
+        context: Context,
+    ): Dialog {
+        return MaterialAlertDialogBuilder(context)
+            .setTitle(R.string.update_available)
+            .setMessage(changelog)
+            .setPositiveButton(R.string.download, null)
+            .setNegativeButton(R.string.tooltip_dismiss, null)
+            .setCancelable(false)
```
Bnyro, line 25: Just out of curiosity, why do we make this non-cancelable?

app/src/main/java/com/github/libretube/ui/dialogs/UpdateAvailableDialog.kt
```diff
@@ -1,33 +1,35 @@
...
-            .setNegativeButton(R.string.cancel, null)
-            .setPositiveButton(context?.getString(R.string.okay)) { _, _ ->
-                val intent = Intent(Intent.ACTION_VIEW).setData(updateInfo.htmlUrl.toUri())
-                startActivity(intent)
-            }
+    fun onCreateDialog(
+        changelog: String,
+        releaseUrl: String,
+        context: Context,
+    ): Dialog {
```
Bnyro, line 19: Please refactor this so that the arguments are passed as intent extras, basically as it's been before.

This implementation might crash/dismiss the dialog once the theme is changed on certain devices.

app/src/main/java/com/github/libretube/ui/dialogs/UpdateAvailableDialog.kt
IndusAryan: as it pops up on app open like a regular update prompt, user can touch outside dialog frame at same time plus it already has a dismiss button.

app/src/main/java/com/github/libretube/ui/dialogs/UpdateAvailableDialog.kt
IndusAryan: i have used the intent bundle for passing data and have tested well on android 9 and 13 , the lint is asking for SNAKE-CASE  in intent data for these two things(changelog and url), should i do it or ignore it because all others are in camelCase

Bnyro (approved):
Thank you!

Bnyro:
You don't need to fix ktlint next time, I usually ignore it as well and just format the whole code from time to time.

Issue #5692, opened 2024-03-02
Mini Player Overlaps Navigation Bar


akano12:
### Steps to reproduce

1. Open any LT supported link like this one for example https://piped.video/watch?v=mtaQroi75M0

### Expected behavior

Video loads normally 

### Actual behavior

Video opens minimized and overlapping navigation bar

### LibreTube version

Latest stable and debug 

### Android version

Android 13

### Other details

I'm aware the video being minimized is reported in https://github.com/libre-tube/LibreTube/issues/3260 

https://github.com/libre-tube/LibreTube/assets/50032603/331155c4-3b5c-4758-925f-35dc49dd46a5

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

PR #5693, opened 2024-03-03
-Fix mini player overlapping navigation bar (#5692)

KhubaibKhan4:
Just Fixed the Issue of player overlapping while using the Motion Layout.


KhubaibKhan4:
Just closed this because the Git Action was failing due to unused imports.

PR #5726, opened 2024-03-11
fix: false message about no comments and btnTop scroll

nik-conder:
In new versions (not yet released), comments are broken. Including responses to comments. 
I tried to fix btn Top and corrected the false text output that there are no comments when in fact there are. I haven't figured out the reason yet, so the code may need to be adjusted or supplemented.

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/fragments/CommentsRepliesFragment.kt
+++ b/app/src/main/java/com/github/libretube/ui/fragments/CommentsRepliesFragment.kt
@@ -4,6 +4,7 @@ import android.os.Bundle
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
+import android.view.ViewTreeObserver
 import androidx.core.view.isGone
 import androidx.core.view.isVisible
 import androidx.core.view.updatePadding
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/github/libretube/ui/fragments/CommentsRepliesFragment.kt
```diff
@@ -88,8 +119,18 @@ class CommentsRepliesFragment : Fragment() {
...
+        scrollListener = null
+    }
+
     override fun onDestroyView() {
         super.onDestroyView()
         _binding = null
     }
+
+    companion object {
+        const val POSITION_START = 0
```
Isira-Seneviratne, line 134:
```suggestion
        private const val POSITION_START = 0
```

Isira-Seneviratne (changes requested):
Looks good to me overall, thanks for the comment fix. I've suggested a couple of small changes.

nik-conder:
Sorry, the following commit accidentally flew in here, which does not relate to the previous fix. Isn't that scary?

Bnyro:
Merged in https://github.com/libre-tube/LibreTube/pull/5751, thanks!

Please don't merge the LibreTube master branch into yours next time and don't disable that others can push your branch.

nik-conder:
@Bnyro Thank you very much. Sorry again, I'm just learning.

> Please don't merge the LibreTube master branch into yours next time

Please specify what is meant here?

Issue #5765, opened 2024-03-18
Default quality setting should respect video aspect ratio


DahilerMehvOlur:
Hello
For example I set default video quality to 720p. When I open horizontal video it works. The video opens automatically in 720p but when I open a vertical or shorts video it opens in 640p. It should be open in 1280p. Because 720p video size is 1280x720.
Please check this video.
https://streamable.com/tu0jc4

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

PR #5797, opened 2024-03-23
fix: default quality setting should respect video aspect ratio

manish99verma:
This fixes the issue #5765 where the default quality settings should respect aspect ratio. Now when we will open the shorts and default quality setting is 720p then it will auto play with 1280p.

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
+++ b/app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
@@ -117,6 +117,7 @@ import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
 import java.util.concurrent.Executors
 import kotlin.math.abs
+import kotlin.math.ceil
 
 @androidx.annotation.OptIn(androidx.media3.common.util.UnstableApi::class)
 class PlayerFragment : Fragment(), OnlinePlayerOptions {
@@ -136,6 +137,7 @@ class PlayerFragment : Fragment(), OnlinePlayerOptions {
     private var channelId: String? = null
     private var keepQueue = false
     private var timeStamp = 0L
+    private var isShort = false
 
     // data and objects stored for the player
     private lateinit var exoPlayer: ExoPlayer
@@ -922,7 +924,7 @@ class PlayerFragment : Fragment(), OnlinePlayerOptions {
             }
 
             val videoStream = streams.videoStreams.firstOrNull()
-            val isShort = PlayingQueue.getCurrent()?.isShort == true ||
+            isShort = PlayingQueue.getCurrent()?.isShort == true ||
                 (videoStream?.height ?: 0) > (videoStream?.width ?: 0)
 
             PlayingQueue.setOnQueueTapListener { streamItem ->
@@ -1252,10 +1254,16 @@ class PlayerFragment : Fragment(), OnlinePlayerOptions {
         }
     }
 
-    private fun setPlayerResolution(resolution: Int) {
+    private fun setPlayerResolution(resolution: Int, isSelectedByUser: Boolean = false) {
+        val transformedResolution = if (!isSelectedByUser && isShort) {
+            ceil(resolution * 16.0 / 9.0).toInt()
+        } else {
+            resolution
+        }
+
         trackSelector.updateParameters {
-            setMaxVideoSize(Int.MAX_VALUE, resolution)
-            setMinVideoSize(Int.MIN_VALUE, resolution)
+            setMaxVideoSize(Int.MAX_VALUE, transformedResolution)
+            setMinVideoSize(Int.MIN_VALUE, transformedResolution)
         }
     }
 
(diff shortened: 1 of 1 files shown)


Bnyro (changes requested):
When autoplay continues and the next video is a short again, the factor will be `(16/9)^2` after your changes.
And when the next video is a normal one and not a short, it still uses the resolution for shorts.

TLDR: Your code shouldn't change the `fullscreenOrientation` variables, it should only read them in the function where the resolution is set and set it to `16/9 * resolution` ditectly instead of modifying the variable.

manish99verma:
Now it will set resolution directly without modifing the default variables and when user will change resoultion for Shorts, let say to 1280 then it will save 720 as a default resoultion for current queue.

Bnyro:
I think it would be better to do this in `setPlayerResolution` instead of at 3 different places.

manish99verma:
That would work. But, i think it will create some problem when user will select a custom resolution while playing a video. So we can pass a boolean to setPlayerResolution whether it is choosed by user or it's default.

Bnyro:
> That would work. But, i think it will create some problem when user will select a custom resolution while playing a video. So we can pass a boolean to setPlayerResolution whether it is choosed by user or it's default.
>
Yes, sounds good :+1:

app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
```diff
@@ -1298,10 +1300,14 @@ class PlayerFragment : Fragment(), OnlinePlayerOptions {
         }
     }
 
-    private fun setPlayerResolution(resolution: Int) {
+    private fun setPlayerResolution(resolution: Int, isSelectedByUser: Boolean) {
+        var transformedResolution = resolution
+        if (!isSelectedByUser && isShort) {
+            transformedResolution = ceil(resolution * 16.0 / 9.0).toInt()
+        }
```
Bnyro, line 1307:
```suggestion
        val transformedResolution = if (!isSelectedByUser && isShort) {
            ceil(resolution * 16.0 / 9.0).toInt()
        } else {
            resolution
        }
```

app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
```diff
@@ -1487,10 +1493,13 @@ class PlayerFragment : Fragment(), OnlinePlayerOptions {
...
             ) { which ->
-                val newResolution = resolutions[which].resolution
-                setPlayerResolution(newResolution)
+                var newResolution = resolutions[which].resolution
+                setPlayerResolution(newResolution, true)
 
                 // save the selected resolution to update on fullscreen change
+                if (isShort) {
+                    newResolution = ceil(newResolution * 9.0 / 16.0).toInt()
+                }
```
Bnyro, line 1502: I think this code here should be removed, since we already handle that in `setPlayerResolution`?

app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
```diff
@@ -1298,10 +1300,16 @@ class PlayerFragment : Fragment(), OnlinePlayerOptions {
         }
     }
 
-    private fun setPlayerResolution(resolution: Int) {
+    private fun setPlayerResolution(resolution: Int, isSelectedByUser: Boolean) {
```
Bnyro, line 1303:
```suggestion
    private fun setPlayerResolution(resolution: Int, isSelectedByUser: Boolean = false) {
```
simplifies things.

app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
Bnyro: I don't really like/agree on this change, it's not a good idea to modify the `newResolution` variable, it should always be in the "same format" (normal horizontal video ratio resolution).

Bnyro: This here makes everything more complicated and I don't think anybody knows what's going on when maintaining that code in the future.

app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
```diff
@@ -1487,10 +1495,13 @@ class PlayerFragment : Fragment(), OnlinePlayerOptions {
                     if (currentQuality == it.resolution) "${it.name} ✓" else it.name
                 }
             ) { which ->
-                val newResolution = resolutions[which].resolution
-                setPlayerResolution(newResolution)
+                var newResolution = resolutions[which].resolution
+                setPlayerResolution(newResolution, true)
 
                 // save the selected resolution to update on fullscreen change
```
Bnyro, line 1501: the comment should be moved below where the actual resolution update code is.

Bnyro (approved):

Issue #5800, opened 2024-03-24
Auto-fullscreen doesn't work after switching to another app


LavellanDWR:
### Steps to reproduce

1. Enable the auto-fullscreen setting. Note: the Android auto rotate is off, i.e. it only rotates with the pop-up button. 
2. Play a video and rotate the device to enter fullscreen landscape mode.
3. Pause the video, rotate back to portrait mode.
4. Switch to another application by minimising LibreTube, opening the recents menu or swiping at the bottom of the screen. 
5. Switch back to LibreTube by either method and rotate the device. 

### Expected behavior

The video is supposed to go full-screen automatically, turning to landscape mode.  

### Actual behavior

The orientation does not change until you minimise the video by swiping down, then tapping on the minimised player and rotating the device again. 

### LibreTube version

0.22.1

### Android version

Android 13

### Other details

This issue was introduced in a recent update, either v0.22.1 or v0.22.0.

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

LavellanDWR:
It doesn't work if you rotate it more than once, either.

jx319:
This also happens to me (Android 11, LibreTube 0.22.1)

PR #5807, opened 2024-03-26
fix: Auto-fullscreen doesn't work after switching to another app

nik-conder:
I tried to understand the reason for this issue (https://github.com/libre-tube/LibreTube/issues/5800), deleting part of the code helped.

The auto orientation still performs its functions when needed and it does not break after the second auto-rotation (i.e. 1 - to landscape, 2 - to portrait, etc.).

In addition, auto-orientation works even after exiting and returning to the application (the steps described in the issue).

p.s. with this part of the code, the `onConfigurationChanged()` method does not respond to the auto-orientation of the screen at all for some reason,  after the second auto-rotation. Problem maybe `restartActivityIfNeeded()` method. Therefore, I hope for the help of more experienced colleagues.

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
+++ b/app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
@@ -1687,17 +1687,6 @@ class PlayerFragment : Fragment(), OnlinePlayerOptions {
         ) {
             return
         }
-
-        if (PlayerHelper.autoFullscreenEnabled) {
-            when (newConfig.orientation) {
-                // go to fullscreen mode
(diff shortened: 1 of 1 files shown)


Bnyro:
This PR completely removes the auto fullscreen logic (Settings -> Player -> Auto fullscreen). With this change, that option would be absolutely useless.

This might fix auto rotation, but that's not what the issue is about.

nik-conder:
Good. But maybe this thought will help someone. Or I'll try to figure out the problem again myself.

Bnyro:
I think I know what's causing the issue, I will write a patch once I have some free time again.

PR #5817, opened 2024-03-26
feat: improve video end screen

FineFindus:
Fixes a problem where the controller was not displayed on the video end screen when "Insert related videos" was disabled. The improved version should now cover edge cases like the last video of a playlist.
I also changed it to show the controller on the end screen permanently (i.e. it doesn't hide after a few seconds), as I found that it was easy to miss.

### DIFF
--- a/app/src/main/java/com/github/libretube/helpers/PlayerHelper.kt
+++ b/app/src/main/java/com/github/libretube/helpers/PlayerHelper.kt
@@ -345,8 +345,10 @@ object PlayerHelper {
 
     fun shouldPlayNextVideo(isPlaylist: Boolean = false): Boolean {
         // if there is no next video, it obviously should not be played
-        if (PlayingQueue.getNext() == null)
+        if (!PlayingQueue.hasNext()) {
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/github/libretube/helpers/PlayerHelper.kt
```diff
@@ -343,6 +344,11 @@ object PlayerHelper {
             )
 
     fun shouldPlayNextVideo(isPlaylist: Boolean = false): Boolean {
+        // if there is no next video, it obviously should not be played
+        if (PlayingQueue.getNext() == null) {
```
Bnyro, line 348:
```suggestion
        if (!PlayingQueue.hasNext()) {
```

FineFindus: I'm not a fan of this style, as I think it makes the code harder to read, but if you prefer it that way…

Bnyro (approved):
Merci :+1:

PR #5860, opened 2024-04-03
chore: remove example unit/instrumentation tests

cortinico:
While browsing this project, I realized you still have the examples JUnit and Instrumentation tests created by Android Studio.

The example JUnit tests were not compiling, so invoking `./gradlew build` was failing.
The example instrumentation test was working fine but you were not using and it was just making the whole Gradle build slower (I also saw you have baseline profiles set up, so you should use them).


Bnyro (approved):

Bnyro:
Thanks :+1:

PR #5931, opened 2024-04-22
fix: crash when importing subscriptions list with invalid avatars

bodaciousbiscuits:
Solves issue of subscription list not loading when channels have old/invalid avatars.

See issue below:

https://github.com/libre-tube/LibreTube/assets/167443179/3d61bab9-52d2-44ab-bb8e-eaa0774c15fb

### DIFF
--- a/app/src/main/java/com/github/libretube/api/obj/Subscription.kt
+++ b/app/src/main/java/com/github/libretube/api/obj/Subscription.kt
@@ -6,6 +6,6 @@ import kotlinx.serialization.Serializable
 data class Subscription(
     val url: String,
     val name: String,
-    val avatar: String,
+    val avatar: String? = null,
     val verified: Boolean
 )


bodaciousbiscuits:
While debugging the problem, the null pointer showed up in the logs:

```bash
04-17 07:51:48.227 19814  4443 E kotlinx.coroutines.StandaloneCoroutine: kotlinx.serialization.json.internal.JsonDecodingException: Unexpected JSON token at offset 27191: Expected string literal but 'null' literal was found at path: $[111].avatar
04-17 07:51:48.227 19814  4443 E kotlinx.coroutines.StandaloneCoroutine: Use 'coerceInputValues = true' in 'Json {}' builder to coerce nulls if property has a default value.
04-17 07:51:48.227 19814  4443 E kotlinx.coroutines.StandaloneCoroutine: JSON input: .....ame":"Channel","avatar":null,"verified":true},{"url":".....
```

Bnyro (approved):
Thanks :+1:

PR #5937, opened 2024-04-26
fix: remove setUsePlatformDiagnostics being called twice

FireMasterK:
(It's called just a few lines above)


Bnyro (approved):
Ah yes, thanks 👍

PR #6035, opened 2024-05-14
fix: button to play previous video doesnt work when repeat enabled

Bnyro:


danielss02389:
I can't watch videos, the connection always fails

PR #6036, opened 2024-05-14
chore: dont share with time code by default

Bnyro:


Bnyro:
> For this #4857?
>
No, just in general because most users probably won't use it / need it by default.

It's still remembered.

Issue #5833, opened 2024-03-28
Media controls notification still there


Ehab-d:
### Steps to reproduce

1. Open a YouTube video
2. Close LibreTube from Recents
3. Media controls notification still there

### Expected behavior

The media controls notification should disappear after closing the app from the Recents.

### Actual behavior

The media controls notification still remains. You have to dismiss the notification manually.

### LibreTube version

0.22.2

### Android version

Android 10

### Other details

A screen record to clarify the issue.

https://github.com/libre-tube/LibreTube/assets/129071698/6e1e3e16-2f12-4c64-a060-68f111a4463c

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

Bnyro:
https://github.com/libre-tube/LibreTube/pull/5998 attempts to fix this by improving the code that's running when the app is being killed, please try the latest builds form https://github.com/libre-tube/NightlyBuilds and report back if the issue still exists.

Ehab-d:
Thanks. I have tried LibreTube 0.23.1 and Nightly Builds as well and can confirm the issue persists.
The issue is not about killing the app in general. It can be reproduced by simply closing the player.

https://github.com/libre-tube/LibreTube/assets/129071698/fa150a5f-a163-40cd-b70a-8c6dae9ce75f

Ehab-d:
However, I see improvement in the code, as it has been resolved #5958 for me. 🚀

Bnyro:
Hmm, I noticed that some other apps have the issue with the notification too (AntennaPod, VibeYou, and probably more). Not really sure how we can resolve that :/

Ehab-d:
No problem. That's why it's a low-priority issue. You will be able to solve it one day.

PR #6063, opened 2024-05-19
fix: Clear notification when app is killed from recent tasks. 

nur-shuvo:
When app is killed from recent task, notification still persists.
This PR fixes https://github.com/libre-tube/LibreTube/issues/5833


app/src/main/java/com/github/libretube/services/OnClearFromRecentService.kt
```diff
@@ -0,0 +1,33 @@
...
+        return START_NOT_STICKY
+    }
+
+    override fun onTaskRemoved(rootIntent: Intent?) {
+        nManager?.cancel(PLAYER_NOTIFICATION_ID)
+        stopSelf()
+    }
+
+    companion object {
+        private const val PLAYER_NOTIFICATION_ID = 1
```
Bnyro, line 31: That constant is already defined somewhere else (iirc in NotificationHelper.kt`)

app/src/main/java/com/github/libretube/services/OnClearFromRecentService.kt
```diff
@@ -0,0 +1,33 @@
...
+import android.app.NotificationManager
+import android.app.Service
+import android.content.Intent
+import android.os.IBinder
+import androidx.core.content.getSystemService
+
+class OnClearFromRecentService : Service() {
+    private var nManager: NotificationManager? = null
+
+    override fun onBind(p0: Intent?): IBinder? {
```
Bnyro, line 12:
```suggestion
    override fun onBind(p0: Intent?): IBinder? = null
```

app/src/main/java/com/github/libretube/services/OnClearFromRecentService.kt
nur-shuvo: I used it, but in the remote debug build it cannot resolve it, i don't know why. Will I do it ?

nur-shuvo:
@Bnyro Can you please check?

Bnyro (approved):
Thanks!

PR #6065, opened 2024-05-20
fix: use string ressource for channel video tab

FineFindus:
Fixes an issue, were the Channel video tab name was still using the `VIDEOS_TAB_KEY` as its display name.

| Before                                                                                                                          	| After                                                                                                                                           	|
|---------------------------------------------------------------------------------------------------------------------------------	|-------------------------------------------------------------------------------------------------------------------------------------------------	|
| Channel tabbar with wrong name 	| Channel tabbar with correctly capitalized name 	|


Bnyro (approved):
Thanks!

PR #6111, opened 2024-06-06
refactor: Implement the subscribe button in Jetpack Compose

Isira-Seneviratne:
[Jetpack Compose](https://developer.android.com/develop/ui/compose) is a declarative UI framework for Android that simplifies Android development greatly.

This PR adds Compose to the project and migrates the subscribe button functionality to use Compose.

To be done:

* Theming
* Subscribe/unsubscribe dialogs

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -19,6 +19,9 @@ android {
         versionName = "0.23.2"
         testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
         resValue("string", "app_name", "LibreTube")
+        vectorDrawables {
+            useSupportLibrary = true
+        }
 
         ksp {
             arg("room.schemaLocation", "$projectDir/schemas")
@@ -69,6 +72,9 @@ android {
 
     packaging {
         jniLibs.excludes.add("lib/armeabi-v7a/*_neon.so")
+        resources {
+            excludes += "/META-INF/{AL2.0,LGPL2.1}"
+        }
     }
 
     tasks.register("testClasses")
@@ -80,9 +86,13 @@ android {
 
     buildFeatures {
         buildConfig = true
+        compose = true
     }
 
     namespace = "com.github.libretube"
+    composeOptions {
+        kotlinCompilerExtensionVersion = "1.5.14"
+    }
 }
 
 dependencies {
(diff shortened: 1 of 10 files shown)


Bnyro:
To be honest I'm not really sure about migrating to Compose currently - we should probably discuss this first. If we only migrate parts of the app to Compose there will most likely be behavior and design/theme differences, so that doesn't really feel good. If we consider it, we should stop doing any updates for multiple months and take that time to fully rewrite the app in Compose.

I quite like Jetpack Compose, but still this would be lots and lots of efforts to migrate LibreTube to it.

I don't have much free time at the moment and this would need a lot of time testing/bug fixing obviously.

Isira-Seneviratne:
> To be honest I'm not really sure about migrating to Compose currently - we should probably discuss this first. If we only migrate parts of the app to Compose there will most likely be behavior and design/theme differences, so that doesn't really feel good. If we consider it, we should stop doing any updates for multiple months and take that time to fully rewrite the app in Compose.
> 
> I quite like Jetpack Compose, but still this would be lots and lots of efforts to migrate LibreTube to it.
> 
> I don't have much free time at the moment and this would need a lot of time testing/bug fixing obviously.

Migrating to Compose piece by piece is [a valid strategy](https://developer.android.com/develop/ui/compose/migrate/strategy). That way, development doesn't have to be paused.

Bnyro:
https://github.com/material-components/material-components-android is still actively developed and there has never been anyone saying that it would become deprecated/abandoned anytime soon. The app is pretty much feature-complete and thus I don't think it would be any worth it to spend efforts into migrating to a new framework.

We have no idea how long LibreTube will continue to work, and that can really just become hit or miss.

Isira-Seneviratne:
It would reduce a lot of unnecessary code, e.g. the fragments could be replaced with separate composables nested in a [scaffold](https://foso.github.io/Jetpack-Compose-Playground/material/scaffold/).

IndusAryan:
Hi guys, nice pr bro, xml is not going anywhere, in the current state libretube will work for many years, XML isn't going anywhere, problem with jetpack is that it lacks the full required documentation, although i am not your team member but as an old contrrubutor i can say Isira is skilled enough to rewrite the app slowly, So i think he can continue development in another branch but he will face problems as other additions will be pushed continiously to main branch as bnyro said, what if you guys release a full stable LT that doesnt require frequent update ?

Bnyro:
As seen recently, it's only a matter of time until YouTube fully makes 3rd party YouTube apps impossible and I thus don't think it's anyhow reasonable to invest a lot of time for rewriting the app.

PR #6142, opened 2024-06-16
fix(SubscriptionsFragment): use null safe call

FineFindus:
Fixes a crash when trying to mark a video as played, that is above the all caught up item.


Bnyro (approved):

Bnyro:
Thanks 👍

Issue #5754, opened 2024-03-17
Video is interrupted on orientation change


imwints:
### Steps to reproduce

* Enable auto-rotate in app and for device
* Start video
* Rotate device

### Expected behavior

Video should play seamlessly

### Actual behavior

Video is interrupted quickly

### LibreTube version

Nightly

### Android version

14

### Other details

This is a followup of https://github.com/libre-tube/LibreTube/issues/5727. I observed this before the fixing commit as well, so this shouldn't be a regression. 

I'm sadly not at home and don't have a machine with Android Studio at hand, otherwise I would debug this properly. A quick search on the internet spits out [this reddit](https://www.reddit.com/r/androiddev/comments/3vwgg9/video_restarts_on_orientation_change_when/) post but the manifest declares the orientation config. So my guess it that it's an issue in this app and how it handles orientation changes.

Maybe you know if this is fixable at all?

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

Bnyro:
Duplicate of https://github.com/libre-tube/LibreTube/issues/5538.

Unfortunately, there's no better way to achieve this. I've already tried to make the transition as seamless as possible in https://github.com/libre-tube/LibreTube/pull/5635, but that's the best we can get.

When rotating the phone, we need to use a different player layout - hence we must recreate the current layout to load the landscape version when you started in portrait mode and vice versa.

That just takes some time - as things must get rendered first.

imwints:
Loading into fullscreen from portrait mode doesn't have this issue. Do you think it would work if the landscape mode would be disabled for small screens and rotating would launch fullscreen mode? I find landscape mode useless on my phone anyway.

Bnyro:
> Loading into fullscreen from portrait mode doesn't have this issue. Do you think it would work if the landscape mode would be disabled for small screens and rotating would launch fullscreen mode? I find landscape mode useless on my phone anyway.
>
That's already possible:

Settings -> General -> Disable auto rotation

Player -> Enable Auto fullscreen

imwints:
Ah alright, I was thinking that this would launch every new Video in fullscreen

Pittvandewitt:
> Duplicate of #5538.
> 
[...]

PR #6307, opened 2024-07-30
fix: improve exoplayer scoping

Pittvandewitt:
This PR splits the PlayerViewModel, so that each fragment or activity where an exoplayer is used has its own properly scoped exoplayer. The result is that an exoplayer is created when the viewmodel of a fragment that needs a player is created and destroyed when the viewmodel is. The exoplayer is no longer recreated upon orientation change.
This means that media keeps playing without interruptions upon orientation change, fixing https://github.com/libre-tube/LibreTube/issues/5754#issuecomment-2230222744

The exoplayer no longer lives inside the player fragment or in the activity scoped viewmodel.


Bnyro (approved):
I don't feel any difference when rotating the screen honestly, but it's reasonable to move the player ownership to the view model.

Thanks!

Pittvandewitt:
> I don't feel any difference when rotating the screen honestly, but it's reasonable to move the player ownership to the view model.
> 
> Thanks!

I don't feel anything in the update in which this PR is merged either. I'll go back to the drawing board when I got some spare time again to find out why.

PR #6314, opened 2024-08-04
fix: visibility of UI elements in Downloads

nik-conder:
In this PR, the visibility of some UI elements has been fixed when the list is empty.


Bnyro (approved):
I've cleaned up the code a small bit, thanks for the PR!

PR #6428, opened 2024-08-31
fix: retain exoplayer on orientation change

Pittvandewitt:
This PR is basically the last step of https://github.com/libre-tube/LibreTube/pull/6307
Now you should feel the improvement upon rotation change ;)

The ExoPlayer is no longer stopped on onDestroy, because it is already released in its viewmodel onDestroy method.

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
+++ b/app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
@@ -845,12 +845,6 @@ class PlayerFragment : Fragment(), OnlinePlayerOptions {
         viewModel.player.removeListener(playerListener)
         viewModel.player.pause()
 
-        runCatching {
-            if (!viewModel.isOrientationChangeInProgress) {
-                viewModel.player.stop()
-            }
-        }
-
         if (PlayerHelper.pipEnabled) {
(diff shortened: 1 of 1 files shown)


Bnyro (approved):

Bnyro (changes requested):
Wait, this would also mean that the current player video stays alive when the user closes a video and it still continues to buffer.

Apart from that `viewModel.isOrientationChange` already conditionally does what you're trying to achieve with this PR, doesn't it?

Pittvandewitt:
When the video closes with `PlayerFragment`, `PlayerViewModel` is also destroyed and `PlayerViewModel.onCleared` is called, where the exoplayer is released. Or does `ExoPlayer.stop` also need to be called before `ExoPlayer.release`?

With this PR I notice no interruption when changing orientation, but with the current implementation of `viewModel.isOrientationChange` I always do, especially when pressing the full screen button.

Bnyro:
> When the video closes with PlayerFragment, PlayerViewModel is also destroyed and PlayerViewModel.onCleared is called,
>
You're right, I thought that would only be called in the activity's `onDestroy` and not in the fragments, but I tested it and it looks like you're right.

Bnyro (approved):

PR #6483, opened 2024-09-10
feat: RYD support for local stream extraction

### LINKED ISSUE #6412: Return YouTube dislike support for local streams extraction
### Steps to reproduce

1. Load a video with enabled "Local stream extraction" option and see dislikes count.
2. Load the same video with disabled "Local stream extraction" option and see dislikes count.

### Expected behavior

Dislikes count is shown correctly OR dislikes entry is hidden when Local stream extraction is enabled

### Actual behavior

Zero dislikes

### LibreTube version

0.25.1

### Android version

Any

### Other details

I assume this happens because dislikes count is proxied through Piped, and local extraction from `returnyoutubedislikeapi.com` isn't implemented yet.

As adding local extraction from `returnyoutubedislikeapi.com` may be controversial (closed server and database), it may be wise to disable this option by default. When disabled, dislikes counter should be hidden.

### Acknowledgements

janisslsm:
fixes #6412 

perhaps adding a toggle under local stream extraction would be useful in case someone doesn't want it? I wasn't sure so I didn't add one.

### DIFF
--- a/app/src/main/java/com/github/libretube/api/StreamsExtractor.kt
+++ b/app/src/main/java/com/github/libretube/api/StreamsExtractor.kt
@@ -58,7 +58,9 @@ object StreamsExtractor {
             category = resp.category,
             views = resp.viewCount,
             likes = resp.likeCount,
-            dislikes = RetrofitInstance.externalApi.getVotes(videoId).dislikes,
+            dislikes = if (PlayerHelper.localRYD)
+                RetrofitInstance.externalApi.getVotes(videoId).dislikes
+            else 0,
             license = resp.licence,
             hls = resp.hlsUrl,
             dash = resp.dashMpdUrl,
--- a/app/src/main/java/com/github/libretube/api/StreamsExtractor.kt
+++ b/app/src/main/java/com/github/libretube/api/StreamsExtractor.kt
@@ -58,9 +58,11 @@ object StreamsExtractor {
             category = resp.category,
             views = resp.viewCount,
             likes = resp.likeCount,
-            dislikes = if (PlayerHelper.localRYD)
-                RetrofitInstance.externalApi.getVotes(videoId).dislikes
-            else 0,
+            dislikes = runCatching {
+                if (PlayerHelper.localRYD)
+                    RetrofitInstance.externalApi.getVotes(videoId).dislikes
+                else 0
+            }.getOrElse { 0 },
             license = resp.licence,
             hls = resp.hlsUrl,
             dash = resp.dashMpdUrl,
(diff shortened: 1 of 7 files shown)


Bnyro:
I'd prefer to make this a preference (enabled by default)

janisslsm:
> I'd prefer to make this a preference (enabled by default)

Screenshot_20240912_185856
is this fine?

ChunkyProgrammer:
There's also ryd proxy, not sure if you'd want to support using that as well https://github.com/TeamPiped/RYD-Proxy (the api call is slightly different)

app/src/main/java/com/github/libretube/api/StreamsExtractor.kt
```diff
@@ -58,6 +58,9 @@ object StreamsExtractor {
             category = resp.category,
             views = resp.viewCount,
             likes = resp.likeCount,
+            dislikes = if (PlayerHelper.localRYD)
+                RetrofitInstance.externalApi.getVotes(videoId).dislikes
```
Bnyro, line 62: This should be inside `runCatching`

Bnyro:
> There's also ryd proxy, not sure if you'd want to support using that as well https://github.com/TeamPiped/RYD-Proxy (the api call is slightly different)
>
I think that's an overkill for now.

app/src/main/java/com/github/libretube/api/StreamsExtractor.kt
janisslsm: fixed

app/src/main/java/com/github/libretube/api/StreamsExtractor.kt
```diff
@@ -58,9 +58,11 @@ object StreamsExtractor {
...
             views = resp.viewCount,
             likes = resp.likeCount,
-            dislikes = if (PlayerHelper.localRYD)
-                RetrofitInstance.externalApi.getVotes(videoId).dislikes
-            else 0,
+            dislikes = runCatching {
+                if (PlayerHelper.localRYD)
+                    RetrofitInstance.externalApi.getVotes(videoId).dislikes
+                else 0
+            }.getOrElse { 0 },
```
Bnyro, line 65:
```suggestion
            dislikes = if (PlayerHelper.localRYD) runCatching {
                RetrofitInstance.externalApi.getVotes(videoId).dislikes
            }.getOrElse { -1 } else -1,
```
The try catch block doesn't need to be executed when the option is disabled anyways.

Apart from that we might want to return -1 instead of 0 elsewise since 0 could be interpreted as a normal dislike count

Bnyro (approved):
Thanks!

PR #6485, opened 2024-09-10
feat: allow dash without piped proxy

janisslsm:
This also allows you to watch newly released videos, HLS streams seem to be broken for newly released videos.


Bnyro:
This should be applied at https://github.com/libre-tube/LibreTube/blob/master/app/src/main/java/com/github/libretube/services/OnlinePlayerService.kt#L377 too.

Bnyro:
I don't really get why DASH works with `Local stream extraction` enabled, but doesn't if you use a working private Piped instance (no matter if `Disable Piped Proxy` is enabled or disabled) ...

janisslsm:
> I don't really get why DASH works with `Local stream extraction` enabled, but doesn't if you use a working private Piped instance (no matter if `Disable Piped Proxy` is enabled or disabled) ...

Yeah, I'm not sure either and I haven't really looked into piped's backend enough to figure it out.

Bnyro:
> Yeah, I'm not sure either and I haven't really looked into piped's backend enough to figure it out.
>
I've compared the YouTube stream URLs used for playback (with local stream extraction vs Piped instance) and both work fine when pasting the URL in the browser...

Bnyro (approved):

Issue #6558, opened 2024-09-28
Shuffle button error after deleting video


Notesure:
### Steps to reproduce

1. Delete the last video in the playlist
2. Error when clicking shuffle button

### Expected behavior

It can be annoying if accidentally touch the button. This is what happened at the moment

### Actual behavior

https://github.com/user-attachments/assets/34e9cb07-ff5e-45d7-bb9a-9d2ead0190a7

### LibreTube version

0.25.2 (57)

### Android version

Misc

### Other details

java.util.NoSuchElementException: List is empty.
	at kotlin.collections.CollectionsKt___CollectionsKt.first(_Collections.kt:221)
	at com.github.libretube.ui.fragments.PlaylistFragment$fetchPlaylist$1.invokeSuspend$lambda$7(PlaylistFragment.kt:256)
	at com.github.libretube.ui.fragments.PlaylistFragment$fetchPlaylist$1.$r8$lambda$_hxFgAIhfI0eRI5JOWYxBe6Q4Ck(Unknown Source:0)
	at com.github.libretube.ui.fragments.PlaylistFragment$fetchPlaylist$1$$ExternalSyntheticLambda4.onClick(D8$$SyntheticClass:0)
	at android.view.View.performClick(View.java:7792)
	at android.widget.TextView.performClick(TextView.java:16112)
	at com.google.android.material.button.MaterialButton.performClick(MaterialButton.java:1218)
	at android.view.View.performClickInternal(View.java:7769)
	at android.view.View.access$3800(View.java:910)
	at android.view.View$PerformClick.run(View.java:30218)
	at android.os.Handler.handleCallback(Handler.java:938)
	at android.os.Handler.dispatchMessage(Handler.java:99)
	at android.os.Looper.loopOnce(Looper.java:226)
	at android.os.Looper.loop(Looper.java:313)
	at android.app.ActivityThread.main(ActivityThread.java:8669)
	at java.lang.reflect.Method.invoke(Native Method)
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:571)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1135)
    

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

PR #6561, opened 2024-09-29
fix: shuffle button crash when playlist is empty

Relwi:
It fixes the issue https://github.com/libre-tube/LibreTube/issues/6558


Bnyro (approved):
Thanks!

Issue #6306, opened 2024-07-29
Playlist menu is basically dead


alejandro-alzate:
### Steps to reproduce

1. pick a video
2. disable the auto-play button
3. open playlist selector (between) sponsor block and settings
4. click any video

### Expected behavior

Video should change to selected video

### Actual behavior

Video in playlist gets highlighted then nothing happens

### LibreTube version

0.24.0 - Initial version
0.25.2 - Current version with the bug as editing this

### Android version

LineageOS 20 (android 13)

### Other details

Demo video 0.24.0: https://www.youtube.com/watch?v=lKt68uPxehU
Demo video 0.25.2: https://www.youtube.com/watch?v=qAdS_ftzfOQ
Auto-play fix discovery: https://www.youtube.com/watch?v=g84bLMcEdgA

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

alejandro-alzate:
I noticed that the issue was tagged can't reproduce
so here i attach a video showcasing it: https://www.youtube.com/watch?v=qAdS_ftzfOQ

alejandro-alzate:
I just noticed that turning on the auto play feature fixes it which is weird.
https://www.youtube.com/watch?v=g84bLMcEdgA

alejandro-alzate:
Hello @Bnyro I was curious if I can get the latest apk with this fix applied.
Seems that the github action is still stuck on `0.25.2`

You can use me as cannon meat on the way there. y'know making you aware of any new bugs popping up and such.

akano12:
@alejandro-alzate
https://github.com/libre-tube/NightlyBuilds

alejandro-alzate:
understood, g'night sir.
I'm installing it now.
I'll let you know if I find funny bussiness.

alejandro-alzate:
wh- huh it says that is `0.25.2 (57)`
Also on install it did not kill the pip.

akano12:
> wh- huh it says that is `0.25.2 (57)`
> Also on install it did not kill the pip.

It's a debug build, separate application that contains bug fixes and feature that aren't released in stable build yet

alejandro-alzate:
Understood, just got a bit thrown off.
Also thoughtful touch giving them a different icon hue, otherwise I wouldn't be able to tell them apart.

[...]

PR #6588, opened 2024-10-04
fix: click on video playlist won't load if autoplay is off

Relwi:
Fixes https://github.com/libre-tube/LibreTube/issues/6306

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
+++ b/app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
@@ -970,7 +970,7 @@ class PlayerFragment : Fragment(), OnlinePlayerOptions {
                     (videoStream?.height ?: 0) > (videoStream?.width ?: 0)
 
             PlayingQueue.setOnQueueTapListener { streamItem ->
-                streamItem.url?.toID()?.let { playNextVideo(it) }
+                streamItem.url?.toID()?.let { playNextVideo(it, true) }
             }
 
             // hide the button to skip SponsorBlock segments manually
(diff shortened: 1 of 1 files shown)


Bnyro (approved):
Thanks!

app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
```diff
@@ -1042,13 +1042,13 @@ class PlayerFragment : Fragment(), OnlinePlayerOptions {
...
      */
-    private fun playNextVideo(nextId: String? = null) {
+    private fun playNextVideo(nextId: String? = null, isOnQueueClick: Boolean = false) {
         if (nextId == null && PlayingQueue.repeatMode == Player.REPEAT_MODE_ONE) {
             viewModel.player.seekTo(0)
             return
         }
 
-        if (!PlayerHelper.isAutoPlayEnabled(playlistId != null)) return
+        if (!PlayerHelper.isAutoPlayEnabled(playlistId != null) && !isOnQueueClick) return
```
Bnyro, line 1051: Couldn't we instead do `!PlayerHelper.isAutoPlayEnabled(playlistId != null) && nextId == null`, since in theory autoplay will never call this function with any video id provided (if I remember correctly how I intended to write that code) whereas the nextId is set when tapping a video in the queue?

Bnyro: (same applies for your other PR, I probably merged that too quickly before thinking)

alejandro-alzate:
cool.

app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
Relwi: yep, it worked in that way too, I opened another PR

PR #6593, opened 2024-10-05
refactor: click on video playlist won't load if autoplay is off

Relwi:


Bnyro (approved):
Thanks for the quick refactor!

PR #6599, opened 2024-10-07
feat: ask user to play downloaded version if video already downloaded

### LINKED ISSUE #6325: When playing a downloaded video from Home screen/Subscriptions, offer to play it online or locally

I'd like, when I download a video, to be able to play it locally somewhere else than in the Downloaded tab. So I'd like LibreTube to identify when I try to play a video that is downloaded, and offer me to play it locally.

### Other details

There should also be a good sync of watch position within the video, between the local and the online version.

### Acknowledgements

angolo:
closes #6325

Added a dialog in PlayerFragment that asks the user, when the selected video has already been downloaded, if they want to play it from the download folder.

Added a setting entry to enable the described feature.

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/dialogs/PlayOfflineDialog.kt
+++ b/app/src/main/java/com/github/libretube/ui/dialogs/PlayOfflineDialog.kt
@@ -1,0 +1,55 @@
+package com.github.libretube.ui.dialogs
+
+import android.app.Dialog
+import android.content.Intent
+import android.os.Bundle
+import android.widget.ArrayAdapter
+import androidx.core.os.bundleOf
+import androidx.fragment.app.DialogFragment
+import androidx.fragment.app.setFragmentResult
+import com.github.libretube.R
+import com.github.libretube.constants.IntentData
+import com.github.libretube.databinding.DialogPlayOfflineBinding
+import com.github.libretube.ui.activities.OfflinePlayerActivity
+import com.github.libretube.util.TextUtils
+import com.google.android.material.dialog.MaterialAlertDialogBuilder
+
+class PlayOfflineDialog : DialogFragment() {
+
+    override fun onCreateDialog(savedInstanceState: Bundle?): Dialog {
+        val binding = DialogPlayOfflineBinding.inflate(layoutInflater)
+        val videoId = requireArguments().getString(IntentData.videoId)
+        binding.videoTitle.text = requireArguments().getString(IntentData.videoTitle)
+
+        val downloadInfo = requireArguments().getStringArray(IntentData.downloadInfo)
+        binding.downloadInfo.adapter =
+            ArrayAdapter(requireContext(), android.R.layout.simple_list_item_1, downloadInfo.orEmpty().map {
+                TextUtils.SEPARATOR + it
+            })
+
+        return MaterialAlertDialogBuilder(requireContext())
+            .setTitle(R.string.dialog_play_offline_title)
+            .setView(binding.root)
+            .setPositiveButton(R.string.yes) { _, _ ->
+                val intent = Intent(requireContext(), OfflinePlayerActivity::class.java)
+                    .putExtra(IntentData.videoId, videoId)
+                requireContext().startActivity(intent)
+
+                setFragmentResult(
+                    PLAY_OFFLINE_DIALOG_REQUEST_KEY,
+                    bundleOf(IntentData.isPlayingOffline to true)
+                )
+            }
+            .setNegativeButton(getString(R.string.cancel)) { _, _ ->
+                setFragmentResult(
+                    PLAY_OFFLINE_DIALOG_REQUEST_KEY,
+                    bundleOf(IntentData.isPlayingOffline to false)
+                )
+            }
+            .show()
+    }
+
+    companion object {
+        const val PLAY_OFFLINE_DIALOG_REQUEST_KEY = "play_offline_dialog_request_key"
+    }
+}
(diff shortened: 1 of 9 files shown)


app/src/main/res/xml/player_settings.xml
```diff
@@ -78,6 +79,13 @@
             app:title="@string/fullscreen_orientation"
             app:useSimpleSummaryProvider="true" />
 
+        <SwitchPreferenceCompat
```
Bnyro, line 82: I think we don't need this preference (we already have quite a lot of player settings which makes it very hard to overview for users), it's okay to always ask to play the downloaded version when a downloaded version has been found.

app/src/main/res/layout/dialog_play_offline.xml
```diff
@@ -0,0 +1,31 @@
...
+<!--
+    <com.google.android.material.checkbox.MaterialCheckBox
+        android:id="@+id/remember"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:layout_gravity="center"
+        android:layout_marginVertical="10dp"
+        android:text="@string/remember" />
+
+-->
```
Bnyro, line 30: I think it's fine to always ask and not remember, so I think this can be dropped.

app/src/main/res/layout/dialog_play_offline.xml
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:orientation="vertical"
+    android:padding="20dp">
+
+
+    <!-- todo: video info could be added
```
Bnyro, line 10: Seems like a good idea to display information about the downloaded video here, title, caption language (if any) and resolution / audio quality should be sufficient.

app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
```diff
@@ -437,7 +452,17 @@ class PlayerFragment : Fragment(), OnlinePlayerOptions {
...
+        localSearchResult?.let {
+            PlayOfflineDialog(it, {
+                (requireActivity() as MainActivity).supportFragmentManager.commit {
+                    remove(this@PlayerFragment)
+                }
+            }) { _, _ ->
+                playVideo()
+            }.show(parentFragmentManager, null)
+
+        } ?: playVideo()
```
Bnyro, line 465:
```suggestion
        if (localSearchResult != null) {
            PlayOfflineDialog(it, {
                (requireActivity() as MainActivity).supportFragmentManager.commit {
                    remove(this@PlayerFragment)
                }
            }) { _, _ ->
                playVideo()
            }.show(parentFragmentManager, null)

        } else {
            playVideo()
        }
```
I'd prefer using normal if conditions for readability

app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
```diff
@@ -390,6 +400,11 @@ class PlayerFragment : Fragment(), OnlinePlayerOptions {
         keepQueue = playerData.keepQueue
         timeStamp = playerData.timestamp
 
+        if (PlayerHelper.isPlayFromDownloadEnabled)
```
Bnyro, line 403: Couldn't we do this directly in `onViewCreated` so that we don't need to create the public `localSearchResult` variable? The variable is only used once, so it's enough to have it locally in the function context.

app/src/main/java/com/github/libretube/ui/dialogs/PlayOfflineDialog.kt
```diff
@@ -0,0 +1,44 @@
...
+            .setTitle(getString(R.string.dialog_play_offline_title))
+            .setView(binding.root)
+            .setPositiveButton(getString(R.string.yes)) { _, _ ->
+                //PreferenceHelper.putBoolean("play_offline_remember", binding.remember.isChecked) //TODO: play offline remember isn't enough explicative
+                val intent = Intent(requireContext(), OfflinePlayerActivity::class.java)
+                intent.putExtra(IntentData.videoId, downloadItem.download.videoId)
+                requireContext().startActivity(intent)
+                onConfirmResult()
+            }
+            .setNegativeButton(getString(R.string.cancel), onNegativeResult)
```
Bnyro, line 38:
```suggestion
            .setNegativeButton(R.string.cancel, onNegativeResult)
```

app/src/main/java/com/github/libretube/ui/dialogs/PlayOfflineDialog.kt
```diff
@@ -0,0 +1,44 @@
...
+import com.github.libretube.helpers.PreferenceHelper
+import com.github.libretube.ui.activities.MainActivity
+import com.github.libretube.ui.activities.OfflinePlayerActivity
+import com.google.android.material.dialog.MaterialAlertDialogBuilder
+
+class PlayOfflineDialog(
+    private val downloadItem: DownloadWithItems,
+    private val onConfirmResult: ()->Unit,
+    private val onNegativeResult: DialogInterface.OnClickListener
+) : DialogFragment() {
```
Bnyro, line 22: It's generally better to not pass any data via dialog constructors, as if the activity is recreated the dialog is being recreated without the constructor arguments.

Instead, we should pass the the video information via arguments (`putExtra` and `getExtra`) and use Fragment result listeners as per https://developer.android.com/guide/fragments/communicate#fragment-result.

Bnyro (approved):

Issue #6010, opened 2024-05-09
Back gesture on replies doesn't return to comments


0xrxL:
### Describe your suggested feature

When I open comments list and also open the answers below a comment, I found more comfortable (with device one hand use) to make a gesture to return to the previous page, instead of tapping on the dedicated button.

Unfortunately I can't do that with LibreTube, because a go back gesture close the entire comment page, independently if I'm on the comments list or answers list.

It would be useful to add the possibility to return to all comments page (from the answers one) with the system gesture, and eventually close the entire comments page once the view is focused on the main comments page.

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

DahilerMehvOlur:
Yes It is uncomfortable thing. Because I'm used to using the system's way of going back to go back.

PR #6600, opened 2024-10-07
Added onBack callback to handle back gesture when inside video's comment's replies

angolo:
#6010

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/sheets/CommentsSheet.kt
+++ b/app/src/main/java/com/github/libretube/ui/sheets/CommentsSheet.kt
@@ -1,9 +1,17 @@
 package com.github.libretube.ui.sheets
 
+import android.app.Dialog
+import android.content.Context
+import android.os.Build
 import android.os.Bundle
+import android.util.Log
+import android.view.KeyEvent
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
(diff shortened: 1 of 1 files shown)


Bnyro:
Tested, this doesn't actually work.

Bnyro:
Did you AI-generate this PR?

angolo:
First of all, what does it make it AI Generated?
Second: I'm sorry, I tested it on my work pc with windows on an emulator with Android 34 and it worked, now building the project on my mac it seems it doesn't work

Third: yes, i forgot to remove the Log's import

Bnyro:
> First of all, what does it make it AI Generated?
>
Sorry, apologies, I just made bad experiences with AI generated PRs recently.

> Second: I'm sorry, I tested it on my work pc with windows on an emulator with Android 34 and it worked, now building the project on my mac it seems it doesn't work
>
Have you tested if it works well without these changes on the emulator? If this change actually fixes the issue on some devices, it's still reasonable to merge.

angolo:
I recovered my work pc, the emulator I used runs API 35, running the debugger it seems that on lower api's it is not registered / doesn't override the default onbackpressed behavior

Bnyro:
I have tried debugging that issue and it seemed like the `onBackPressedDispatcher` isn't getting triggered when pressing the system back button in the replies fragment, but I've really no idea why...

PR #6719, opened 2024-11-11
feat: hide crash dialogs when developing

FineFindus:
Hides the crash dialogs when the application is being developed, as it is obvious and the crash log is easily available


Bnyro (approved):
Makes sense, thank you!

PR #6725, opened 2024-11-13
feat: display full video sponsorblock categories

FineFindus:
Displays the full SponsorBlock categories as a chip below the video, similar to how the web version works

Fully sponsored video
Exclusive access video

### DIFF
--- a/app/src/main/java/com/github/libretube/api/PipedApi.kt
+++ b/app/src/main/java/com/github/libretube/api/PipedApi.kt
@@ -43,7 +43,8 @@ interface PipedApi {
     @GET("sponsors/{videoId}")
     suspend fun getSegments(
         @Path("videoId") videoId: String,
-        @Query("category") category: String
+        @Query("category") category: String,
+        @Query("actionType") actionType: String? = null
     ): SegmentData
 
     @GET("dearrow")
--- a/app/src/main/java/com/github/libretube/ui/views/DescriptionLayout.kt
+++ b/app/src/main/java/com/github/libretube/ui/views/DescriptionLayout.kt
@@ -12,10 +12,13 @@ import androidx.core.view.isGone
 import androidx.core.view.isVisible
 import androidx.recyclerview.widget.LinearLayoutManager
 import com.github.libretube.R
+import com.github.libretube.api.obj.Segment
 import com.github.libretube.api.obj.Streams
 import com.github.libretube.databinding.DescriptionLayoutBinding
+import com.github.libretube.enums.SbSkipOptions
 import com.github.libretube.extensions.formatShort
 import com.github.libretube.helpers.ClipboardHelper
+import com.github.libretube.helpers.PlayerHelper
 import com.github.libretube.ui.activities.VideoTagsAdapter
 import com.github.libretube.util.HtmlParser
 import com.github.libretube.util.LinkHandler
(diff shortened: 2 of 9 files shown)


FineFindus:
Maybe it would be better to move the badge into the same line as the title? Having it above leaves a lot of empty space.

Badge inline with video title

app/src/main/java/com/github/libretube/api/PipedApi.kt
```diff
@@ -46,6 +46,13 @@ interface PipedApi {
         @Query("category") category: String
     ): SegmentData
 
+    @GET("sponsors/{videoId}")
+    suspend fun getSegments(
+        @Path("videoId") videoId: String,
+        @Query("category") category: String,
+        @Query("actionType") actionType: String
```
Bnyro, line 53: I'd prefer to make that parameter nullable and reuse the function above

app/src/main/java/com/github/libretube/ui/views/DescriptionLayout.kt
```diff
@@ -40,6 +41,20 @@ class DescriptionLayout(
...
+        segments.filter { it.actionType == Segment.TYPE_FULL }.firstNotNullOfOrNull {
+            when (it.category) {
+                "sponsor" -> context.getString(R.string.category_sponsor)
+                "exclusive_access" -> context.getString(R.string.category_exclusive_access)
+                else -> null
+            }
+        }?.let {
+            binding.playerSponsorBadge.isVisible = true
+            binding.playerSponsorBadge.text = it
+
```
Bnyro, line 54: nit: blank line

Bnyro:
> Maybe it would be better to move the badge into the same line as the title? Having it above leaves a lot of empty space.
>
None of it is a perfect solution, but I'd personally prefer having the badge on a different line than the title.

We can decrease the padding between the video and the title though if the badge is visible, so that it's at least taking a little bit less total space.

app/src/main/java/com/github/libretube/ui/views/DescriptionLayout.kt
```diff
@@ -40,6 +41,19 @@ class DescriptionLayout(
...
+        segments.filter { it.actionType == Segment.TYPE_FULL }.firstNotNullOfOrNull {
+            when (it.category) {
+                "sponsor" -> context.getString(R.string.category_sponsor)
+                "exclusive_access" -> context.getString(R.string.category_exclusive_access)
+                else -> null
+            }
+        }?.let {
+            binding.playerSponsorBadge.isVisible = true
+            binding.playerSponsorBadge.text = it
+        }
```
Bnyro, line 54: We should hide the badge when no matching category is found, otherwise it stays visible even though there are no SponsorBlock segments in the next video.

Bnyro:
Last but not least, please add a preference at https://github.com/libre-tube/LibreTube/blob/master/app/src/main/res/xml/sponsorblock_settings.xml to allow users to toggle displaying the sponsor categories (can be enabled by default).

Apart from the other comment above, lgtm.

app/src/main/java/com/github/libretube/ui/views/DescriptionLayout.kt
FineFindus: I think it should already be hidden, since `android:visibility="gone"` is set?

FineFindus:
I'm not a fan of adding new preferences, could we use the existing sponsor category for it? I think it already covers use case of now wishing to see the sponsor hints.

Bnyro:
> I'm not a fan of adding new preferences
>
I agree with this generally, I'm not too sure about it because it's quite a big chip there above the title that not everybody might be comfortable with.

If we find a better way to display it (e.g. collapsed by default in the description layout), we definitely won't need a new preference.

app/src/main/java/com/github/libretube/ui/views/DescriptionLayout.kt
FineFindus: *sigh* Android...

Bnyro: android-rs when? :eyes:

FineFindus:
> If we find a better way to display it (e.g. collapsed by default in the description layout), we definitely won't need a new preference.

I was thinking about doing that, but I feel like it would defeat the purpose, since a user is unlikely to look in the description, and atleast the sponsor badge is something that should be very visible

Bnyro:
Sorry for the conflicts, I refactored the app to use the [MediaLibraryService](https://developer.android.com/media/media3/session/serve-content) to unify a lot of logic and improve the stability of the background playback.

Bnyro:
`actionType` needs to be passed at https://github.com/libre-tube/LibreTube/blob/master/app/src/main/java/com/github/libretube/services/OnlinePlayerService.kt#L215 now.

Bnyro (approved):
Thanks!

PR #6829, opened 2024-11-26
fix: maximise player when playing new video

FineFindus:
Fixes a regression from 12be4c716cd88fd00f413defd96a7d63292000ed, whereupon trying to play a new video, the player would stay minimized. Additionally, the playing video would be added to the queue, causing the previously played video to be played again.
This is confusing to users, as it breaks the expectation of the player opening when trying to play a video.

Whilst, this causes the player to open, the previously video will continue to play, until the new one is loaded, which might also be unexpected. Maybe it should be paused during the loading time and the spinner be shown?

### DIFF
--- a/app/src/main/java/com/github/libretube/helpers/NavigationHelper.kt
+++ b/app/src/main/java/com/github/libretube/helpers/NavigationHelper.kt
@@ -26,6 +26,7 @@ import com.github.libretube.ui.base.BaseActivity
 import com.github.libretube.ui.fragments.AudioPlayerFragment
 import com.github.libretube.ui.fragments.PlayerFragment
 import com.github.libretube.ui.views.SingleViewTouchableMotionLayout
+import com.github.libretube.util.PlayingQueue
 
 object NavigationHelper {
     private val handler = Handler(Looper.getMainLooper())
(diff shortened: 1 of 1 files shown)


Bnyro (approved):

Bnyro:
> Whilst, this causes the player to open, the previously video will continue to play, until the new one is loaded, which might also be unexpected. Maybe it should be paused during the loading time and the spinner be shown?
>
That's hard to answer, it ultimately depends on the loading time if it's confusing to the user or not. If the next video loads within 300ms, that's probably quite intuitive, but for long loading times (or even API/network errors when loading the video info) it's likely not.

FineFindus:
Will be right back with an algorithm that can predict the future :)

I think the issue right now is that the user clicks on a video and then undetermined amount of time later the action actually takes place, which just feels like the application is lagging. Having at least an indicator that app is doing some work in the background might help solve that.

Bnyro:
> I think the issue right now is that the user clicks on a video and then undetermined amount of time later the action actually takes place, which just feels like the application is lagging. Having at least an indicator that app is doing some work in the background might help solve that.
>
Sure, feel free to make a pull request if you got an idea how to do so (or show the loading indicator as you described previously) :+1:

PR #6870, opened 2024-12-10
feat(PlayerFragment): smooth out predictive back progress

FineFindus:
Sets a limit for the maximum progress of the predictive back gesture in the PlayerFragment. Before reaching the maximum progress, the amount of displayed progress is gradually smooth out, such that further progress appears less.
This makes the back gesture behave more like the one used by the bottom-sheets.
Ref: f143e0d89430b7dd10aff091a52d125694af6e22

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
+++ b/app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
@@ -118,6 +118,7 @@ import kotlinx.coroutines.runBlocking
 import kotlinx.coroutines.withContext
 import kotlin.math.abs
 import kotlin.math.ceil
+import kotlin.math.pow
 
 
 @androidx.annotation.OptIn(androidx.media3.common.util.UnstableApi::class)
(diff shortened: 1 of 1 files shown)


Bnyro:
I'm personally not really convinced by this.

The goal is to support "full" predictive back gestures for the other fragments as well so that the previous fragment is shown when the back progress is `100%`, as seen in some other Jetpack Compose apps.

Unfortunately there doesn't seem to be an inbuild way to do so with the `NavController` APIs of Android that we're using.

FineFindus:
I agree, I've been using this for the last couple of days, and it feels slightly out-of-sync with the OS implementation.

> The goal is to support "full" predictive back gestures for the other fragments as well so that the previous fragment is shown when the back progress is 100%, as seen in some other Jetpack Compose apps.

I'm confused, that's already the case for the player with the current implementation?

Bnyro:
> I'm confused, that's already the case for the player with the current implementation?
>
Yes, it is, but it's not for the other "normal" fragments such as the `ChannelFragment`, `PlaylistFragment`, ...

Issue #4401, opened 2023-08-08
video upload time and views in playlist and bookmark


Eliya-Manoj:
### Describe your suggested feature

Feature that shows video upload time and views in playlist and bookmark. This is very useful for the ongoing playlist.

### Other details

 Almost like this
Before
![6237765975793579777_121](https://github.com/libre-tube/LibreTube/assets/90308888/279d45d8-ed0d-4c94-b3da-2e6d9724a251)
After
![6237765975793579778_121](https://github.com/libre-tube/LibreTube/assets/90308888/582d288e-c4e6-4c1e-8e8c-2196121dd2f0)

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

Bnyro:
Unfortunately, that's not supported by the API.

PR #6920, opened 2025-01-05
feat(Playlist): display views and upload date

FineFindus:
As the Piped API now provides the number of views/upload date for a playlist item, we can display them in the playlist view.

Ref: https://github.com/libre-tube/LibreTube/issues/4401

~~This however leads to the channel being displayed below the views/upload date, which is suboptimal. I'm not sure what is the best way to solve that, either displaying it the info line itself (like for other related streams) or modifying to view to display it above the video info?~~
Since related streams already show channel, views and upload date in this way, I think it is best to continue doing so here. It also works when the information is missing, such as when using created playlists.

| Before                                                                                                                            	| After                                                                                                                 	|
|-----------------------------------------------------------------------------------------------------------------------------------	|-----------------------------------------------------------------------------------------------------------------------	|
| Playlist with items only showing the uploader 	| Playlist with items showing views 	|

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/adapters/PlaylistAdapter.kt
+++ b/app/src/main/java/com/github/libretube/ui/adapters/PlaylistAdapter.kt
@@ -80,12 +80,14 @@ class PlaylistAdapter(
 
         holder.binding.apply {
             videoTitle.text = streamItem.title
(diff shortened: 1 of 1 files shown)


Bnyro (approved):
Awesome, thank you!

PR #6941, opened 2025-01-10
feat(LocalFeedRepository): fetch only enabled channel tabs

FineFindus:
Updates the local feed extraction to only fetch channel tabs that are enabled. This cuts down the amount of requests that are sent, but ultimately not used.

As a side note, in my [fork](https://github.com/FineFindus/LibreTube/blob/70c0ab04cdf926f09d115dc3f0389767e1afd0f0/app/src/main/java/com/github/libretube/api/StreamsExtractor.kt#L119) I've experimented with using the more general `FeedInfo` first, to check if there are even new videos available, which seems to reduce the chance to be rate-limited, so it might be worthwhile to take a look at ;)

### DIFF
--- a/app/src/main/java/com/github/libretube/repo/LocalFeedRepository.kt
+++ b/app/src/main/java/com/github/libretube/repo/LocalFeedRepository.kt
@@ -6,6 +6,7 @@ import com.github.libretube.api.obj.StreamItem
 import com.github.libretube.constants.PreferenceKeys
 import com.github.libretube.db.DatabaseHolder
 import com.github.libretube.db.obj.SubscriptionsFeedItem
+import com.github.libretube.enums.ContentFilter
 import com.github.libretube.extensions.parallelMap
 import com.github.libretube.helpers.NewPipeExtractorInstance
 import com.github.libretube.helpers.PreferenceHelper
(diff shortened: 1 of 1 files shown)


Bnyro (approved):
Thank you!

Your implementation looks promising, I'll see what I can re-use for my implementation :)

My idea was to reduce the chances of being blocked by YouTube by refreshing the feed only once a day or so and cache it in the app's database, but we probably should add more measures to decrease chances of being blocked as much as possible.

FineFindus:
> My idea was to reduce the chances of being blocked by YouTube by refreshing the feed only once a day or so and cache it in the app's database, but we probably should add more measures to decrease chances of being blocked as much as possible.

I've also added some [throttling](https://github.com/FineFindus/LibreTube/commit/6966ff75e882abe84b016e86da16f809d08f9c4b), similar to how it is done in NewPipe, a couple of weeks ago, however I think the values need to be adjusted, since it, I still experience some rate-limiting.
Since we store the feed, we could also use separate tiers, i.e. fetch channels that upload daily more often than once that upload only once a year, but I'm not sure if that wouldn't defeat the point of having the feed :)

Bnyro:
> I've also added some [throttling](https://github.com/FineFindus/LibreTube/commit/6966ff75e882abe84b016e86da16f809d08f9c4b), similar to how it is done in NewPipe, a couple of weeks ago, however I think the values need to be adjusted, since it, I still experience some rate-limiting.
>
True, there's certainly a lot that could be done there, some throttling is probably a good idea too, feel free to open a PR for that :+1:

> Since we store the feed, we could also use separate tiers, i.e. fetch channels that upload daily more often than once that upload only once a year, but I'm not sure if that wouldn't defeat the point of having the feed :)
>
Hmm, but we probably can't really automate guessing how often a channel uploads videos reliably, so users would need to set that manually :thinking:

The best thing to avoid rate limits is probably to not fetch the whole feed at once but in small parts instead, but I'm not too sure how we could really implement that with a decent UX. Users probably want their feed to show all new videos when they refresh it, so fetching only like 10% of the channels subscribed to each day (so users always have a feed that's 10 days old for some channels that haven't been recently refreshed in this example) is also not a perfect solution.

PR #6949, opened 2025-01-11
feat(LocalFeedRepository): throttle feed extraction

FineFindus:
Throttles the local feed extration in-order to avoid rate-limiting. This is done, similary to NewPipe, by introducting an articificial delay every 50 fetched feeds of on average 1 second. These values may have to be changed in the future.

Ref: https://github.com/libre-tube/LibreTube/pull/6941#issuecomment-2585206589

### DIFF
--- a/app/src/main/java/com/github/libretube/repo/LocalFeedRepository.kt
+++ b/app/src/main/java/com/github/libretube/repo/LocalFeedRepository.kt
@@ -13,6 +13,7 @@ import com.github.libretube.extensions.toID
 import com.github.libretube.helpers.NewPipeExtractorInstance
 import com.github.libretube.helpers.PreferenceHelper
 import com.github.libretube.ui.dialogs.ShareDialog.Companion.YOUTUBE_FRONTEND_URL
+import kotlinx.coroutines.delay
 import org.schabi.newpipe.extractor.channel.ChannelInfo
 import org.schabi.newpipe.extractor.channel.tabs.ChannelTabInfo
 import org.schabi.newpipe.extractor.channel.tabs.ChannelTabs
@@ -20,6 +21,7 @@ import org.schabi.newpipe.extractor.feed.FeedInfo
 import org.schabi.newpipe.extractor.stream.StreamInfoItem
 import java.time.Duration
 import java.time.Instant
+import java.util.concurrent.atomic.AtomicInteger
 
 class LocalFeedRepository : FeedRepository {
     private val relevantTabs =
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/github/libretube/repo/LocalFeedRepository.kt
```diff
@@ -59,6 +61,7 @@ class LocalFeedRepository : FeedRepository {
     }
 
     private suspend fun refreshFeed(channelIds: List<String>, minimumDateMillis: Long) {
+        val extractionCount = AtomicInteger()
```
Bnyro, line 64: We don't actually need an atomic integer here, the outer for loop runs only once at a time.

app/src/main/java/com/github/libretube/repo/LocalFeedRepository.kt
```diff
@@ -70,6 +73,12 @@ class LocalFeedRepository : FeedRepository {
             }.filterNotNull().flatten().map(StreamItem::toFeedItem)
 
             DatabaseHolder.Database.feedDao().insertAll(collectedFeedItems)
+
+            // throttle feed extraction to avoid rate limiting
+            val count = extractionCount.getAndIncrement()
```
Bnyro, line 78: We're fetching `CHUNK_SIZE` feeds in parallel above. Thus the counter would probably need to be incremented by `CHUNK_SIZE` instead of 1, right?

Or we move the counter incrementation to the try catch statement above, so that it's executed for each channel we fetch.

app/src/main/java/com/github/libretube/repo/LocalFeedRepository.kt
```diff
@@ -59,8 +61,15 @@ class LocalFeedRepository : FeedRepository {
     }
 
     private suspend fun refreshFeed(channelIds: List<String>, minimumDateMillis: Long) {
+        val extractionCount = AtomicInteger()
         for (channelIdChunk in channelIds.chunked(CHUNK_SIZE)) {
             val collectedFeedItems = channelIdChunk.parallelMap { channelId ->
+                // throttle feed extraction to avoid rate limiting
+                val count = extractionCount.getAndIncrement()
+                if (count != 0 && count % BATCH_SIZE == 0) {
```
Bnyro, line 69: This will likely only let one thread sleep while the other parallel ones are still working as usual. It might make sense to move this if condition to the end of the for loop, and instead of checking if `count % BATCH_SIZE == 0` we check if `count >= BATCH_SIZE` and reset it to zero when it's the case? :)

Bnyro (approved):
Thank you!

PR #6954, opened 2025-01-14
Fix back navigation from comment replies

Pittvandewitt:
Fixes https://github.com/libre-tube/LibreTube/issues/6010


Bnyro (approved):
Works perfect, thank you!

PR #6955, opened 2025-01-15
Observe livedata once

Pittvandewitt:
The livedata observe method takes a lifecycleowner as a parameter. This takes care of the registering and unregistering, so we don't have to do this in onResume and onPause. Added benefit is that the loading indicator is not displayed every time you return to the home page.


Bnyro (approved):

Bnyro:
Thank you!

PR #6971, opened 2025-01-16
Rework RecyclerViews to set adapter once

Pittvandewitt:
On most pages, the recyclerview adapter is set on every refresh. By doing this, you basically loose all benefits of a recyclerview. Examples are animations and efficiency in adding items.
By refreshing the subscriptions page, the list will no longer flicker as items that are still the same will no longer be updated.

There are still one or two cases left, but the fragments they live in need some changes outside the scope of this PR before they can be fixed.

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/activities/VideoTagsAdapter.kt
+++ b/app/src/main/java/com/github/libretube/ui/activities/VideoTagsAdapter.kt
@@ -2,21 +2,29 @@ package com.github.libretube.ui.activities
 
 import android.view.LayoutInflater
 import android.view.ViewGroup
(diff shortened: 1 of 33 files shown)


Pittvandewitt:
Are you planning on merging this PR? Because I'd rather focus on moving API and database calls to the viewmodel, so that work is performed in the right lifecycle scope.

Bnyro:
> Are you planning on merging this PR? Because I'd rather focus on moving API and database calls to the viewmodel, so that work is performed in the right lifecycle scope.
>
Yes, sure I am. However I want to merge this after the next release, because this PR also requires some deeper looking-into and testing to prevent bugs.

PR #6975, opened 2025-01-16
Rework view binding

Pittvandewitt:
This PR reworks viewbinding in fragments. 
- The layout is now set in the constructor, making the code a little more concise. 
- val binding = binding entries are now removed
- Global viewbinding references are converted to local where possible


Bnyro (approved):
That's a nice shorthand constructor I wasn't aware of, thank you!

Issue #6970, opened 2025-01-16
Clicking to home button when opening sections on the "Home" page does not respond.


Figim:
### Steps to reproduce

1. Enter Featured, Playlist, Trending sections
2. Click on the "Home" button.
3. Does not return to home (main page) section

### Expected behavior

Return to the main state with the home button when accessing section on the homepage

### Actual behavior

https://github.com/user-attachments/assets/c8ecd765-3b74-4a0e-8ce9-a0182b1d5a60

### LibreTube version

0.26.0 (58)

### Android version

Android

### Other details
Similar bug:
1. Search for something
2. Then click on another section.
3. Then click to home button
4. Home button not selected. No indicator

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [x] I have written a short but informative title.
- [x] I will fill out all of the requested information in this form.

Bnyro:
Probably caused by https://github.com/libre-tube/LibreTube/pull/6956, cc @Pittvandewitt 

I thought I tested this PR and everything worked well though, idk

PR #6976, opened 2025-01-16
Fix navigating back home

Pittvandewitt:
Fixes https://github.com/libre-tube/LibreTube/issues/6970

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/fragments/HomeFragment.kt
+++ b/app/src/main/java/com/github/libretube/ui/fragments/HomeFragment.kt
@@ -35,6 +35,7 @@ import com.github.libretube.ui.models.HomeViewModel
 import com.github.libretube.ui.models.SubscriptionsViewModel
 import com.google.android.material.snackbar.Snackbar
 
+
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/github/libretube/ui/fragments/HomeFragment.kt
```diff
@@ -64,23 +65,23 @@ class HomeFragment : Fragment() {
         }
 
         binding.featuredTV.setOnClickListener {
-            findNavController().navigate(R.id.subscriptionsFragment)
+            findNavController().navigate(R.id.action_homeFragment_to_libraryFragment)
```
Bnyro, line 68: This one should actually be `R.id.action_homeFragment_to_subscriptionsFragment` (which doesn't yet exist) to preserve the previous navigation behavior.

Pittvandewitt: Fixed

Bnyro (approved):
Thank you!

PR #6996, opened 2025-01-22
refactor: refactor WelcomeActivity and associated logic

Pittvandewitt:
- Perform Database and Network requests in the ViewModel to make it survive configuration changes
- InstanceHelper is now a class and not an object to make it eligible for GC
- IO work is ensured to be performed in the IO dispatcher by defining it at function level
- Next button is enabled with enabled=true instead of setting alpha
- WelcomeActivity UI state now survives process death
- Preparations are made to not use context in the ViewModel (No AndroidViewModel)

As a side question, do you appreciate these kind of PRs or is it just distracting you from the vision you have with the project? If yes, I'll continue making more improvements across different parts of the codebase.


Bnyro:
> As a side question, do you appreciate these kind of PRs or is it just distracting you from the vision you have with the project? If yes, I'll continue making more improvements across different parts of the codebase.
>
I really appreciate such PRs honestly, the code base is quite old and partially uses some older libraries and components (e.g. `com.android` instead of `androidx`). I've recently been refactoring a lot of the internal code as well, e.g. the `MediaService` logic or the handling of different subscription and feed extraction methods to make it more robust. Most of my refactors are not UI-related though because I'm not really working with Android UI that much at all and thus lack knowledge about the current state of new `androidx` ways of doing things and other improvements in the Android ecosystem.

Currently I'm however looking into creating a new release soon and thus work on fixing the most critical bugs before making a new release, because a new release is necessary due to YouTube changes.

Therefore I already apologize now for the delay on reviewing requests, I'm mostly spending the time I work on LibreTube to prepare everything for a new release now.

Pittvandewitt:
Good to know! I'll happily make a PR from time to time refactoring a few components to better separate the logic, in similar fashion as this PR.

I agree releasing the update is of much higher priority right now, especially with the iOS player response error.

app/src/main/java/com/github/libretube/ui/models/WelcomeViewModel.kt
```diff
@@ -0,0 +1,110 @@
...
+) : ViewModel() {
+
+    private val _uiState = savedStateHandle.getStateFlow("uiState", UiState())
+    val uiState = _uiState.asLiveData()
+
+    init {
+        viewModelScope.launch {
+            instanceRepository.getInstances()
+                .onSuccess { instances ->
+                    savedStateHandle["uiState"] = _uiState.value.copy(instances = instances)
```
Bnyro, line 38: `"uiState"` could probably be moved out into a constant?

app/src/main/java/com/github/libretube/ui/adapters/callbacks/DiffUtilCallback.kt
Bnyro, line 1: It probably makes sense to add this in a different PR and refactor all the existing `ListAdapter` usages to use it as well?

Pittvandewitt: Agreed. I'll move this change to another PR.

Bnyro (changes requested):

Bnyro:
I've added a comment about one small regression, apart from that good to merge from my side :+1:

PR #7008, opened 2025-01-23
fix(SubscriptionsFragment): skip all-caught-up check for upcoming videos

FineFindus:
Skips the check if a video has been already seen (caught up on) for videos which have not been released yet, such as premieres. Doing the check could lead to the all caught up continously displayed for upcoming videos.


Bnyro (approved):
Noticed that too, thank you!

Issue #6991, opened 2025-01-21
After going to the search section, when you click another navigation button below and return to the current page, there is no "select icon"


Figim:
### Steps to reproduce

**Example**:
1. Go to the search section
2. Click on the trend button
3. Click on the home button
4. "Trend button" selected as icon instead of home button

### Example: Expected behavior

The home button must be selected

### Actual behavior

https://github.com/user-attachments/assets/ccf7a8c9-b613-4702-af91-8df73bbc9c37

### LibreTube version

0.26.8 (58)

### Android version

Android 12

### Other details

No

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [x] I have written a short but informative title.
- [x] I will fill out all of the requested information in this form.

Bnyro:
Hmm, maybe @Pittvandewitt has an idea what's causing that?

PR #7010, opened 2025-01-23
Fix backstack when opening search fragment

Pittvandewitt:
Fixes https://github.com/libre-tube/LibreTube/issues/6991


Bnyro (approved):
I changed `restoreState` to `false`, because it otherwise restored the previous search query after returning home, which is not really intuitive.

Thanks!

PR #7018, opened 2025-01-24
chore: add an icon for the development IDEA

Pittvandewitt:
It appears when selecting the project in Android Studio and in the upper left corner next to the project name. The corners are rounded to match it with the default generated icons. Square or round looks bad imo.


Bnyro (approved):
Thank you!

Issue #7025, opened 2025-01-25
Without the 'Home' item from the navigational bar, clicking any of its items causes the app to partially exit.


abanoy:
### Steps to reproduce

1. Open up the app.
2. Click the three vertical dots on the top right, then 'Settings,' 'Appearance,' and then 'Order' from the 'Navigational Bar' tab.
3. De-select 'Home,' if not already, and then select any options that aren't the 'Home' option.
4. Restart the app.
5. Click any of the items from the navigational bar below.

### Expected behavior

No such interruption would be made when clicking that navigational bar item in question.

### Actual behavior

Without the home tab on the navigational bar, clicking any other items within that section would cause the app to partially exit (Not a complete exit as if the app crashed).

Clicking back the app again would now load that section in question and would not partially exit on any other tabs as there's at least one selected.

However, this bug seems to be nullified if you initially do any word search, load the results, and then click one of the navigational bar items, despite the home item being absent.

### LibreTube version

0.27.0

### Android version

14

### Other details

Before 0.27.0, such as 0.25.2 and 0.26.0, this behavior wasn't occurring when there was no 'Home' item within that navigational bar.

Also, the temporary workaround to this bug in 0.27.0 is to simply include the home item.

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [x] I have written a short but informative title.
- [x] I will fill out all of the requested information in this form.

Bnyro:
Cc @Pittvandewitt

PR #7046, opened 2025-01-29
fix: navigation without home tab enabled

Pittvandewitt:
Fixes https://github.com/libre-tube/LibreTube/issues/7025


Bnyro (approved):
Thanks!

Issue #7039, opened 2025-01-29
Let's Encrypt certificate not valid for Android 7 and older


Pittvandewitt:
### Steps to reproduce

LibreTube is unable to use API calls on Android 7 or older because https://piped-instances.kavin.rocks/ and more domains use Let's Encrypt for their SSL certificates. The error thrown is javax.net.ssl.SSLHandshakeException: java.security.cert.CertPathValidatorException: Trust anchor for certification path not found.

For context: https://community.letsencrypt.org/t/support-for-android-7-and-older-from-oct-2024/216446

There are 2 solutions I can think of:
1. Include the Let’s Encrypt root in LibreTube
2. Bump the minSdk to 26 (Oreo)

Which one do you prefer? I'll happily make a PR for either solution.

### Expected behavior

Content would load

### Actual behavior

Content doesn't load

### LibreTube version

Any

### Android version

Android <= 7.1.1

### Other details

Issue is occurring since Oct 2024

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [x] I have written a short but informative title.
- [x] I will fill out all of the requested information in this form.

Pittvandewitt:
Here is how the certificate can be added: https://stackoverflow.com/a/78309587
Via xml on Android 7 or the clunky way on older Android versions.

The fact that nobody complained about this makes me wonder, why not use minSdk 26?

FineFindus:
I think we should increase minSdk regardless, considering the app is broken on older Android version (cf. https://github.com/libre-tube/LibreTube/issues/7032)

Pittvandewitt:
> the app is broken on older Android version

LibreTube includes a phonetically spelled english translation. The character set used is not available on older versions.
Either removing the translation or providing a better alternative (values-en) will fix those issues.

Applying fixes to better support these older Android versions is not that hard, but in the end it's not my call to make. The idea of not having to worry about (close to) obsolete devices is also appealing.

Bnyro:
> I think we should increase minSdk regardless, considering the app is broken on older Android version (cf. [#7032](https://github.com/libre-tube/LibreTube/issues/7032), [#6986](https://github.com/libre-tube/LibreTube/discussions/6986))
>
I agree on bumping minSdk, there have been multiple issues with the app on old Android versions and I don't think it's worth keeping them supported.

Max20720:
[...]

PR #7047, opened 2025-01-29
BREAKING CHANGE: bump minSdk to 26

Pittvandewitt:
Based on https://github.com/libre-tube/LibreTube/issues/7039 and https://github.com/libre-tube/LibreTube/discussions/6986 I took the freedom to bump the minSdk to 26

### DIFF
--- a/app/src/main/java/com/github/libretube/helpers/LocaleHelper.kt
+++ b/app/src/main/java/com/github/libretube/helpers/LocaleHelper.kt
@@ -2,7 +2,6 @@ package com.github.libretube.helpers
 
 import android.content.Context
 import android.content.res.Configuration
-import android.os.Build
 import android.telephony.TelephonyManager
 import androidx.core.content.getSystemService
 import androidx.core.os.ConfigurationCompat
@@ -29,7 +28,7 @@ object LocaleHelper {
 
     fun updateLanguage(context: Context) {
         val locale = getAppLocale()
-        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) updateResources(context, locale)
+        updateResources(context, locale)
         updateResourcesLegacy(context, locale)
     }
 
--- a/app/src/main/java/com/github/libretube/helpers/LocaleHelper.kt
+++ b/app/src/main/java/com/github/libretube/helpers/LocaleHelper.kt
@@ -29,7 +29,6 @@ object LocaleHelper {
     fun updateLanguage(context: Context) {
         val locale = getAppLocale()
         updateResources(context, locale)
-        updateResourcesLegacy(context, locale)
     }
 
     private fun updateResources(context: Context, locale: Locale) {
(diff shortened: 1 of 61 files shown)


app/src/main/java/com/github/libretube/helpers/LocaleHelper.kt
```diff
@@ -29,7 +28,7 @@ object LocaleHelper {
 
     fun updateLanguage(context: Context) {
         val locale = getAppLocale()
-        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) updateResources(context, locale)
```
Pittvandewitt, line 32: Wasn't this meant to be an if else check, or is it intentional to always call updateResourcesLegacy?

app/src/main/java/com/github/libretube/util/TextUtils.kt
```diff
@@ -111,30 +110,21 @@ object TextUtils {
...
-                } else {
-                    RelativeDateTimeFormatter.RelativeUnit.MONTHS to months
-                }
-                RelativeDateTimeFormatter.getInstance()
-                    .format(time.toDouble(), RelativeDateTimeFormatter.Direction.LAST, timeFormat)
+            val (timeFormat, time) = if (years > 0) {
+                RelativeDateTimeFormatter.RelativeUnit.YEARS to years
             } else {
-                val (timeAgoRes, time) = if (years > 0) {
-                    R.plurals.years_ago to years
```
Pittvandewitt, line 132: This plural is no longer used. It could be removed from the translations.

app/src/main/java/com/github/libretube/util/TextUtils.kt
```diff
@@ -111,30 +110,21 @@ object TextUtils {
...
-                } else {
-                    R.plurals.months_ago to months
```
Pittvandewitt, line 134: Same for this plural

Bnyro: I agree, that's reasonable.

They're probably not yet translated in all languages, so it makes sense to prevent translators for spending time with translating them.

app/src/main/java/com/github/libretube/helpers/LocaleHelper.kt
Bnyro: I think so too, `updateResourcesLegacy` can probably be removed now.

Bnyro:
Okay, I just noticed that after this PR, the language in the main activity can no longer be changed (settings -> language), it only appears to be different in all other activities.

So we might need to re-add `updateResourcesLegacy` for now, even though that's probably not the right solution to prevent that bug.

Bnyro:
Fixed with https://github.com/libre-tube/LibreTube/pull/7078, the previous code for updating the resources was just completely wrong because it created a new configuration context that wasn't actually attached to the activity ...

PR #7052, opened 2025-01-30
refactor: upgrade to Coil3

Pittvandewitt:


app/src/main/java/com/github/libretube/helpers/ImageHelper.kt
```diff
@@ -110,14 +116,15 @@ object ImageHelper {
             .data(url)
             .build()
 
-        return imageLoader.execute(request).drawable?.toBitmapOrNull()
+        imageLoader.execute(request)
+        return imageLoader.execute(request).image?.asDrawable(context.resources)?.toBitmapOrNull()
```
Bnyro, line 120:
```suggestion
        return imageLoader.execute(request).image?.asDrawable(context.resources)?.toBitmapOrNull()
```
typo I guess?

Bnyro, line 120: Couldn't we call `imageLoader.execute(request).image?.toBitmap()` directly here instead of converting to drawable first btw?

Bnyro (approved):
Thank you!

PR #7053, opened 2025-01-30
Use default animations

Pittvandewitt:
This PR reverts to the default animations now that the NavHost is using them. This results in somewhat better predictive back navigation. It doesn't yet use the material3 recommended animation patterns, but that is for a future commit. Before those animations can be implemented, a lot needs to be changed to make sure all data is loaded before the animation starts. (Using `Fragment.postponeEnterTransition` and `Fragment.startPostponedEnterTransition`)


Bnyro:
I'm on Android 15, and with these changes the predictive back animations between fragments are completely gone on my device. Activity animations still work, so it's certainly not a general predictive back issue but rather specific to this PR.

I don't have a clue why things don't work the same as in the screen recording you shared on my end :man_shrugging:

Pittvandewitt:
I ran the code changes on real devices running Android 15 and 16, both working as expected. It also works using an emulator. Maybe Android Studio is using an old cache in your case? Attached an apk built from this branch to double check
[app-arm64-v8a-debug.apk.zip](https://github.com/user-attachments/files/18725503/app-arm64-v8a-debug.apk.zip)

Bnyro:
> I ran the code changes on real devices running Android 15 and 16, both working as expected. It also works using an emulator. Maybe Android Studio is using an old cache in your case? Attached an apk built from this branch to double check [app-arm64-v8a-debug.apk.zip](https://github.com/user-attachments/files/18725503/app-arm64-v8a-debug.apk.zip)
>
Yes, even with a clean install, I tested the apk you provided.

@FineFindus if you have time and motiviation, could you please test this as well on a real device that supports predictive back gestures? I'm asking because I remember that you've been looking into the predictive player back animation recently and thus likely have a device to test it.

FineFindus:
Tested with the attached apk, works fine for me (though I noticed that the Watch History/Downloads pages no longer have any back animation).

Pittvandewitt:
> Yes, even with a clean install, I tested the apk you provided.

Any chance you have set animation scale to off in your developer settings?

> (though I noticed that the Watch History/Downloads pages no longer have any back animation)

Animations added

PR #7055, opened 2025-01-30
fix(MarkableTimeBar): align segments to timebar

FineFindus:
The segments drawn on the timeline seem to be off by one pixel. I'm not quite sure what's causing this, as the calculation is done differently to the timeline itself.
Since the calculation depends on the device, and I haven't been able to test this on multiple devices, further testing should be done before merging.

Misaligned segment


Bnyro:
IMG_20250203_181108

That seems to be off for 1 or 2 pixels on my device :/

Bnyro:
> Was it aligned correctly before?
>
It was for me, yes.

>  I guess this means we need to follow the height calculation more closely
>
Yes, there's probably some error in the general way we try to calculate the height with, probably rebuilding that calculation from scratch makes sense.

Bnyro (approved):
That does the job, thank you!

PR #7061, opened 2025-02-01
fix(player): save watch position on playback end

FineFindus:
This fixes an issue, where the watch position would not correctly be saved as the fully watched video. This could occur when the last part of a video was skipped by a SponsorBlock segment.

### DIFF
--- a/app/src/main/java/com/github/libretube/services/AbstractPlayerService.kt
+++ b/app/src/main/java/com/github/libretube/services/AbstractPlayerService.kt
@@ -94,8 +94,13 @@ abstract class AbstractPlayerService : MediaLibraryService(), MediaLibrarySessio
         override fun onPlaybackStateChanged(playbackState: Int) {
             super.onPlaybackStateChanged(playbackState)
 
-            if (playbackState == Player.STATE_READY) {
-                isTransitioning = false
+            when (playbackState) {
+                Player.STATE_ENDED -> {
+                    exoPlayer?.let { PlayerHelper.saveWatchPosition(it, videoId) }
+                }
+                Player.STATE_READY -> {
+                 isTransitioning = false
+                }
             }
         }
     }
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
```diff
@@ -274,6 +274,7 @@ class PlayerFragment : Fragment(R.layout.fragment_player), OnlinePlayerOptions {
 
             // check if video has ended, next video is available and autoplay is enabled/the video is part of a played playlist.
             if (playbackState == Player.STATE_ENDED) {
+                binding.player.player?.let { PlayerHelper.saveWatchPosition(it, videoId) }
```
Bnyro, line 277: This should be done in `AbstractPlayerService.kt`.

app/src/main/java/com/github/libretube/services/AbstractPlayerService.kt
```diff
@@ -95,6 +95,7 @@ abstract class AbstractPlayerService : MediaLibraryService(), MediaLibrarySessio
             super.onPlaybackStateChanged(playbackState)
 
             if (playbackState == Player.STATE_READY) {
+                exoPlayer?.let { PlayerHelper.saveWatchPosition(it, videoId) }
```
Bnyro, line 98: Rethinking it, shouldn't this be in `Player.STATE_IDLE`? `STATE_READY` is triggered when a new, other video starts playing.

FineFindus: No, I just misread `STATE_READY` as `STATE_ENDED` :)

Bnyro: Oh right, it's been called `STATE_ENDED`, not `STATE_ENDED` :)

Bnyro (approved):
Thank you!

Issue #7065, opened 2025-02-03
High priority: Support PR like NewPipe to use Extractor PoToken


Figim:
### Describe your suggested feature

**NewPipe Extractor:****:
https://github.com/TeamNewPipe/NewPipeExtractor/pull/1272

** I know about this PR, but You need to make a PR for it to work on LibreTube**** **Of course, after the PR merger**

**The PR in the NewPipe application**: You should prepare such PR. Information is given there. Thus, bot http 403 etc. Playback errors are fixed (### not yet merged) 

**PR**: https://github.com/TeamNewPipe/NewPipe/pull/11955

### Other details

Thanks for LibreTube 😊

### Related bugs

1. https://github.com/libre-tube/LibreTube/issues/6389

2. https://github.com/libre-tube/LibreTube/discussions/6898

3. This issue has been fixed again: https://github.com/libre-tube/LibreTube/issues/7012

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [x] I have written a short but informative title.
- [x] I will fill out all of the requested information in this form.

PR #7069, opened 2025-02-03
feat(StreamsExtractor): generate PoToken

FineFindus:
Implements support for locally generating PoTokens using the device web view. This is a direct port of
https://github.com/TeamNewPipe/NewPipe/pull/11955 to native Kotlin.
However, there are some notable differences:
 - NewPipe use an additional `PoTokenGenerator` interface. As there is only one implementation right now, this is skipped
 - NewPipe handles broken web views on older devices. Since we're increasing `minSdk` in https://github.com/libre-tube/LibreTube/pull/7047, I'm hoping we can get away without

~~It is marked as a draft, since while it does generate valid PoTokens, playback does not work (`Source error: open failed: ENOENT`; help is appreciated).~~

Closes: https://github.com/libre-tube/LibreTube/issues/7065

### DIFF
--- a/app/src/main/java/com/github/libretube/api/poToken/JavaScriptUtil.kt
+++ b/app/src/main/java/com/github/libretube/api/poToken/JavaScriptUtil.kt
@@ -1,0 +1,133 @@
+package com.github.libretube.api.poToken
+
+import okio.ByteString.Companion.decodeBase64
+import okio.ByteString.Companion.toByteString
+
+import kotlinx.serialization.json.Json
+import kotlinx.serialization.json.JsonNull
+import kotlinx.serialization.json.JsonObject
+import kotlinx.serialization.json.JsonPrimitive
+import kotlinx.serialization.json.jsonArray
+import kotlinx.serialization.json.jsonPrimitive
+import kotlinx.serialization.json.long
+
+/**
+ * Parses the raw challenge data obtained from the Create endpoint and returns an object that can be
+ * embedded in a JavaScript snippet.
+ */
+fun parseChallengeData(rawChallengeData: String): String {
+    val scrambled = Json.parseToJsonElement(rawChallengeData).jsonArray
+
+    val challengeData = if (scrambled.size > 1 && scrambled[1].jsonPrimitive.isString) {
+        val descrambled = descramble(scrambled[1].jsonPrimitive.content)
+        Json.parseToJsonElement(descrambled).jsonArray
+    } else {
+        scrambled[0].jsonArray
+    }
+
+    val messageId = challengeData[0].jsonPrimitive.content
+    val interpreterHash = challengeData[3].jsonPrimitive.content
+    val program = challengeData[4].jsonPrimitive.content
+    val globalName = challengeData[5].jsonPrimitive.content
+    val clientExperimentsStateBlob = challengeData[7].jsonPrimitive.content
+
+
+    val privateDoNotAccessOrElseSafeScriptWrappedValue = challengeData[1]
+        .takeIf { it !is JsonNull }
+        ?.jsonArray
+        ?.find { it.jsonPrimitive.isString }
+
+    val privateDoNotAccessOrElseTrustedResourceUrlWrappedValue = challengeData[2]
+        .takeIf { it !is JsonNull }
+        ?.jsonArray
+        ?.find { it.jsonPrimitive.isString }
+
+
+    return Json.encodeToString(
+        JsonObject.serializer(), JsonObject(
+            mapOf(
+                "messageId" to JsonPrimitive(messageId),
+                "interpreterJavascript" to JsonObject(
+                    mapOf(
+                        "privateDoNotAccessOrElseSafeScriptWrappedValue" to (privateDoNotAccessOrElseSafeScriptWrappedValue
+                            ?: JsonNull),
+                        "privateDoNotAccessOrElseTrustedResourceUrlWrappedValue" to (privateDoNotAccessOrElseTrustedResourceUrlWrappedValue
+                            ?: JsonNull)
+                    )
+                ),
+                "interpreterHash" to JsonPrimitive(interpreterHash),
+                "program" to JsonPrimitive(program),
+                "globalName" to JsonPrimitive(globalName),
+                "clientExperimentsStateBlob" to JsonPrimitive(clientExperimentsStateBlob)
+            )
+        )
+    )
+}
+
+/**
+ * Parses the raw integrity token data obtained from the GenerateIT endpoint to a JavaScript
+ * `Uint8Array` that can be embedded directly in JavaScript code, and an [Int] representing the
+ * duration of this token in seconds.
+ */
+fun parseIntegrityTokenData(rawIntegrityTokenData: String): Pair<String, Long> {
+    val integrityTokenData = Json.parseToJsonElement(rawIntegrityTokenData).jsonArray
+    return base64ToU8(integrityTokenData[0].jsonPrimitive.content) to integrityTokenData[1].jsonPrimitive.long
+}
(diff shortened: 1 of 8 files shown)


FineFindus:
Please don't add unnecessary comments. Linking an issue to be closed is fine, but the rest just adds noise. I'm already well aware of the NewPipe PR (see description above).

Bnyro:
The crash log is a temporary DNS issue on your side, that's not LibreTube related.

FineFindus:
A few questions: 
- Since HLS does not work with potokens either, should it be removed/disabled?
- JavaScriptUtil.kt uses `nanojson`, should it be ported as well? I left it for now, as it would make it easier to follow future changes.

YT-Advanced:
> - Since HLS does not work with potokens either, should it be removed/disabled?

This issues seem to be NewPipeExtractor bug itself, since HLS don't need `pot` but it still be appended in streaming url

> - JavaScriptUtil.kt uses `nanojson`, should it be ported as well? I left it for now, as it would make it easier to follow future changes.

Uhm u should port it also

Bnyro:
> Since HLS does not work with potokens either, should it be removed/disabled?
>
It shouldn't be completely removed because it might still work with some private Piped instances, however disabling it by default is possible.

> JavaScriptUtil.kt uses nanojson, should it be ported as well? I left it for now, as it would make it easier to follow future changes. 
>
Yes, I would like to keep the used libraries consistent across the code. Since the NanoJSON code part is only very small and the code differs already due to the fact that it's ported to Kotlin, I think it's reasonable to port it.

I don't have much time to look into the problems you mentioned at the moment (and the next few weeks) unfortunately due to tons of exams at university. I'll see if I can find some time to allocate for reviewing soon, but no guarantees.

FineFindus:
Disabled and ported.

> I don't have much time to look into the problems you mentioned at the moment (and the next few weeks) unfortunately due to tons of exams at university. I'll see if I can find some time to allocate for reviewing soon, but no guarantees.

Sure, no pressure. I would also like to clean up the commits before merging. Good luck with your exams.

FineFindus:
>  Currently if something in parseChallengeData or parseIntegrityTokenData fails the app crashes.

Please attach a crash log/reproducible case and don't post vague comments about 'something' crashing.

jellyfish404:
Yo still found a solution for PO token?

app/src/main/java/com/github/libretube/api/poToken/JavaScriptUtil.kt
```diff
@@ -0,0 +1,133 @@
...
+ * and converts it to the specific base64 representation for poTokens.
+ */
+fun u8ToBase64(poToken: String): String {
+    return poToken.split(",")
+        .map { it.toUByte().toByte() }
+        .toByteArray()
+        .toByteString()
+        .base64()
+        .replace("+", "-")
+        .replace("/", "_")
```
Bnyro, line 97: Since that method is partially reverting `base64ToByteString` , I'm wondering why we don't need to replace `=` with `.` as well here.

But since it works this way, the current implementation is right, it's just strange imo

app/src/main/java/com/github/libretube/api/poToken/JavaScriptUtil.kt
```diff
@@ -0,0 +1,133 @@
...
+import kotlinx.serialization.json.long
+
+/**
+ * Parses the raw challenge data obtained from the Create endpoint and returns an object that can be
+ * embedded in a JavaScript snippet.
+ */
+fun parseChallengeData(rawChallengeData: String): String {
+    val scrambled = Json.parseToJsonElement(rawChallengeData).jsonArray
+
+    val challengeData = if (scrambled.size > 1 && scrambled[1].jsonPrimitive.isString) {
```
Bnyro, line 21: If `scrambled.size > 1` is false, the `else` branch will crash with an `IndexOutOfBoundsException`. So we should either just remove that unnecessary size check or do proper size checking and error handling/throwing before the `if else` condition here.

app/src/main/java/com/github/libretube/api/poToken/PoTokenWebView.kt
```diff
@@ -0,0 +1,279 @@
...
+        onPause()
+        removeAllViews()
+        destroy()
+    }
+    //endregion
+
+    companion object {
+        private val TAG = PoTokenWebView::class.simpleName
+        private const val REQUEST_KEY = "O43z0dpjhgX20SCx4KAo"
+        private const val JS_INTERFACE = "PoTokenWebView"
```
Bnyro, line 265: I assume that this is just the class name, so we could in theory also do `PoTokenWebView::class.simpleName!!` here instead of hardcoding?

Bnyro:
First of all, thank you for all your work spent into this PR! Everything works very nicely, the only downside is the increased loading time (as you mentioned), but that's certainly worth it.

About my review:
- I've added/pushed some small changes to abuse the Kotlin syntax, please feel free to just squash them if they're okay
- I'm a bit worried about the `GOOGLE_API_KEY` and the `SECRET_KEY` used here? I assume they are from NewPipe? If we hardcode them, this might make it easy for YouTube to fingerprint our client because they are publicly visible and YouTube could just block all request using them. That's nothing blocking this PR, it's just a side note that we maybe could consider making a build option for providing these keys in the future (if obtaining them is documented somewhere).
- I've left some other small nitpicks/questions, feel free to answer or just ignore/resolve them if they're irrelevant

With that being mentioned, I think that there's nothing in the way from merging this and creating a release soon after (if I find time for it).

Bnyro:
@gechoto I agree on you that we could still improve the errror handling here in the future, but that's nothing important enough to block this PR and could still be done with follow-up pull requests.

app/src/main/java/com/github/libretube/api/poToken/JavaScriptUtil.kt
FineFindus: No, as it's possible that `size == 1`. As much as I dislike exceptions, I also think it's better to just let the exception be thrown here and then caught further up for handling, as handling it here is not really possible or would require more code changes.

FineFindus:
> I'm a bit worried about the GOOGLE_API_KEY and the SECRET_KEY used here? I assume they are from NewPipe?

They are, but I doubt there is a way without using them, since they're required for the request. They have been in use since at least 2023 and are included in dozens of repos at this point, so I doubt there is any use in hiding them now :)

app/src/main/java/com/github/libretube/api/poToken/JavaScriptUtil.kt
FineFindus: Seems to be just how YouTube encodes them, no idea why though

app/src/main/java/com/github/libretube/api/poToken/JavaScriptUtil.kt
Bnyro: https://github.com/libre-tube/LibreTube/pull/7069/files#diff-bdd37ad197f5af84f6e4c0a47e640f0f0045754b709c2265e0cb0e7ae86cbacfR25 crashed if `size == 1`, or am I missing something?

FineFindus: No, just me being confused by `0`-indexing. But this should indeed access the [first element and not the second one](
https://github.com/LuanRT/BgUtils/blob/54c511b2bd4e3e3707f30d2907f33167e1b61b35/src/core/challengeFetcher.ts#L46). I assume it's a bug in NewPipe, though it seems like it's never actually triggered.

Bnyro (approved):
Thanks for looking into the review and rebasing!

PR #7074, opened 2025-02-05
refactor: simplify add to playlist dialog

Pittvandewitt:

- Perform Database and Network requests in the ViewModel to make it survive configuration changes
- IO work is ensured to be performed in the IO dispatcher by defining it at function level
- Playlists are only fetched once
- AddToPlaylistDialog UI state now survives process death

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/dialogs/AddToPlaylistDialog.kt
+++ b/app/src/main/java/com/github/libretube/ui/dialogs/AddToPlaylistDialog.kt
@@ -1,6 +1,5 @@
 package com.github.libretube.ui.dialogs
 
-import android.annotation.SuppressLint
 import android.app.Dialog
 import android.content.DialogInterface
 import android.os.Bundle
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/github/libretube/ui/dialogs/AddToPlaylistDialog.kt
```diff
@@ -44,90 +33,51 @@ class AddToPlaylistDialog : DialogFragment() {
...
-        if (success) {
-            appContext.toastFromMainDispatcher(
-                appContext.getString(R.string.added_to_playlist, playlistName)
-            )
-        } else {
-            appContext.toastFromMainDispatcher(R.string.fail)
-        }
+        return MaterialAlertDialogBuilder(requireContext())
+            .setTitle(R.string.addToPlaylist)
+            .setNegativeButton(R.string.createPlaylist) { _, _ ->
```
Bnyro, line 72: For some reason, pressing this button completely dismisses both dialogs. It's likely that this happens because clicking the dialog button causes the dialog to be dismissed. Thus the death of `AddToPlaylistDialog` causes its `childFragmentManager` to die, which is responsible for displaying `CreatePlaylistDialog`.

That's the reason for the previous workaround with `getButton(DialogInterface.BUTTON_NEGATIVE).setOnClickListener {}`

app/src/main/java/com/github/libretube/ui/models/PlaylistViewModel.kt
```diff
@@ -1,7 +1,109 @@
...
+                )
+            }.onFailure {
+                savedStateHandle[UI_STATE] = _uiState.value.copy(
+                    message = UiState.Message(R.string.unknown_error)
+                )
+            }
+        }
+    }
+
+    fun onPostiveButtonClick(playlistIndex: Int) {
```
Bnyro, line 49: nit: this should probably be called something like `addToPlaylistAtIndex` instead, `onPositiveButtonClick` is not that descriptive.

Bnyro (approved):
Thank you!

PR #7076, opened 2025-02-05
Refactor Comments

Pittvandewitt:

- Perform navigation in the Fragment instead of the RV adapter
- Pager now lives in the ViewModel and is emitted as LiveData. LiveData makes more sense because of the free lifecycle benefits you get compared to using Flow (by not having to use `repeatOnLifecycle(Lifecycle.State.STARTED)`)

Other:
- Smooth scrolling now works when pressing the scroll up button

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/adapters/CommentsPagingAdapter.kt
+++ b/app/src/main/java/com/github/libretube/ui/adapters/CommentsPagingAdapter.kt
@@ -4,66 +4,46 @@ import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
 import androidx.appcompat.content.res.AppCompatResources
-import androidx.core.os.bundleOf
 import androidx.core.text.method.LinkMovementMethodCompat
 import androidx.core.text.parseAsHtml
-import androidx.core.view.isGone
 import androidx.core.view.isVisible
 import androidx.core.view.updateLayoutParams
 import androidx.core.view.updatePadding
-import androidx.fragment.app.Fragment
-import androidx.fragment.app.commit
-import androidx.fragment.app.replace
 import androidx.paging.PagingDataAdapter
 import com.github.libretube.R
 import com.github.libretube.api.obj.Comment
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/github/libretube/ui/adapters/CommentsPagingAdapter.kt
```diff
@@ -4,104 +4,78 @@ import android.view.LayoutInflater
...
             commentAuthor.setBackgroundResource(
                 if (comment.channelOwner) R.drawable.comment_channel_owner_bg else 0
             )
             commentInfos.text = comment.commentedTime
 
             commentText.movementMethod = LinkMovementMethodCompat.getInstance()
-            val linkHandler = LinkHandler {
-                clickEventConsumedByLinkHandler = true
-                handleLink?.invoke(it)
-            }
```
Bnyro, line 77: This code existed for a reason.

When you click the link in the first comment of video `https://youtu.be/AdygBbbEnco` for example, the app opens the links **and** the comment's replies at the same time. However, when clicking a link, the replies should **not** be opened as a side effect.

Bnyro (changes requested):

app/src/main/java/com/github/libretube/ui/adapters/CommentsPagingAdapter.kt
```diff
@@ -4,111 +4,84 @@ import android.view.LayoutInflater
...
         holder.binding.apply {
+            val comment = getItem(position)!!
             commentAuthor.text = comment.author
             commentAuthor.setBackgroundResource(
                 if (comment.channelOwner) R.drawable.comment_channel_owner_bg else 0
             )
-            comment.commentedTimeMillis?.let {
-                commentInfos.text = TextUtils.formatRelativeDate(it)
-            } ?: comment.commentedTime
+            commentInfos.text = comment.commentedTime
```
Bnyro, line 53: Please undo that line, perhaps you accidentally changed it on rebase

PR #7077, opened 2025-02-05
fix: don't hardcode package name for setting different app icons

Bnyro:


robcle:
@Bnyro Libretube crashes and does not change the icon for me now.

Issue #7072, opened 2025-02-04
New videos not added to feed


FineFindus:
### Steps to reproduce

1. Refresh feed when there are new videos
2. Notice how the badge is updated
3. Notice that new videos are not added to the feed adapter

### Expected behavior

The new videos should be shown in the feed. This is a recent regression (somewhere after https://github.com/libre-tube/LibreTube/pull/7031 was merged), maybe introduced when refactoring the adapter?

### Actual behavior

Videos are only added to the feed once revisiting the subscription page

### LibreTube version

master

### Android version

Android 15

### Other details

Local Feed extraction is enabled.

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [x] I have written a short but informative title.
- [x] I will fill out all of the requested information in this form.

FineFindus:
I think this can also be triggered by switching channel groups since they use the same code path. I've bisected it down to 0cf7abb07dbe041e35e45e1ea06063c804a8c77f (cc @Pittvandewitt), but I'm out of my depth here.

PR #7085, opened 2025-02-09
fix: multiple recyclerview adapter regressions in SubscriptionsFragment

Pittvandewitt:
Fixes https://github.com/libre-tube/LibreTube/issues/7072

Fixes a crash when selecting a different subscription group and going back to the 'all' group

Be aware that this PR removes the local pagination. I think more performance improvements can be gained by making changes to the videos adapter compared to what the local pagination can achieve.


Bnyro:
Local pagination had been added in order to make it easier to adapt when https://github.com/TeamPiped/Piped/pull/2353 is merged, though it currently doesn't seem so.

Bnyro (approved):

Issue #7090, opened 2025-02-11
Not to refresh when bluetooth keyboard connected


bermount:
### Describe your suggested feature

App to not refresh when bluetooth keyboard connected.

### Other details

I found that in AndroidManifest.xml, 
changing

android:configChanges="orientation|screenLayout|screenSize|smallestScreenSize"

to

android:configChanges="keyboard|keyboardHidden|orientation|screenLayout|screenSize|smallestScreenSize"

I've tested it and it prevents the app from refreshing when bluetooth keyboard is connected. 
But as I don't have much knowledge about the codes, I don't know if it's the right way to solve the problem or what other consequences it might bring. 
I hope it'll be fixed though. 

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [x] I have written a short but informative title.
- [x] I will fill out all of the requested information in this form.

Bnyro:
I think that's the right way for handling it.

That change tells Android that the app handles the keyboard configuration change itself and thus shouldn't be forced to restart by Android.

Please feel free to create a pull request with these changes :+1: 

PS: This should be applied to **all** activities the app has.

PR #7091, opened 2025-02-12
fix: prevent app restart on configuration change

bermount:
- #7090 Prevent app restart on configuration change(keyboard changes, switching between system dark mode and light mode, etc)

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -114,14 +114,14 @@
 
         <activity
             android:name=".ui.activities.OfflinePlayerActivity"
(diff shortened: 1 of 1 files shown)


Bnyro (approved):
Thank you!

Pittvandewitt:
Is there a reason apart from setting the grid column size why configChanges is overridden at all?

Bnyro:
> Is there a reason apart from setting the grid column size why configChanges is overridden at all?
>
Fullscreen behavior in `PlayerFragment.kt`, including the `Auto-fullscreen` setting.

PR #7112, opened 2025-02-18
feat: optimize extraction performance & simplify code

Briiqn:
+ bump NewPipeExtractor to 0.24.5
+ simplify NewPipeDownloaderImpl.kt implementation
+ add parallel processing and avatar caching to StreamsExtractor.kt
+ add batching to DeArrow API calls.


app/src/main/java/com/github/libretube/api/StreamsExtractor.kt
```diff
@@ -2,28 +2,24 @@ package com.github.libretube.api
...
 import com.github.libretube.R
-import com.github.libretube.api.obj.ChapterSegment
-import com.github.libretube.api.obj.Message
-import com.github.libretube.api.obj.MetaInfo
-import com.github.libretube.api.obj.PipedStream
-import com.github.libretube.api.obj.PreviewFrames
-import com.github.libretube.api.obj.StreamItem
-import com.github.libretube.api.obj.Streams
-import com.github.libretube.api.obj.Subtitle
+import com.github.libretube.api.obj.*
```
Bnyro, line 5: Please don't do wildcard imports.

settings.gradle.kts
```diff
@@ -3,6 +3,7 @@ pluginManagement {
         gradlePluginPortal()
         google()
         mavenCentral()
+        maven("https://jitpack.io")
```
Bnyro, line 6: That's not required, bumping the NewPipeExtractor also works without it seems.

app/src/main/java/com/github/libretube/util/NewPipeDownloaderImpl.kt
```diff
@@ -11,50 +9,48 @@ import org.schabi.newpipe.extractor.downloader.Response
...
+            .method(httpMethod, dataToSend?.toRequestBody())
             .url(url)
-            .addHeader(USER_AGENT_HEADER_NAME, USER_AGENT)
+            .addHeader("User-Agent", USER_AGENT)
 
-        for ((headerName, headerValueList) in request.headers()) {
-            requestBuilder.removeHeader(headerName)
-            for (headerValue in headerValueList) {
+        headers.forEach { (headerName, headerValueList) ->
+            headerValueList.forEach { headerValue ->
```
Bnyro, line 28: Using normal for loops is completely valid as well and easier to read in my opinion.

app/src/main/java/com/github/libretube/util/DeArrowUtil.kt
```diff
@@ -9,49 +9,61 @@ import com.github.libretube.api.obj.Streams
...
+            Log.e(this::class.java.name, "Failed to fetch DeArrow content: ${e.message}")
             null
         }
     }
 
+    private suspend fun fetchDeArrowContentParallel(
+        videoIds: List<String>
+    ): Map<String, DeArrowContent> = coroutineScope {
+        videoIds.chunked(50)
+            .map { batch -> async { fetchDeArrowContentBatch(batch) ?: emptyMap() } }
```
Bnyro, line 39: What exactly is this parallelization trying to solve?
We previously did one API request for all video ids, so this change attempts to make things faster by fetching 50 video ids per request and parallelize all requests?
Wouldn't that also possibly make chances of being rate-limited from DeArrow higher?
Don't we usually have page sizes <= 20 anyways usually, so there's no benefit at all because it's always just one request?

app/src/main/java/com/github/libretube/api/StreamsExtractor.kt
```diff
@@ -151,7 +181,7 @@ object StreamsExtractor {
         )
     }
 
-    fun getExtractorErrorMessageString(context: Context, exception: Exception): String {
+   fun getExtractorErrorMessageString(context: Context, exception: Exception): String {
```
Bnyro, line 184: nit: wrong indention

app/src/main/java/com/github/libretube/api/StreamsExtractor.kt
```diff
@@ -58,13 +55,71 @@ fun StreamInfoItem.toStreamItem(
...
+                }.getOrElse { -1 } else -1
+            }
+
+            val relatedStreams = async {
+                resp.relatedItems
+                    .filterIsInstance<StreamInfoItem>()
+                    .chunked(4)
+                    .map { chunk ->
+                        chunk.map { item ->
+                            async { item.toStreamItem() }
```
Bnyro, line 79: I don't think that this specific part needs to be parallelized, `toStreamItem` doesn't do any API calls.

app/src/main/java/com/github/libretube/util/NewPipeDownloaderImpl.kt
```diff
@@ -11,50 +9,48 @@ import org.schabi.newpipe.extractor.downloader.Response
 import org.schabi.newpipe.extractor.exceptions.ReCaptchaException
 
 class NewPipeDownloaderImpl : Downloader() {
-    private val client: OkHttpClient = OkHttpClient.Builder()
-        .readTimeout(READ_TIMEOUT_SECONDS, TimeUnit.SECONDS)
```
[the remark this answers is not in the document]
Briiqn: Hey, the default timeout for okhttp is 10 seconds for rw  & connect which i believe is sufficient enough, I can always change it back if you want.

app/src/main/java/com/github/libretube/util/DeArrowUtil.kt
Briiqn: Oh it is, i just saw this, it was late and i thought it was 1 request per video id
image

Bnyro (approved):
LGTM, thanks

PR #7113, opened 2025-02-18
feat(LocalFeedRepository): filter members only videos

FineFindus:
YouTube has started to insert Members only videos into the normal video feed. As there not watchable without signing in, they only clutter the subscription feed.
This filters them out before they're displayed.

Requires https://github.com/TeamNewPipe/NewPipeExtractor/pull/1280

This is still a draft as extractor PR is not yet merged. It might also make sense to still store the videos, so they can be used when deciding if a full channel fetch is needed?

### DIFF
--- a/app/src/main/java/com/github/libretube/repo/LocalFeedRepository.kt
+++ b/app/src/main/java/com/github/libretube/repo/LocalFeedRepository.kt
@@ -21,6 +21,7 @@ import org.schabi.newpipe.extractor.channel.tabs.ChannelTabInfo
 import org.schabi.newpipe.extractor.channel.tabs.ChannelTabs
 import org.schabi.newpipe.extractor.feed.FeedInfo
 import org.schabi.newpipe.extractor.stream.StreamInfoItem
+import org.schabi.newpipe.extractor.stream.StreamInfoItem.ContentAvailability
 import java.time.Duration
 import java.time.Instant
 import java.util.concurrent.atomic.AtomicInteger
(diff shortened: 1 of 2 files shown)


Bnyro:
> This is still a draft as extractor PR is not yet merged. It might also make sense to still store the videos, so they can be used when deciding if a full channel fetch is needed?
>
That makes sense, yes. There shouldn't be that many paid videos, so it likely won't make much of a difference in storage space used to the end user.

FineFindus:
I've been thinking about this a bit over the last few days, and I don't think it's possible/a good idea to save them, as there are some channels that will first upload a video as a members-only video, and only later set it to public for all (á la early access). This could lead to videos not being shown at all or being shown late. In both cases, users would miss out on videos.

Bnyro:
> I've been thinking about this a bit over the last few days, and I don't think it's possible/a good idea to save them, as there are some channels that will first upload a video as a members-only video, and only later set it to public for all (á la early access). This could lead to videos not being shown at all or being shown late. In both cases, users would miss out on videos.
>
That's a valid argument, I agree.

FineFindus:
As we now include the required NewPipeExtractor API in our [extractor fork](https://github.com/libre-tube/LibreTube/pull/7293), we can merge this as well. However, it should be noted that it is absolutely possible that the upstream PR will have API-breaking changes before being merged (as it already did once), so it may require a fix up in the future.

Bnyro (approved):
Awesome, thanks for working on that upstream :)

(fyi: the force push was just a rebase)

PR #7117, opened 2025-02-23
fix(SubscriptionsFragment): skip upcoming videos when adding caught-up

FineFindus:
Fixes a regression introduced in https://github.com/libre-tube/LibreTube/commit/d34d34e9e6bad1e9e2f32da8bb9ea9aeebb63589, which caused the original fix (https://github.com/libre-tube/LibreTube/commit/111cd1b3507da4240cb339382b63aef622e3bfff) to no longer function correctly.

I also used to opportunity to rename the title of the caught-up item to `All caught up`, in order to avoid having both the title and the subtitle starting with `You`.


Bnyro (approved):
Thank you!

PR #7141, opened 2025-03-01
feat(Feed): update local feed data when watching video

FineFindus:
As local feed videos are only fetched once, this can lead to outdated data being displayed in the feed. As we already fetch the metadata when the video is being played, use this data to update the locally stored feed data.

### DIFF
--- a/app/src/main/java/com/github/libretube/db/dao/SubscriptionsFeedDao.kt
+++ b/app/src/main/java/com/github/libretube/db/dao/SubscriptionsFeedDao.kt
@@ -4,6 +4,7 @@ import androidx.room.Dao
 import androidx.room.Insert
 import androidx.room.OnConflictStrategy
 import androidx.room.Query
+import androidx.room.Update
 import com.github.libretube.db.obj.SubscriptionsFeedItem
 
 @Dao
@@ -13,6 +14,9 @@ interface SubscriptionsFeedDao {
 
     @Insert(onConflict = OnConflictStrategy.REPLACE)
     suspend fun insertAll(feedItems: List<SubscriptionsFeedItem>)
+
+    @Update
+    suspend fun update(feedItem: SubscriptionsFeedItem)
 
     @Query("SELECT EXISTS (SELECT * FROM feedItem WHERE videoId = :videoId)")
     suspend fun contains(videoId: String): Boolean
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/github/libretube/services/OnlinePlayerService.kt
```diff
@@ -138,9 +139,14 @@ open class OnlinePlayerService : AbstractPlayerService() {
...
 
-        // save the current stream to the queue
         streams?.toStreamItem(videoId)?.let {
+            // save the current stream to the queue
             PlayingQueue.updateCurrent(it)
+
+            // update feed item with newer information, e.g. more up-to-date views
+            DatabaseHolder.Database.feedDao().update(
+                it.toFeedItem()
+            )
```
Bnyro, line 149: I don't like to bypass the `FeedRepository` logic here, this should only be done if the `LocalFeedRepository` is used and thus we should likely add an abstract method `submitFeedItemChange` or something similar here:
https://github.com/libre-tube/LibreTube/blob/master/app/src/main/java/com/github/libretube/repo/FeedRepository.kt

I know that'd result in the same behavior because `@Update` wouldn't actually update anything if there's no local feed item, but that makes it easier to overview in the future.

app/src/main/java/com/github/libretube/db/dao/SubscriptionsFeedDao.kt
```diff
@@ -14,6 +15,9 @@ interface SubscriptionsFeedDao {
     @Insert(onConflict = OnConflictStrategy.REPLACE)
     suspend fun insertAll(feedItems: List<SubscriptionsFeedItem>)
 
+    @Update(onConflict = OnConflictStrategy.REPLACE)
```
Bnyro, line 18: Isn't `REPLACE` the default behavior anyways, as it's supposed to update an existing item? I think providing that strategy is only needed for inserting items.

app/src/main/java/com/github/libretube/services/OnlinePlayerService.kt
FineFindus: I'm personally not a fan of that abstraction, but I changed it.

Bnyro (approved):
Thanks!

Bnyro:
Fyi, I just rebased the PR.

PR #7142, opened 2025-03-01
fix(Feed): extract shorts uploadDate from feedInfo

FineFindus:
Shorts extracted from the Shorts tab do not have an uploadDate. They are currently saved with their uploadDate set to -1, leading in them to appear at the end of the feed. To fix this, we can use the generic FeedInfo that contains the uploadDate, which also fixes an issue, where old shorts were fetched on every refresh.

Ref: https://github.com/libre-tube/LibreTube/pull/7111


Bnyro (approved):
Thank you!

PR #7151, opened 2025-03-02
feat: add support for using local NewPipe Extractor

Bnyro:
This allows users to do everything completely locally, similar to like NewPipe does it.

That's a consequence of the fact that YouTube makes it harder and harder to run public Piped instances and therefore more and more users that relied on these instances are no longer able to use LibreTube properly.

We still lack a proper UI for that on the Welcome screen to allow user to already decide to use LibreTube completely without Piped when opening the app for the first time.

Nevertheless with this change, it's still possible to use an authentication instance to sync subscriptions and playlists using the Piped API, and fetch everything else (videos, search, trends, comments, ...) directly via the NewPipeExtractor, but users can also decide to do everything locally only.


Bnyro:
Merging now to avoid conflicts with future PRs.

milutinke:
Awesome, I'll finally be able to use the app again.
Thank you for your hard work.

PR #7154, opened 2025-03-03
feat: Implement splash screen for older Android versions

Isira-Seneviratne:
Implement a splash screen similar to the one on Android 12+ for older versions of Android using the Core Splashscreen library.

[This guide](https://developer.android.com/reference/kotlin/androidx/core/splashscreen/SplashScreen) has more information on its features and limitations.

* Before (Android 8.1)

https://github.com/user-attachments/assets/0c88c965-99de-4e5b-a094-4f56c26abbf5

* After (Android 8.1)

https://github.com/user-attachments/assets/8efb1e38-7a79-4f53-918c-58c2c181783e

### DIFF
--- a/app/src/main/res/drawable/splash_anim.xml
+++ b/app/src/main/res/drawable/splash_anim.xml
@@ -1,27 +1,6 @@
 <animated-vector xmlns:android="http://schemas.android.com/apk/res/android"
-    xmlns:aapt="http://schemas.android.com/aapt">
-    <aapt:attr name="android:drawable">
-        <vector
-            android:width="64dp"
-            android:height="64dp"
-            android:tint="?attr/colorSecondary"
-            android:viewportWidth="1280"
-            android:viewportHeight="1280">
-            <group
-                android:name="scaleGroup"
-                android:pivotX="640"
-                android:pivotY="640">
-                <path
-                    android:fillColor="#000"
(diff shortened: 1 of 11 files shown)


app/src/main/res/drawable/splash_anim.xml
```diff
@@ -1,27 +1,6 @@
 <animated-vector xmlns:android="http://schemas.android.com/apk/res/android"
-    xmlns:aapt="http://schemas.android.com/aapt">
-    <aapt:attr name="android:drawable">
-        <vector
-            android:width="64dp"
-            android:height="64dp"
-            android:tint="?attr/colorSecondary"
```
Bnyro, line 7: I assume that this color has been intentionally removed for compatibility reasons?

With the default tint, the background contrast is quite bad on the dark theme with SDK >= 31.

Isira-Seneviratne: Yeah, that wasn't being used. Will revert that change and check again.

Isira-Seneviratne: I tried reverting the tint change, but the icon is always a black icon then, so I kept the change.

Bnyro: I've added different `color` resources to make it look decent in light and in dark mode now, I think using the primary / secondary color doesn't make much sense here after rethinking it.

Bnyro (approved):
Thank you!

PR #7161, opened 2025-03-04
chore: update roadmap due to implemented local NewPipe extractor feature

Figim:
There is now a full "NewPipe extractor" option. Remove this


Bnyro (approved):
Good catch, thanks!

PR #7165, opened 2025-03-04
feat: increase corner-radius of watch-progressbar

FineFindus:
Adds a rounded corner to the watch-progressbar, making it more inline with the general design of the app, as well as how other progressbars are styled.

| Before                                                                                                               	| After                                                                                                                 	|
|----------------------------------------------------------------------------------------------------------------------	|-----------------------------------------------------------------------------------------------------------------------	|
| Watchprogress with straight edge 	| Watchprogress with rounded corner 	|


Bnyro (approved):
Neat, thank you!

PR #7176, opened 2025-03-08
fix: selected caption option not properly showing current selection

### LINKED ISSUE #7170: Caption options does not reflect the state on load video
### Steps to reproduce

The following bug "should" be easy to reproduce.
Once (the setup)
1. goto libretube settings
2. set a default language for captions
3. optionally close libretube and open it again

Repeatable (the bug):
1. Open a video with that has captions of the selected language.
2. Confirm it is showing captions.
3. Look for the video settings → captions

### Expected behavior

Notice that captions will say "None"

Image of the issue attached:

Image

### Actual behavior

Show the selected language on the list if applicable instead of "None"

### LibreTube version

Libretube Debug 0.27.1 (60) (Aplicable to its stable branch also)

### Android version

Lineage os 20 (android 13)

### Other details

This is a minor non function breaking issue, but let's log it anyways.
It also may be easy to fix (may is the keyword) as my naive outside view tells me that it is only a matter of adding another task of updating options during a video load, but I might be wrong and I just rather prefer allow who knows their stuff cook.

Also it may or may not be intentional but I rather ask, is the right alignment of captions intentional or a bug?
as it may be intentional I add it here as a foot note instead another issue

### Acknowledgements

Bnyro:
closes #7170


alejandro-alzate:
ow cool

Issue #7193, opened 2025-03-13
Sponsor Block is not working in "FULL LOCAL MODE"


Golukumar00:
### Steps to reproduce

when I use the app without turning ON  " Full local Mode"  sponser block works but when I On the setting it doesn't work at all.
Here is the video 🔗 https://youtu.be/LJOQKpAfvqU

### Expected behavior

This should be working 

### Actual behavior

Not working 

### LibreTube version

Recent nightly build

### Android version

Android 12

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [x] I have written a short but informative title.
- [x] I will fill out all of the requested information in this form.

Bnyro:
> However, Dislikes are still fetched when in local in full local mode, so I think it makes sense to also enable SponsorBlock, etc. I'm happy to make a PR for that.
>
We previously used Piped's API for that, hence it's been left unimplemented here. We can indeed use the SponsorBlock API directly, but I didn't do that yet because I only wanted to focus on the core functionality for the beginning.
PRs welcome 👍

PR #7195, opened 2025-03-13
feat(local): implement support for SponsorBlock and DeArrow

FineFindus:
Implements support for SponsorBlock and DeArrow when using the full local mode.

Please note that I was unable to confirm that DeArrow works correctly, as neither the local nor the Piped API worked, so I'm assuming it is an external failure.

Closes: https://github.com/libre-tube/LibreTube/issues/7193

### DIFF
--- a/app/src/main/java/com/github/libretube/api/NewPipeMediaServiceRepository.kt
+++ b/app/src/main/java/com/github/libretube/api/NewPipeMediaServiceRepository.kt
@@ -20,6 +20,8 @@ import com.github.libretube.api.obj.StreamItem.Companion.TYPE_PLAYLIST
 import com.github.libretube.api.obj.StreamItem.Companion.TYPE_STREAM
 import com.github.libretube.api.obj.Streams
 import com.github.libretube.api.obj.Subtitle
+import com.github.libretube.extensions.parallelMap
+import com.github.libretube.extensions.sha256Sum
 import com.github.libretube.extensions.toID
 import com.github.libretube.helpers.NewPipeExtractorInstance
 import com.github.libretube.helpers.PlayerHelper
@@ -48,7 +50,6 @@ import org.schabi.newpipe.extractor.stream.AudioStream
 import org.schabi.newpipe.extractor.stream.StreamInfo
 import org.schabi.newpipe.extractor.stream.StreamInfoItem
 import org.schabi.newpipe.extractor.stream.VideoStream
-import java.security.MessageDigest
 
 
 private fun VideoStream.toPipedStream() = PipedStream(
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/github/libretube/api/NewPipeMediaServiceRepository.kt
```diff
@@ -319,14 +321,50 @@ class NewPipeMediaServiceRepository : MediaServiceRepository {
...
+        actionType: List<String>?
+    ): SegmentData {
+        // use hashed video id for privacy
+        // https://wiki.sponsor.ajay.app/w/API_Docs#GET_/api/skipSegments/:sha256HashPrefix
+        val hashedId = MessageDigest.getInstance("SHA-256")
+            .digest(videoId.toByteArray())
+            .toHexString()
+
+        return RetrofitInstance.externalApi.getSegments(
+            hashedId.substring(0..4),
```
Bnyro, line 337: `substring(0..4)` would be `hashedId[0]...hashedId[4]` (5 chars), while below you used `substring(0, 4)` which excludes `hashedId[4]`. Probably a typo?

app/src/main/java/com/github/libretube/api/NewPipeMediaServiceRepository.kt
```diff
@@ -319,14 +321,50 @@ class NewPipeMediaServiceRepository : MediaServiceRepository {
...
         videoId: String,
-        category: String,
-        actionType: String?
-    ): SegmentData = SegmentData()
+        category: List<String>,
+        actionType: List<String>?
+    ): SegmentData {
+        // use hashed video id for privacy
+        // https://wiki.sponsor.ajay.app/w/API_Docs#GET_/api/skipSegments/:sha256HashPrefix
+        val hashedId = MessageDigest.getInstance("SHA-256")
```
Bnyro, line 332: I would prefer moving the hashing function to a helper method since its implementation details are unrelated to the logic of this method, it could even make sense to create an extension function `String.sha256sum()`.

app/src/main/java/com/github/libretube/api/NewPipeMediaServiceRepository.kt
```diff
@@ -319,14 +321,50 @@ class NewPipeMediaServiceRepository : MediaServiceRepository {
...
+                        .toHexString()
+                    RetrofitInstance.externalApi.getDeArrowContent(
+                        hashedId.substring(0, 4)
+                    )
+                }
+            }.reduce { acc, map -> acc + map }.mapValues { (videoId, value) ->
+                value.copy(
+                    thumbnails = value.thumbnails.map { thumbnail ->
+                        thumbnail.takeIf { it.original } ?: thumbnail.copy(
+                            thumbnail = "https://dearrow-thumb.ajay.app/api/v1/getThumbnail?videoID=$videoId&time=${thumbnail.timestamp}"
```
Bnyro, line 362: Please put the base url to a constant.

app/src/main/java/com/github/libretube/api/NewPipeMediaServiceRepository.kt
```diff
@@ -319,14 +321,50 @@ class NewPipeMediaServiceRepository : MediaServiceRepository {
...
+        // use hashed video id for privacy
+        // https://wiki.sponsor.ajay.app/w/API_Docs/DeArrow#GET_/api/branding/:sha256HashPrefix
+        val digest = MessageDigest.getInstance("SHA-256")
+        return videoIds.split(',').chunked(25)
+            .flatMap {
+                it.parallelMap { videoId ->
+                    val hashedId = digest
+                        .digest(videoId.toByteArray())
+                        .toHexString()
+                    RetrofitInstance.externalApi.getDeArrowContent(
```
Bnyro, line 354: Maybe we should use `runCatching` here because if one chunk of data fails to load (for whatever reasons, e.g. ratelimits), we could still display the results from the other requests.

app/src/main/java/com/github/libretube/api/NewPipeMediaServiceRepository.kt
FineFindus: Just me being used to Rust Ranges :)

Bnyro (approved):
Thank you!

PR #7206, opened 2025-03-16
 chore: update privacy policy due to recent addition of full local data extraction

Figim:
should now change according to NewPipe Extractor


Bnyro (approved):
I've modified it a bit, thank you :+1:

PR #7214, opened 2025-03-16
perf(LocalFeedRepository): speed up local feed extraction

FineFindus:
Speeds up the local feed extraction, by fetching more feeds concurrently and only applying a large timeout after fetching the full channels. This is based on the idea that YouTube's RSS feeds (extracted by the generic `FeedInfo`) have [no/much less rate-limiting](https://github.com/TeamNewPipe/NewPipe/issues/11817), likely due to the nature of RSS readers. I couldn't find an explicit statement from YouTube about that, but it makes sense. I've also [increased to amount of concurrently fetched feeds](https://github.com/libre-tube/LibreTube/commit/7503ef122daf45c9929e3ab122e6bc4b18adc314) in my fork for the last couple of months and had no noticeable impact so far.
A small timeout after each chunk is still applied, to be a good citizen of the web and not overwhelm/spam the server.

The used values may need to be adjusted in the future, either to allow for even faster extraction or throttling if a rate-limit is introduced.

### DIFF
--- a/app/src/main/java/com/github/libretube/repo/LocalFeedRepository.kt
+++ b/app/src/main/java/com/github/libretube/repo/LocalFeedRepository.kt
@@ -81,31 +81,32 @@ class LocalFeedRepository : FeedRepository {
         if (channelIds.isEmpty()) return
 
         val totalExtractionCount = AtomicInteger()
-        val chunkedExtractionCount = AtomicInteger()
+        val channelExtractionCount = AtomicInteger()
         withContext(Dispatchers.Main) {
             onProgressUpdate(FeedProgress(0, channelIds.size))
         }
 
         for (channelIdChunk in channelIds.chunked(CHUNK_SIZE)) {
-            // add a delay after each BATCH_SIZE amount of visited channels
-            val count = chunkedExtractionCount.get();
+            val count = channelExtractionCount.get();
             if (count >= BATCH_SIZE) {
-                delay(BATCH_DELAY.random())
-                chunkedExtractionCount.set(0)
+                // add a delay after each BATCH_SIZE amount of fully-fetched channels
+                delay(CHANNEL_BATCH_DELAY.random())
+                channelExtractionCount.set(0)
             }
 
             val collectedFeedItems = channelIdChunk.parallelMap { channelId ->
                 try {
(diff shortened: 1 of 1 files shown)


FineFindus:
While I agree that immediately showing new items is a good idea, I doubt a good implementation is possible, as we don't know which channels have uploaded, so the new entries would constantly be reshuffled, which I think would be rather confusing and possibly nauseating. Additionally, constantly adding new items at (essentially) random points in the recycler view might lead to performance problems.

That aside, I don't see how this is related to this PR? I think it would be better to open a new issue.

Bnyro:
> While I agree that immediately showing new items is a good idea, I doubt a good implementation is possible, as we don't know which channels have uploaded (and in which order), so the new entries would constantly be reshuffled, which I think would be rather confusing and possibly nauseating. Additionally, constantly adding new items at (essentially) random points in the recycler view might lead to performance problems.
>
I agree that immediately showing the loaded videos ain't a good idea. Apart from the mentioned points it's also not very easy to implement with the currently used app architecture, as we still need some kind of compatibility with the Piped way of loading the feed.

Bnyro:
What's the reason for the current draft state? The PR seems ready to be merged at first glance, doesn't it?

FineFindus:
Code-wise yes, and I thought so too, but yesterday I noticed some rate limiting in playback. I also took a look at how RSS feed readers handle concurrent fetching, so I would like to adjust the values here again and test it in the next few days.

app/src/main/java/com/github/libretube/repo/LocalFeedRepository.kt
```diff
@@ -81,31 +81,32 @@ class LocalFeedRepository : FeedRepository {
         if (channelIds.isEmpty()) return
 
         val totalExtractionCount = AtomicInteger()
-        val chunkedExtractionCount = AtomicInteger()
+        val channelExtractionCount = AtomicInteger()
```
Bnyro, line 84: We could possibly add some small comments about the different atomic integers and their difference here, but that's not that important

Bnyro (approved):
Didn't experience any throttling while testing and the changes are certainly a UX improvement :+1:

Issue #7240, opened 2025-03-24
Channel content does not load when scrolling up with "Full local Mode" 


Figim:
### Steps to reproduce

1. Enable "Full Local Mode" 
2. Go with the famous channel. Thousands of **videos, live streams, shorts, playlists** 
3. Scroll up to see all the other content in the "videos, live streams, shorts, playlists" sections. Also, not all videos in playlists are loading.  However, LibreTube doesn't loading anything. 

### Expected behavior

We need to scroll up to see everything in the channel. 

### Actual behavior

**LibreTube (videos, live streams, shorts, playlists)**

This is the case in all sections.

https://github.com/user-attachments/assets/a4191c45-6bbf-4fab-a3ec-2e1ce0ba5bb3

https://github.com/user-attachments/assets/e62e8544-8f90-4c12-afc0-3a2cf6b66dbf

**LibreTube ( playlists videos)**

https://github.com/user-attachments/assets/067405d2-7a5f-4185-911b-a998b31e48e3

### LibreTube version

0.27.1. / Debug Full Local Mode

### Android version

Misc

### Other details

**Newpipe loading everything** 

https://github.com/user-attachments/assets/1d4c09ca-2d05-4059-bf4f-705c5043a4fa

https://github.com/user-attachments/assets/54376276-fedf-449a-aaeb-cb8b4a728a7d

https://github.com/user-attachments/assets/2aad7893-408f-4db2-8656-3329446cf566

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [x] I have written a short but informative title.
- [x] I will fill out all of the requested information in this form.

PR #7244, opened 2025-03-25
fix(Local): encode nextpage data as base64

FineFindus:
Fixes an issue, with local mode enabled, users where unable to continously load channel data.
This was caused by the channel nextpage data being stored as a JVM identifier (the default byte array .toString behaviour), rather than the expected base64 encoded string.

Closes: https://github.com/libre-tube/LibreTube/issues/7240


Bnyro (approved):
Thanks for investigating this!

PR #7258, opened 2025-03-31
fix(WatchHistory): always show filter bar

FineFindus:
Fixes an issue, where the users could get soft-locked, by selecting a filter, which had no items, leading to the filter bar being hidden, and thus making the user unable to deselect the filter. This is fixed, by only hiding recyclerView/No items views.

| Before                                                                                                               	| After                                                                                                             	|
|----------------------------------------------------------------------------------------------------------------------	|-------------------------------------------------------------------------------------------------------------------	|
| Watch history with no filter bar 	| Watch history with filter bar 	|


app/src/main/java/com/github/libretube/ui/fragments/WatchHistoryFragment.kt
```diff
@@ -149,7 +149,8 @@ class WatchHistoryFragment : DynamicLayoutManagerFragment(R.layout.fragment_watc
 
         viewModel.filteredWatchHistory.observe(viewLifecycleOwner) { history ->
             binding.historyEmpty.isGone = history.isNotEmpty()
-            binding.historyContainer.isVisible = history.isNotEmpty()
+            binding.playAll.isEnabled = history.isNotEmpty()
```
Bnyro, line 152: We should probably also disable the clear history button here?

FineFindus: I thought about it, but I don't think so, since the clear button removes all items, and not just the visible ones. I guess it should rather be disabled, if the overall history is empty and not just the filtered one?

Bnyro (approved):
Ty!

Issue #7279, opened 2025-04-02
Inconsistent logo design across app


handokota:
### Steps to reproduce

1. Install and launch the app.  
2. Compare the **shape/design** of the logo in:  
   - The **app icon** (home screen),
   - **Splash screen** (on launch),
   - **Help** and **About** pages (in-app).
   - **Notification icon** (status bar/notifications).

### Expected behavior

All instances of the logo should use the exact same design/shape.

### Actual behavior

The logo’s shape differs between screens.

### LibreTube version

0.27.1

### Android version

Android 14

### Other details

App icon:
![Image](https://github.com/user-attachments/assets/76e7f116-f0b6-44cf-8ad4-670b557d5af1)

Splash screen, help, about pages:
![Image](https://github.com/user-attachments/assets/cb949e3a-172e-49bc-b3af-2720b2d04127)

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [x] I have written a short but informative title.
- [x] I will fill out all of the requested information in this form.

PR #7280, opened 2025-04-02
fix(ui): standardize logo design across all app instances

handokota:
see #7279


Bnyro (approved):
Thank you!

I've pushed some small size changes since the icon has a different size now, looks good to me apart from that!

PR #7308, opened 2025-04-15
refactor: split video and trending layout into different adapters

Bnyro:
The layouts don't really have much in common and probably never should have been placed in the same file (it would probably have been much better to do some sort of delegation instead).

However that's irrelevant now, this closely coupled code is no longer needed because we recently removed the "Alternative Videos Layout" option, so adapters are now always from a fixed type (either card/trending layout or video row) no matter the users preferences.


app/src/main/java/com/github/libretube/ui/adapters/TrendingAdapter.kt
```diff
@@ -0,0 +1,120 @@
...
+    @SuppressLint("SetTextI18n")
+    override fun onBindViewHolder(holder: TrendingViewHolder, position: Int) {
+        val video = getItem(holder.bindingAdapterPosition)
+        val videoId = video.url.orEmpty().toID()
+
+        val context = (holder.trendingRowBinding ?: holder.allCaughtUpBinding)!!.root.context
+        val activity = (context as BaseActivity)
+        val fragmentManager = activity.supportFragmentManager
+
+        // Trending layout
```
FineFindus, line 65: I don't think that comment is necessary anymore.

FineFindus (approved):
LGTM. It might be a good idea to rename the `Trending` layout as it's not just used for the trending feed, and thus the name is a bit misleading, but nothing springs to mind for me.

Bnyro:
> LGTM. It might be a good idea to rename the `Trending` layout as it's not just used for the trending feed, and thus the name is a bit misleading, but nothing springs to mind for me.
>
I thought about that too, maybe something like `VideoCardAdapter`?

PR #7319, opened 2025-04-20
feat: display sponsorblock video icon on thumbnail for fully sponsored videos

FineFindus:
Displays an icon on the thumbnail for fully sponsored videos. This allows users to easily identify and avoid these videos, if they prefer so. This works similarly to how the SponsorBlock extension works for the website.

SponsorBlock icon on thumbnail

This is still marked as a draft as:
- it always uses the official SponsorBlock server, as Piped does not support the required API (not sure if it is worth fixing, with the current state of the Piped backend?)
- it increases the number of requests made, we should probably cache or store them directly in the DB (though that might lead to outdated data)
- it's only shown for VideoCards (not sure if it even makes sense for the 'normal' video adapter)
- the icon position and size feels suboptimal, due to the difference to the duration
- not sure if that is something that we even want to implement :)

### DIFF
--- a/app/src/main/java/com/github/libretube/api/obj/Segment.kt
+++ b/app/src/main/java/com/github/libretube/api/obj/Segment.kt
@@ -16,7 +16,7 @@ data class Segment(
     val category: String? = null,
     val description: String? = null,
     val locked: Int? = null,
-    private val segment: List<Float> = listOf(),
+    private val segment: List<Float> = listOf(0f, 0f),
     val userID: String? = null,
     val videoDuration: Double? = null,
     val votes: Int? = null,
--- a/app/src/main/java/com/github/libretube/ui/adapters/VideoCardsAdapter.kt
+++ b/app/src/main/java/com/github/libretube/ui/adapters/VideoCardsAdapter.kt
@@ -4,8 +4,10 @@ import android.annotation.SuppressLint
 import android.view.LayoutInflater
 import android.view.ViewGroup
 import androidx.core.os.bundleOf
+import androidx.core.view.isVisible
 import androidx.core.view.updateLayoutParams
 import androidx.recyclerview.widget.ListAdapter
+import com.github.libretube.api.MediaServiceRepository
 import com.github.libretube.api.obj.StreamItem
 import com.github.libretube.constants.IntentData
 import com.github.libretube.databinding.AllCaughtUpRowBinding
@@ -21,6 +23,10 @@ import com.github.libretube.ui.extensions.setWatchProgressLength
 import com.github.libretube.ui.sheets.VideoOptionsBottomSheet
 import com.github.libretube.ui.viewholders.VideoCardsViewHolder
 import com.github.libretube.util.TextUtils
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
 
 class VideoCardsAdapter(private val columnWidthDp: Float? = null) :
     ListAdapter<StreamItem, VideoCardsViewHolder>(DiffUtilItemCallback()) {
@@ -106,6 +112,14 @@ class VideoCardsAdapter(private val columnWidthDp: Float? = null) :
                 sheet.arguments = bundleOf(IntentData.streamItem to video)
                 sheet.show(fragmentManager, VideoCardsAdapter::class.java.name)
                 true
+            }
+
+            CoroutineScope(Dispatchers.IO).launch {
+                val sponsor = runCatching { MediaServiceRepository.instance.getVideoLabels(videoId) }.getOrNull()
+
+                withContext(Dispatchers.Main) {
+                    sponsorBadge.isVisible = sponsor?.segments?.isNotEmpty() == true
+                }
             }
         }
     }
--- a/app/src/main/res/layout/trending_row.xml
+++ b/app/src/main/res/layout/trending_row.xml
@@ -33,6 +33,32 @@
                 android:layout_width="wrap_content"
                 android:layout_height="wrap_content"
                 android:layout_gravity="end"
+                android:layout_margin="5dp"
+                app:cardBackgroundColor="@color/duration_background_color"
+                app:cardCornerRadius="8dp"
+                app:cardElevation="0dp"
+                app:layout_constraintStart_toStartOf="parent"
+                app:layout_constraintTop_toTopOf="parent">
+
+                <ImageView
+                    android:id="@+id/sponsor_badge"
+                    android:layout_width="wrap_content"
+                    android:layout_height="wrap_content"
+                    android:paddingHorizontal="6dp"
+                    android:paddingVertical="2dp"
+                    android:src="@drawable/ic_block"
+                    android:visibility="gone"
+                    android:layout_gravity="center"
+                    app:tint="@color/duration_text_color"
+                    tools:ignore="RtlSymmetry"
+                    tools:visibility="visible" />
+
+            </androidx.cardview.widget.CardView>
+
+            <androidx.cardview.widget.CardView
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:layout_gravity="end"
                 android:layout_marginEnd="5dp"
                 android:layout_marginBottom="5dp"
                 app:cardBackgroundColor="@color/duration_background_color"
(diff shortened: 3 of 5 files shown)


Bnyro:
> it always uses the official SponsorBlock server, as Piped does not support the required API (not sure if it is worth fixing, with the current state of the Piped backend?)
>
I'm not planning to continue Piped Backend development, and it doesn't seem like anybody else does. So I think we should just implement this as a helper function and use it in both, the local and the piped media repo. (Let me know if you have a better idea though)
> it increases the number of requests made, we should probably cache or store them directly in the DB (though that might lead to outdated data)
>
We could possibly create a new cache file and http client for it (https://square.github.io/okhttp/features/caching/), though that'd only work if SponsorBlock's backend has the caching headers set accordingly (`Cache-Control:max-age=...`).
Storing this data inside a database is an overkill for such a small/simple feature imo.
> it's only shown for VideoCards (not sure if it even makes sense for the 'normal' video adapter)
>
It probably makes the most sense for the feed I guess :shrug:, other places are probably not worth the additional requests.
> the icon position and size feels suboptimal, due to the difference to the duration
>
The card behind could maybe be replaced with a shape that's similar to the SponsorBlock shield? Or we directly edit the SVG, though I'm not sure if Android XML/SVG files support transparency/alpha channels.

FineFindus:
> We could possibly create a new cache file and http client for it (https://square.github.io/okhttp/features/caching/), though that'd only work if SponsorBlock's backend has the caching headers set accordingly (Cache-Control:max-age=...).

I feel like most of the request made from `ExternalApi` should be cached. From my testing SponsorBlock doesn't seem to set `Cache-Control` anywhere (https://github.com/ajayyy/SponsorBlockServer/issues/576, CC: @ajayyy: the issue was closed but seems to not have been fully resolved :))
I might have a patch to add `ETag`s to `/getVideoLabels`, but I'm having trouble getting the server up and running to test my changes.

FineFindus:
I decided to just cache every response, as this avoids making a request for every thumbnail (even if the data was already requested recently, but the thumbnail was recycled/user scrolled up and down). (Not sure if we should persist the cache across app restarts)

Alternatively, if we only show the icon in the subscription feed, we could also fetch the data when the feed is refreshed and store it in the feed database (though that would only apply to the local feed).

I swapped the icon for a custom one, inspired by the one that YouTube uses for paid product placements, though it still feels a bit on the heavy-side, maybe the size should be reduced a bit? 
Updated sponsor icon

app/src/main/res/layout/trending_row.xml
```diff
@@ -29,6 +29,32 @@
...
                 android:layout_height="match_parent" />
 
+            <androidx.cardview.widget.CardView
+                android:id="@+id/sponsor_badge_card"
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:layout_gravity="end"
+                android:layout_margin="5dp"
+                android:visibility="gone"
+                app:cardBackgroundColor="@color/duration_background_color"
```
Bnyro: I still feel like this is very hard to see - might be better to use something like `colorPrimary` or `colorSecondary` and perhaps we can set `android:alpha` here to make it a bit transparent?

Not a perfect solution, but should work I think.

app/src/main/java/com/github/libretube/ui/adapters/VideoCardsAdapter.kt
```diff
@@ -124,7 +126,14 @@ class VideoCardsAdapter(private val columnWidthDp: Float? = null) :
                 true
             }
 
+            // always hide the icon, to avoid issues where the icon is recycled and shown until the web requests succeeds
+            sponsorBadgeCard.isVisible = false
             CoroutineScope(Dispatchers.IO).launch {
+                val sponsor = SponsorBlockLabelHelper.getVideoLabels(videoId)
```
Bnyro:
```suggestion
                if (!PlayerHelper.sponsorBlockEnabled) return@launch

                val sponsor = SponsorBlockLabelHelper.getVideoLabels(videoId)
```

app/src/main/java/com/github/libretube/api/obj/Segment.kt
```diff
@@ -13,32 +13,32 @@ import kotlinx.serialization.Transient
...
     val description: String? = null,
-    val locked: Int,
-    private val segment: List<Float> = listOf(),
+    val locked: Int? = null,
+    private val segment: List<Float> = listOf(0f, 0f),
     val userID: String? = null,
-    val videoDuration: Double,
-    val votes: Int,
+    val videoDuration: Double? = null,
+    val votes: Int? = null,
```
Bnyro: Are all these values always null? In this case it could make sense to define a completely new data class (or write a comment why these are nullable but at the same time we can assert them as non-null in `toDownloadSegment` because only the full video segments have these nullable while normal SponsorBlock segments have these as non-null?)

app/src/main/java/com/github/libretube/api/SponsorBlockLabelHelper.kt
```diff
@@ -0,0 +1,35 @@
+package com.github.libretube.api
+
+import android.util.LruCache
+import com.github.libretube.api.obj.SegmentData
+import com.github.libretube.extensions.sha256Sum
+
+private data class CacheObject(val value: SegmentData?)
+
+object SponsorBlockLabelHelper {
+    // we cannot use segment data directly, as LruCache expects non-null objects, but we also want to cache unlabeled videos
```
Bnyro: Why are we defining `CacheObject?` as nullable then? Couldn't this be omitted?

app/src/main/java/com/github/libretube/api/SponsorBlockLabelHelper.kt
```diff
@@ -0,0 +1,35 @@
...
+    private val memoryCache = LruCache<String, CacheObject?>(256)
```
Bnyro: As most of the videos probably don't have a label anyways, this could probably be set to something higher than 256?

app/src/main/java/com/github/libretube/api/SponsorBlockLabelHelper.kt
FineFindus: Because I misread the API as always returning non-null :)

app/src/main/java/com/github/libretube/api/SponsorBlockLabelHelper.kt
FineFindus: Sure, though we still need to store the videos without a label as well. I'm happy to change it, however we only request the labels for video which have been shown, so I think it will be somewhat hard to even view 256 videos (and even then the cache would evict the least viewed).

app/src/main/java/com/github/libretube/ui/adapters/VideoCardsAdapter.kt
FineFindus: Done, though I'm not that happy using a shared prefs call on every, as they [may block](https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/core/java/android/app/SharedPreferencesImpl.java;l=278;drc=61197364367c9e404c7da6900658f1b16c42d0da).

Bnyro:
Fyi, I've slightly changed the design:

foo

FineFindus:
I've also been experimenting with another icon (I think this one should be more visible, as it has more filled space) :)
image

Bnyro:
I think that the current icon has the advantage that it's more "horizontal", so it looks more like a card (like the duration card). The new icon has an aspect ratio of almost 1/1, so it's almost a square and doesn't fit so well in the card design presumably?

Either ways, I'm fine with both if we keep the color changes :)

FineFindus:
Both icons have the same size, they just have some additional vertical padding in the card.

> Either ways, I'm fine with both if we keep the color changes :)

I'm not too sure about the color, I think it's a bit too eye-catching for a simple video label.

Bnyro:
> I'm not too sure about the color, I think it's a bit too eye-catching for a simple video label.
>
Well, I'd say it's reasonable because
- probably only few videos will have it, so users are not used to looking into the top left for the icon
- it's an important information to know that a video is fully sponsored (i.e. you might not want to watch it then)

ghost:
Would it not be better to make the icon slightly transparant to be more consistent with the time indicator in the bottom right also being slightly transparent?

FineFindus:
I don't disagree with either point, however I think we should reserve the primary color elements that demand more immediate attention? Maybe we could use the secondary or tertiary color instead?  (i.e. the color feels very strong, I think a bit more of a subtle color would work better)

> Would it not be better to make the icon slightly transparant to be more consistent with the time indicator in the bottom right also being slightly transparent?

This discussion is about how the icon/card should look. Making the icon itself transparent isn't a great idea, due to accessibility.

Bnyro:
Secondary/Tertiary sounds good 👍

PR #7358, opened 2025-04-27
feat(Player): automatically select best buffer duration

FineFindus:
Automatically selects the 'best' max buffer duration based on the amount of available memory. This improves the users experience as the player is less likely to constantly load, whilst avoid having user to configure the technical setting themselves (without any help/guidelines).

The exact values may need to be adjusted, based on user feedback.


app/src/main/java/com/github/libretube/helpers/PlayerHelper.kt
```diff
@@ -532,22 +527,40 @@ object PlayerHelper {
...
-            .setBackBuffer(1000 * 60 * 3, true)
-            .setBufferDurationsMs(
-                MINIMUM_BUFFER_DURATION,
-                max(bufferingGoal, MINIMUM_BUFFER_DURATION),
-                DefaultLoadControl.DEFAULT_BUFFER_FOR_PLAYBACK_MS,
-                DefaultLoadControl.DEFAULT_BUFFER_FOR_PLAYBACK_AFTER_REBUFFER_MS
-            )
-            .build()
-    }
+    fun getLoadControl(): LoadControl = DefaultLoadControl.Builder()
```
Bnyro, line 556: I think this method is only called when the player is created for the first time, so it stays the same during autoplay or starting other videos (except if the current player is getting closed by the user).

But isn't it very likely that the available RAM is going to change when playing the next video, e.g. via autoplay?

FineFindus: The method also isn't very accurate, since it doesn't include memory used by the kernel :) 
Changed it to be based on the total memory that the device has.

Alternatively we could also just choose a value, but in either case I would like to remove this setting, as it's not really user-friendly or helpful.

Bnyro: The PR actually tries to simplify the preferences and remove "unnecessary"/"unintuitive" preferences that are not really easy to understand or useful.

So adding a new option to toggle between auto and custom would be like the opposite of what this PR is trying to achieve imho.

app/src/main/res/values/strings.xml
```diff
@@ -158,8 +158,6 @@
     <string name="show_updates">Check updates automatically</string>
     <string name="related_streams">Related content</string>
     <string name="related_streams_summary">Show similar streams alongside what you watch</string>
-    <string name="buffering_goal">Preloading</string>
-    <string name="buffering_goal_summary">Max. amount of seconds of video to buffer</string>
```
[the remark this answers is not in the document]
FineFindus: I assumed Weblate would manage that itself when someone updates the translations for that language? It's not that much work, but it seems something that should be done automatically (e.g. it is for `.po` files).

app/src/main/java/com/github/libretube/helpers/PlayerHelper.kt
FineFindus: > Sorry for comments here but as a user I feel there should be two options in setting " auto" and "custom" . You can set "auto" by default but if some users want to change it they will.

What's the issue with the (current) default, that requires you changing it?

> For me this setting save a lots of data uses

The (network) data that is consumed is the same, it's just distributed differently.

app/src/main/res/values/strings.xml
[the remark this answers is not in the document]
Bnyro: Not 100% sure. In the past, I already had build issues due to some string resources messed up by Weblate, but I'm not 100% what used to cause it back then.

FineFindus:
Thinking about it a bit more, maybe we should also take into account the playback speed (as that is one of the major factors of what the 'best' buffer duration is). But I'm not sure how to best deal with changing speeds.

app/src/main/res/values/strings.xml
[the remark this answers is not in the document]
Bnyro: Here's what happens if you only remove the source string: https://github.com/you-apps/WallYou/actions/runs/14822641921/job/41611687792#step:4:300. So no, Weblate doesn't seem to handle it.

dhanuarf:
> Thinking about it a bit more, maybe we should also take into account the playback speed (as that is one of the major factors of what the 'best' buffer duration is). But I'm not sure how to best deal with changing speeds.

Also, different video quality means different buffer size for the same length of duration.

Actually I prefer the way it is configured right now, as @Golukumar00 mentioned above, preloading something that you dont like seems kind of wasteful especially on a limited internet bandwidth.

PR #7376, opened 2025-05-05
refactor: make feed notifications more robust against subscription changes

### LINKED ISSUE #7366: Make feed notifications more robust against subscription changes

If a user unsubscribes the channel with the "newest" video in the feed, feed notifications will behave unexpectedly.

That's because of https://github.com/libre-tube/LibreTube/blob/master/app/src/main/java/com/github/libretube/workers/NotificationWorker.kt#L91 just won't find the videoId in the current feed anymore because the uploader's channel is no longer subscribed. So in the end, the notification worker doesn't know which streams are now and which ones are old.

We shouldn't store the videoId, instead we should use the latest time the feed was updated and show all notifications newer than the last feed update.

### Other details

_No response_

### Acknowledgements

Bnyro:
closes #7366


FineFindus (commented):
Using `lastFeedWatchedTime` will cause the all-caught-up marker to be moved.
From a UX perspective that would be unexpected, as the app would notify a user about new videos, but than not actually display them as new in the feed.
Not sure, if that is intended, but I think it would be better to use a separate variable to store the notification time.

Bnyro:
> Using lastFeedWatchedTime will cause the all-caught-up marker to be moved.
From a UX perspective that would be unexpected, as the app would notify a user about new videos, but than not actually display them as new in the feed.
>
Yes, I thought about this too, but forgot to write something about in in the PR description.
> Not sure, if that is intended, but I think it would be better to use a separate variable to store the notification time.
The problem is that we already have 2 very similar variables
- `LAST_FEED_REFRESH_TIMESTAMP_MILLIS`: only used for local feed extraction
- `LAST_WATCHED_FEED_TIME`: last time the subscriptions feed was opened

and I hesitated to add yet another one to prevent additional complexity.

But yes, you're right, the way it's now with this PR doesn't make much sense.

FineFindus:
Maybe we could reuse `LAST_FEED_REFRESH_TIMESTAMP_MILLIS` here, since if it isn't updated there is no new video and if we update it in the notification, the new video must have been already fetched.

Alternatively I think it is fine to add a new variable for this as
- it's not visible/modifieable by the user
- it doesn't interact with the rest of the system

so the additional complexity is quite low.

Bnyro:
Not 100% happy with it yet, but at least it should work (although untested).

app/src/main/java/com/github/libretube/constants/PreferenceKeys.kt
```diff
@@ -112,7 +112,7 @@ object PreferenceKeys {
     const val SELECTED_FEED_FILTERS = "filter_feed"
     const val FEED_SORT_ORDER = "sort_oder_feed"
     const val LOCAL_FEED_EXTRACTION = "local_feed_extraction"
-    const val LAST_FEED_REFRESH_TIMESTAMP_MILLIS = "last_feed_refresh_timestamp_millis"
+    const val LAST_LOCAL_FEED_REFRESH_TIMESTAMP_MILLIS = "last_feed_refresh_timestamp_millis"
```
FineFindus, line 115: I would prefer to have this change in a separate commit, as it has nothing to do with the notification changes

app/src/main/java/com/github/libretube/constants/PreferenceKeys.kt
```diff
@@ -140,11 +140,13 @@ object PreferenceKeys {
...
-    const val LAST_WATCHED_FEED_TIME = "last_watched_feed_time"
     const val AUTH_PREF_FILE = "auth"
     const val IMAGE_PROXY_URL = "image_proxy_url"
     const val SELECTED_CHANNEL_GROUP = "selected_channel_group"
     const val SELECTED_DOWNLOAD_SORT_TYPE = "selected_download_sort_type"
     const val LAST_SHOWN_INFO_MESSAGE_VERSION_CODE = "last_shown_info_message_version"
+
+    // use the helper methods at PreferenceHelper to access these
+    const val LAST_USER_SEEN_FEED_TIME = "last_watched_feed_time"
+    const val LAST_REFRESHED_FEED_TIME = "last_updated_feed_time"
```
FineFindus, line 151: I'm not super happy about a comment telling user not to use something and would prefer to encode this in the type-system, but that's unrelated to the PR.

FineFindus (commented):
Just to confirm, we're no longer updating the `LastFeedWatchedTime` every time the feed is shown, but now only when the feed is fetched?

app/src/main/java/com/github/libretube/constants/PreferenceKeys.kt
Bnyro: Yeah, me neither.

Bnyro:
> Just to confirm, we're no longer updating the LastFeedWatchedTime every time the feed is shown, but now only when the feed is fetched?
>
That usually should happen at the same time because we had only been updating the feed via the subscriptions fragment (and the home fragment, but I had a boolean for it to skip updating the variable then). I've pushed changes to clarify it, there are 3 places where we update any of these variables now:
- both of them, when the user views the subscriptions fragment/feed
- only the last refreshed time (but not the watched/seen time) in the NotificationWorker
- the refreshed time everytime the feed is loaded in the view model (which is also done by the home fragment)

Bnyro:
I'll cleanup the commit history in a sec.

FineFindus (approved):
> the refreshed time everytime the feed is loaded in the view model (which is also done by the home fragment)

I think it might be worthwhile discussing if changing that to avoid cases where the all-caught-up marker disappears (e.g. a users searches something, causing the fragment to reload, loosing the marker), but this isn't the right place for it.

Issue #7403, opened 2025-05-09
Rename "Legacy subscriptions view" to "Grid subscriptions view"


ghost:
### Describe your suggested feature

I think "legacy" could give the impression that the feature receives little maintenance?

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [x] I have written a short but informative title.
- [x] I will fill out all of the requested information in this form.

Bnyro:
https://github.com/libre-tube/LibreTube/pull/7406

PR #7406, opened 2025-05-09
feat: remove legacy subscription view

FineFindus:
Drops the legacy subscription view, which was marked as legacy ~3 years ago. Given the long time, lack of updates since, it should be safe to remove. This further reduces the amount of 'unnecessary' options.

Ref: https://github.com/libre-tube/LibreTube/pull/7276
Ref: https://github.com/libre-tube/LibreTube/commit/298023b7026160863cfbb5c7c03876fd3c231d09
Closes: https://github.com/libre-tube/LibreTube/issues/7403


Bnyro (commented):
I think that this is a good step, but since this option has probably been used by tablet users mostly, we should think about improving the default subscriptions view for large screens (e.g. respecting the grid column preference and putting multiple channels in a row). Otherwise with tablets in landscape, you only have the channel avarar + name at the left and then a lot of blank space until the subscribe button.

Bnyro:
https://github.com/libre-tube/LibreTube/blob/master/app/src/main/java/com/github/libretube/ui/viewholders/LegacySubscriptionViewHolder.kt and https://github.com/libre-tube/LibreTube/blob/master/app/src/main/res/layout/legacy_subscription_channel.xml can be removed as well.

FineFindus:
Maybe we could rethink the design and use bottom sheet instead? That way the current design would work on most screen sizes.

Bnyro:
> Maybe we could rethink the design and use bottom sheet instead? That way the current design would work on most screen sizes.
>
I don't mind rethinking the design, but wouldn't a bottom sheet just make it more complicated to browse the subcriptions? Although it coiluld for instance allow us to add a search feature for subscribed channels (like in the channel groups menu), so that could be nice too.

PR #7429, opened 2025-05-18
fix(player): clear current stream when closing player

FineFindus:
Fixes an issue, where the `currentStream` was not cleared when closing the player. This could lead to an issue, where not all video options where displayed, even if the video was no longer playing.

Ref: https://github.com/libre-tube/LibreTube/commit/f262210ef74f268445f8831bd51f54ac0cb1c019

### DIFF
--- a/app/src/main/java/com/github/libretube/services/AbstractPlayerService.kt
+++ b/app/src/main/java/com/github/libretube/services/AbstractPlayerService.kt
@@ -409,6 +409,8 @@ abstract class AbstractPlayerService : MediaLibraryService(), MediaLibrarySessio
                 mediaLibrarySession = null
             }
 
+            PlayingQueue.clear()
+
             ServiceCompat.stopForeground(this, ServiceCompat.STOP_FOREGROUND_REMOVE)
             stopSelf()
 
--- a/app/src/main/java/com/github/libretube/util/PlayingQueue.kt
+++ b/app/src/main/java/com/github/libretube/util/PlayingQueue.kt
@@ -37,6 +37,7 @@ object PlayingQueue {
         }
         queueJobs.clear()
         queue.clear()
+        currentStream = null
     }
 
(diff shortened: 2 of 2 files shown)


app/src/main/java/com/github/libretube/util/PlayingQueue.kt
```diff
@@ -37,6 +37,7 @@ object PlayingQueue {
         }
         queueJobs.clear()
         queue.clear()
+        currentStream = null;
```
Bnyro, line 40:
```suggestion
        currentStream = null
```

app/src/main/java/com/github/libretube/services/AbstractPlayerService.kt
```diff
@@ -409,6 +409,8 @@ abstract class AbstractPlayerService : MediaLibraryService(), MediaLibrarySessio
                 mediaLibrarySession = null
             }
 
+            PlayingQueue.clearCurrent()
```
Bnyro, line 412: Shouldn't we clear the full queue instead here by calling `PlayingQueue#clear` and additionally change the `clear` method to null the `currentStream`?

FineFindus: Probably, is there any reason why we're not doing this already? Or just that the queue and stream is reset anyway when a new player is created?

Bnyro: > is there any reason why we're not clearing this already? Or just that the queue and stream is reset anyway when a new player is created?
>
There's no reason for that, we should always have cleared the `currentStream`. It's just been overlooked because we resetted it when a new player is created, so it's not been causing any obvious issues before the player service refactorings a few months ago, yeah.

Bnyro (approved):

Issue #7384, opened 2025-05-07
Libretube is crashing trying reproduce a new video


ScriptSlayer0:
### Steps to reproduce

java.lang.RuntimeException: java.lang.reflect.InvocationTargetException
 at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:594)
 at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1013)
Caused by: java.lang.reflect.InvocationTargetException
 at java.lang.reflect.Method.invoke(Native Method)
 at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:584)
 ... 1 more
Caused by: java.lang.reflect.InvocationTargetException
 at java.lang.reflect.Constructor.newInstance0(Native Method)
 at java.lang.reflect.Constructor.newInstance(Constructor.java:343)
 at androidx.room.Room.createViewModel(Unknown Source:7)
 at androidx.lifecycle.ViewModelProvider$AndroidViewModelFactory.create(SourceFile:25)
 at androidx.lifecycle.ViewModelProvider$AndroidViewModelFactory.create(SourceFile:14)
 at androidx.lifecycle.ViewModelProvider$AndroidViewModelFactory.create(SourceFile:2)
 at androidx.lifecycle.SavedStateViewModelFactory.create(SourceFile:17)
 at androidx.lifecycle.ViewModelProvider$Factory.create(SourceFile:5)
 at androidx.work.impl.model.WorkTagDao_Impl.getViewModel$lifecycle_viewmodel_release(Unknown Source:69)
 at retrofit2.Retrofit.getValue(Unknown Source:72)
 at com.github.libretube.ui.fragments.PlayerFragment.getViewModel(Unknown Source:2)
 at com.github.libretube.ui.fragments.PlayerFragment.onViewCreated(Unknown Source:642)
 at androidx.fragment.app.Fragment.performViewCreated(Unknown Source:14)
 at androidx.fragment.app.FragmentStateManager.createView(Unknown Source:265)
 at androidx.fragment.app.FragmentStateManager.moveToExpectedState(Unknown Source:174)
 at androidx.fragment.app.FragmentManager.executeOpsTogether(Unknown Source:1129)
 at androidx.fragment.app.FragmentManager.removeRedundantOperationsAndExecute(Unknown Source:81)
 at androidx.fragment.app.FragmentManager.execSingleAction(Unknown Source:118)
[...]

PR #7434, opened 2025-05-21
feat(Preferences/SponsorBlock): drop `VISIBLE` option 

FineFindus:
Drops the `VISIBLE` option for SponsorBlock, as it provides no additional value and leads to bad UX, as a user is able to see the segment, but able to skip it.
Existing users are migrated to the `MANUAL` option. To achieve this, new supporting code is introduced, that allows us to automatically migrate preferences.

Ref: https://github.com/libre-tube/LibreTube/issues/7384#issuecomment-2860415128
Ref: https://github.com/libre-tube/LibreTube/commit/f5c40157f3e38f019edbbc2568e7c869ffd4f968
Ref: https://github.com/libre-tube/LibreTube/commit/9ceb258881bebb5ee36b6105b491e021930b07bd

### DIFF
--- a/app/src/main/java/com/github/libretube/helpers/PreferenceHelper.kt
+++ b/app/src/main/java/com/github/libretube/helpers/PreferenceHelper.kt
@@ -5,7 +5,6 @@ import android.content.SharedPreferences
 import android.util.Log
 import androidx.core.content.edit
 import androidx.preference.PreferenceManager
-import com.github.libretube.BuildConfig
 import com.github.libretube.LibreTubeApp
 import com.github.libretube.R
 import com.github.libretube.constants.PreferenceKeys
@@ -13,6 +12,13 @@ import com.github.libretube.enums.SbSkipOptions
 
 object PreferenceHelper {
     private val TAG = PreferenceHelper::class.simpleName
+
+    /**
+     * Preference migration from [fromVersion] to [toVersion].
+     */
+    private class PreferenceMigration(
+        val fromVersion: Int, val toVersion: Int, val onMigration: () -> Unit
+    )
 
     /**
      * for normal preferences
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/github/libretube/helpers/PreferenceHelper.kt
```diff
@@ -30,6 +37,32 @@ object PreferenceHelper {
...
+                    val key = "${category}_category"
+                    val stored = getString(key, "visible")
+                    if (stored == "visible") {
+                        putString(key, SbSkipOptions.MANUAL.name.lowercase())
+                    }
+                }
+        }
+
+        // mark as successfully migrated
+        putInt(PreferenceKeys.PREFERENCE_VERSION, BuildConfig.VERSION_CODE)
```
FineFindus, line 63: Not sure if it's better to use the version code here, or a separate constant, to avoid having the migrations run on every startup for builds where the version is not increased on every update, e.g. Debug/Nightly builds?

Bnyro: I think it's better to manually increment the number, because not every version will have a preference migration, so it'd run the whole function body unnecessarily with each update even if there's no migration necessary.

Depending on how much we want to over-engineer it: In theory we could build a simple `PreferenceMigration` class similar to the room migrations and maintain a list of migrations numbered from 1 upwards, see https://github.com/libre-tube/LibreTube/blob/master/app/src/main/java/com/github/libretube/db/AppDatabase.kt. We could then have a `PreferenceMigration(1, 2) { override fun onMigrate() {} }`, that runs the migration.

FineFindus: Implemented in https://github.com/libre-tube/LibreTube/pull/7434/commits/c746647041515ed8e52077753978fb099a53a543.

FineFindus:
Thinking about it, maybe it also makes sense to combine `AUTOMATIC` and `AUTOMATIC_ONCE` as they are nearly identical?

app/src/main/java/com/github/libretube/helpers/PreferenceHelper.kt
Bnyro: Awesome, thank you!

Bnyro:
> Thinking about it, maybe it also makes sense to combine `AUTOMATIC` and `AUTOMATIC_ONCE` as they are nearly identical?
>
We already have the icon to "temporarily" allow SponsorBlock segments, so the "automatic once" behavior could be achieved by temporarily allowing segments, yes.

It would be confusing for the average people if the segment doesn't skip again and again by default, so `AUTOMATIC` should be the one to keep if we remove one.

I however think that it'd make sense to make some kind of community poll or open an issue about it to see if there are many users of this feature, because otherwise people will start complaining about yet another removed feature again...

Bnyro (approved):

PR #7466, opened 2025-05-31
Chore: Remove context parameter from InstanceRepository

SEKI-YUTA:
The `Context` parameter was unused in the `InstanceRepository` constructor and has been removed. This simplifies the class and its instantiation.


Bnyro (approved):
Thank you!

PR #7444, opened 2025-05-22
feat: update ui components to material 3 expressive

Bnyro:
ref https://m3.material.io/blog/building-with-m3-expressive

The audio player so far:
Screenshot_20250523-201544_LibreTube Debug

For other changes fragments, please see the commit history of this PR.

Slightly related to the changes here: https://github.com/libre-tube/LibreTube/pull/7406

## TODO
- I'm not really happy with the library fragment yet.I wanted to use custom Material Design 3 shapes, but idk how to apply them to buttons. Suggestions welcome.

PR #7507, opened 2025-06-28
feat: support edge to edge and bump to sdk35

Bnyro:
Related:
- see https://developer.android.com/develop/ui/views/layout/edge-to-edge

Comments:
- in theory material components (bottom bar, app toolbar, ...) should handle this automatically according to the docs, but they don't?! thus we manually adjust the padding of them as of now
- the settings screen is overlapping with the navigation bar at the bottom, but I guess that's how edge to edge is supposed to work? maybe we need to add some additional bottom padding, after the last settings item, but idk how
- the player in the player fragment now overlaps with the status bar: according to edge to edge principles this should be fine?
- the code for updating the mini player transition (in MainActivity.kt) is horrible, but it's needed because the bottom bar height differs depending on the type of navigation (gestures, 2-buttons, 3-buttons) in the Android settings, so we can't hardcode it

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/activities/MainActivity.kt
+++ b/app/src/main/java/com/github/libretube/ui/activities/MainActivity.kt
@@ -1,8 +1,6 @@
 package com.github.libretube.ui.activities
 
-import android.annotation.SuppressLint
 import android.content.Intent
-import android.content.res.Configuration
 import android.os.Build
 import android.os.Bundle
 import android.view.KeyEvent
@@ -44,7 +42,6 @@ import com.github.libretube.helpers.NavigationHelper
 import com.github.libretube.helpers.NetworkHelper
 import com.github.libretube.helpers.PreferenceHelper
 import com.github.libretube.helpers.ThemeHelper
-import com.github.libretube.helpers.WindowHelper
 import com.github.libretube.ui.base.BaseActivity
 import com.github.libretube.ui.dialogs.ErrorDialog
 import com.github.libretube.ui.dialogs.ImportTempPlaylistDialog
@@ -561,16 +558,6 @@ class MainActivity : BaseActivity() {
         }
 
         return false
-    }
-
-    @SuppressLint("SwitchIntDef")
-    override fun onConfigurationChanged(newConfig: Configuration) {
-        super.onConfigurationChanged(newConfig)
-
-        when (newConfig.orientation) {
-            Configuration.ORIENTATION_PORTRAIT -> WindowHelper.toggleFullscreen(window, false)
-            Configuration.ORIENTATION_LANDSCAPE -> WindowHelper.toggleFullscreen(window, true)
-        }
     }
 
     private fun navigateToBottomSelectedItem(item: MenuItem): Boolean {
--- a/app/src/main/java/com/github/libretube/ui/views/OnlinePlayerView.kt
+++ b/app/src/main/java/com/github/libretube/ui/views/OnlinePlayerView.kt
@@ -22,7 +22,6 @@ import com.github.libretube.enums.PlayerCommand
 import com.github.libretube.extensions.toID
 import com.github.libretube.helpers.PlayerHelper
 import com.github.libretube.helpers.PreferenceHelper
-import com.github.libretube.helpers.WindowHelper
 import com.github.libretube.obj.BottomSheetItem
 import com.github.libretube.services.AbstractPlayerService
 import com.github.libretube.ui.base.BaseActivity
@@ -156,7 +155,6 @@ class OnlinePlayerView(
         this.playerOptions = playerOptions
 
         commonPlayerViewModel.isFullscreen.observe(viewLifecycleOwner) { isFullscreen ->
-            WindowHelper.toggleFullscreen(activity.window, isFullscreen)
             updateTopBarMargin()
 
             binding.fullscreen.isInvisible = PlayerHelper.autoFullscreenEnabled


app/src/main/res/xml/player_scene.xml
```diff
@@ -43,16 +43,6 @@
     </ConstraintSet>
 
     <ConstraintSet android:id="@+id/end">
-        <Constraint
```
Bnyro, line 46: This view doesn't exist anymore since a few years...

app/src/main/java/com/github/libretube/ui/activities/AboutActivity.kt
```diff
@@ -28,6 +30,12 @@ class AboutActivity : BaseActivity() {
         binding.toolbar.setNavigationOnClickListener {
             onBackPressedDispatcher.onBackPressed()
         }
+        // manually apply additional padding for edge-to-edge compatibility
+        ViewCompat.setOnApplyWindowInsetsListener(binding.toolbar) { v, insets ->
+            val systemBars = insets.getInsets(WindowInsetsCompat.Type.systemBars())
+            v.setPadding(0, systemBars.top, 0, 0)
+            insets
+        }
```
FineFindus: Considering nearly the same code is used in many screens, we should move it to the base activity.

FineFindus (changes requested):
> in theory material components (bottom bar, app toolbar, ...) should handle this automatically according to the docs, but they don't?! thus we manually adjust the padding of them as of now

I think it would be worthwhile investigating that, considering the amount of code that is required for working around that.

> the settings screen is overlapping with the navigation bar at the bottom, but I guess that's how edge to edge is supposed to work? maybe we need to add some additional bottom padding, after the last settings item, but idk how

6ec5104197901d74359d60a1687e1c7a4b65f701 :)

> the player in the player fragment now overlaps with the status bar: according to edge to edge principles this should be fine?

We should avoid that, even if it goes against the design principles, as it may be distracting or even impossible to properly watch depending on the device (camera cutout and system icons).

app/src/main/java/com/github/libretube/ui/activities/AboutActivity.kt
Bnyro: The thing is that all activities have different views, so `binding.toolbar` is always a different object. Apart from that we sometimes apply bottom bar and top toolbar padding, and sometimes we only apply top toolbar padding.

I'm not too happy with this either. All I could come up with is a small extension function `View#onSystemBarsMeasured` listener which provides the system bars as a parameter of the anonymous function so that we can save the `val systemBars = insets.getInsets(WindowInsetsCompat.Type.systemBars())` call.

EDIT: okay, I think it makes sense to make a helper method in base activity, I'll do that.

Bnyro:
> I think it would be worthwhile investigating that, considering the amount of code that is required for working around that.
>
I did for about 20 minutes, but I just couldn't figure it out and didn't want to waste any further time with it.

> https://github.com/libre-tube/LibreTube/commit/6ec5104197901d74359d60a1687e1c7a4b65f701 :)
>
Thanks, that's exactly how I planned it to be :)

> We should avoid that, even if it goes against the design principles, as it may be distracting or even impossible to properly watch depending on the device (camera cutout and system icons).
>
Hmm, but a black bar at the top would also look bad. It'd be cool to have some blurred "continuation" of the top of the video, so that it looks natural, but I guess that's too complicated :sweat_smile:

app/src/main/java/com/github/libretube/ui/base/BasePreferenceFragment.kt
```diff
@@ -32,6 +36,21 @@ abstract class BasePreferenceFragment : PreferenceFragmentCompat() {
...
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+        // add bottom padding to the list, to ensure that the last item is not overlapped by the system bars
+        ViewCompat.setOnApplyWindowInsetsListener(listView) { v, insets ->
+            val systemInsets = insets.getInsets(WindowInsetsCompat.Type.systemBars())
+            v.setPadding(
+                v.paddingLeft,
+                v.paddingTop,
+                v.paddingRight,
+                systemInsets.bottom
```
Bnyro:
```suggestion
                v.paddingBottom + systemInsets.bottom
```
That doesn't only apply here: We should probably add it to the existing padding values everywhere instead of assuming a padding of 0 for each view, that could lead to waste of time in debugging stuff in the future.

I'll push the changes.

Bnyro:
> We should avoid that, even if it goes against the design principles, as it may be distracting or even impossible to properly watch depending on the device (camera cutout and system icons).
>
That's the only thing left then, I'm just not 100% sure yet how to handle it. Possibly just using the default background. Probably makes sense to see how other apps do it, if we find some that support edge to edge already.

FineFindus:
I've been testing this over the last few days and discovered 2 more bugs:
- Bottomsheets are a bit too, they partially overlap the video in portrait mode
-  After using auto-rotation the appbar vanishes

FineFindus:
> After using auto-rotation the appbar vanishes

Tracked this down to https://github.com/libre-tube/LibreTube/blob/b255e16a3f561022834826e30fab68d96cfbe8f3/app/src/main/java/com/github/libretube/ui/activities/MainActivity.kt#L124-L141

I'm guessing that the inset update when using full-screen also triggers the listener? A workaround is to just take the `max` of the current padding and the inset, though I would prefer to avoid such workarounds.

> Bottomsheets are a bit too, they partially overlap the video in portrait mode

Thinking about this a bit, the overlap distance seems suspiciously close to the height of the system bar, which the player view does not have.

Bnyro:
> We should avoid that, even if it goes against the design principles, as it may be distracting or even impossible to properly watch depending on the device (camera cutout and system icons).
>
Added the top padding above the player back.

> Thinking about this a bit, the overlap distance seems suspiciously close to the height of the system bar, which the player view does not have.
>
Fixed, you were right. The calculation for the sheet height just didn't consider the top and bottom system bar insets yet, I modified it to do so.

> After using auto-rotation the appbar vanishes
>
The bottom navigation bar, or the toolbar? I just tried it, and the only thing I noticed is that the bottom bar text was a bit cut off. 

If we manage to get this auto-rotation thing fixed, we should be good to merge this finally. I didn't have much time to look into LibreTube recently, sorry for the delay caused by that.

FineFindus:
> The bottom navigation bar, or the toolbar? I just tried it, and the only thing I noticed is that the bottom bar text was a bit cut off.

Both, actually. I've been testing this in my fork together with https://github.com/libre-tube/LibreTube/pull/7444, so maybe that's causing it?

> If we manage to get this auto-rotation thing fixed, we should be good to merge this finally. I didn't have much time to look into LibreTube recently, sorry for the delay caused by that.

Sounds good to me, no worries. https://github.com/libre-tube/LibreTube/commit/9ab6ef9f0da02f605107b4c333a5439bbac84f07 can be used as a workaround.

Bnyro:
> Both, actually. I've been testing this in my fork together with https://github.com/libre-tube/LibreTube/pull/7444, so maybe that's causing it?
>
I'll give it another try tomorrow, possibly yes.

Bnyro:
@FineFindus https://github.com/libre-tube/LibreTube/pull/7507/commits/76595c6cfec5477efebadd280a7603701f7d77e6 should hopefully fix this, at least it does with the device I tested it.

Seems like we still had very old leftover code, that always switched to immersive mode (i.e. hides the system bars) when going to landscape...

app/src/main/java/com/github/libretube/ui/activities/HelpActivity.kt
```diff
@@ -16,6 +16,8 @@ class HelpActivity : BaseActivity() {
         binding.toolbar.setNavigationOnClickListener {
             onBackPressedDispatcher.onBackPressed()
         }
+        // manually apply additional padding for edge-to-edge compatibility
+        updateToolbarTopPadding(binding.toolbar)
```
FineFindus: We can use [`android:fitsSystemWindows="true"`](https://developer.android.com/reference/android/view/View#attr_android:fitsSystemWindows) on the toolbar instead.

app/src/main/java/com/github/libretube/ui/activities/WelcomeActivity.kt
```diff
@@ -29,6 +29,8 @@ class WelcomeActivity : BaseActivity() {
 
         val binding = ActivityWelcomeBinding.inflate(layoutInflater)
         setContentView(binding.root)
+        // manually apply additional padding for edge-to-edge compatibility
+        updateRootViewVerticalPadding(binding.root)
```
FineFindus: Same here.

app/src/main/java/com/github/libretube/ui/fragments/AudioPlayerFragment.kt
```diff
@@ -113,8 +114,17 @@ class AudioPlayerFragment : Fragment(R.layout.fragment_audio_player), AudioPlaye
...
+        val insets =
+            WindowInsetsCompat.toWindowInsetsCompat(activity.window.decorView.rootWindowInsets ?: return)
+        val systemBars = insets.getInsets(WindowInsetsCompat.Type.systemBars())
+        with (binding.audioPlayerMain) {
+            setPadding(
+                paddingLeft,
+                paddingTop + systemBars.top,
+                paddingRight,
+                paddingBottom + systemBars.bottom
+            )
```
FineFindus: May be better to use [`android:fitsSystemWindows="true"`](https://developer.android.com/reference/android/view/View#attr_android:fitsSystemWindows) here as well and add some additional padding in the layout file, to keep the layout in one file.

FineFindus (dismissed):
Can confirm, it's fixed now.

We can actually get the code a bit more concise, but moving the toolbar padding to the layout file.

Screenshot of inset in comments sheet
Another thing I noticed, the comments sheet is also not yet using edge-to-edge.

app/src/main/java/com/github/libretube/ui/activities/WelcomeActivity.kt
Bnyro: We also have to add bottom padding here in order to not overlap the ok/cancel with the bottom bar, so I think it's better to keep it here.

app/src/main/java/com/github/libretube/ui/fragments/AudioPlayerFragment.kt
Bnyro: I've tested doing this, but it only worked sometimes when opening the audio player, and sometimes just didn't apply any at all, no clue why...

I guess we've to set it programmatically for now.

Bnyro:
Finding out about `fitsSystemWindows` earlier could have probably saved me a lot of time :/

PR #7512, opened 2025-06-28
feat: support for setting locale in Android settings for app (A13+)

Bnyro:
Changes:
- dynamically build the list of languages for SDK32 and below, so that we don't have to update the list of languages manually (we have 69 languages on Weblate, but the settings only had ~50 before this PR)
- for SDK33 and above, allow setting the app's language via the Android settings and hide the language preference in the app
- display each language in its locale, so that people that understand the language can directly spot it (not everybody knows their language name in English)
- see https://developer.android.com/guide/topics/resources/app-languages


app/src/main/java/com/github/libretube/helpers/LocaleHelper.kt
```diff
@@ -9,19 +9,22 @@ import com.github.libretube.obj.Country
 import java.util.Locale
 
 object LocaleHelper {
+    @Deprecated("Only used for SDKs below 33 for compatibility. TODO: remove in the future.")
```
FineFindus: I don't think that the TODO is necessary, as the deprecation warning already indicates that I'll be removed.

FineFindus (approved):

Issue #7056, opened 2025-01-31
Long press to adjust playback speed


RzdPai:
### Describe your suggested feature

Long press to adjust playback speed

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [x] I have written a short but informative title.
- [x] I will fill out all of the requested information in this form.

Figim:
**Do you mean fast forward when long pressing the player?** 

**This is actually useful because it is a great option to get the most out of your speed. (You use it when needed) There should be an option in the settings to adjust the speed, regardless of the main speed menu. Quick switch by long pressing the player**

RzdPai:
> **Do you mean fast forward when long pressing the player?** 
> 
> **This is actually useful because it is a great option to get the most out of your speed. (You use it when needed) There should be an option in the settings to adjust the speed, regardless of the main speed menu. Quick switch by long pressing the player**

Yes

PowerUser64:
It would be cool if there was something like pressing and holding to fast-forward, then drag left and right to adjust playback speed. Maybe even rewind could be possible be dragging to a negative speed?

programming-with-ia:
+1

programming-with-ia:
+1

Figim:
This feature has already been implemented.

PR #7523, opened 2025-07-04
Feat(Fast forward video on long press)

Bilal-Ahmad123:


Figim:
https://github.com/libre-tube/LibreTube/issues/7056#issuecomment-3028834268

app/src/main/java/com/github/libretube/ui/views/CustomExoPlayerView.kt
```diff
@@ -383,7 +388,33 @@ abstract class CustomExoPlayerView(
...
+            seekJob?.cancel()
+            handler.removeCallbacks(runnable)
+        }
+        return super.dispatchTouchEvent(event)
+    }
+
+
+    override fun onTouchEvent(event: MotionEvent):Boolean{
+        return super.onTouchEvent(event)
+    }
```
Bnyro: Please integrate this into https://github.com/libre-tube/LibreTube/blob/master/app/src/main/java/com/github/libretube/ui/listeners/PlayerGestureController.kt instead of placing it here directly in the `CustomExoPlayerView`. That's the place where all other gestures are implemented as well.

SE-Bilal-Ahmad:
@Bnyro no problem i will try to resolve the comments ASAP

SE-Bilal-Ahmad:
@Bnyro I am a bit confused what i implemented was video forwarding on long press the way youtube has implemented but after taking a look at your comment "Apart from this, this should only be enabled if PlayerHelper.swipeGesturesEnabled is true." i am confused should it be forwarded and rewind the way vlc does not by long press but by swipe gesture. so should the video be forwared and rewind based on swipe gesture and not long press?

Bnyro:
> @Bnyro I am a bit confused what i implemented was video forwarding on long press the way youtube has implemented but after taking a look at your comment "Apart from this, this should only be enabled if PlayerHelper.swipeGesturesEnabled is true." i am confused should it be forwarded and rewind the way vlc does not by long press but by swipe gesture. so should the video be forwared and rewind based on swipe gesture and not long press?
>
No, I was just looking for a way to allow users to disable it, and I think that the most similar option to this is `swipeGesturesEnabled`. We could also add a new preference, but we currently try to simplify the preference page to make them easier to use, so we want to avoid adding new setings.

So making it a long press action is fine 👍

Bnyro:
Thanks for your work so far!

I've pushed a WIP commit with some changes to use the proper APIs in `PlayerGestureController`, see https://github.com/libre-tube/LibreTube/compare/738eab0423819620c2eca3fe89fec01b16ceb7d3..567d8425545fc78351673b17e8907a47999d3a02.

Long presses are still not 100% correctly detected, I'll see if I can figure that out some other day.

Bnyro:
Rebased, still doesn't work properly because I don't know how to integrate such long presses in our `PlayerGestureController`, two cases:
- we detect a long press, so the fast forward action has to consume the event
- there's no long press but instead it's a swipe gesture to minimize the player, in that case we should not consume the event
However, when the first `ACTION_DOWN` is triggered, we can't know in which case we're in, so we can't know if we have to consume the event or not, and the wrong choice breaks the app.

Figim:
This is how it is on YouTube: Long press to fast forward 2x after 3 seconds. The screen is flickering.

SE-Bilal-Ahmad:
Right now the issue which i am facing which @Bnyro has stated that upon sliding the player it doesnt minimizes the player it breaks the existing functionality

dhanuarf:
Instead of using `GestureDetector`, detecting the long press manually inside `onTouchEvent` would be much more flexible I think? We have much more control in there.

Bnyro:
> Instead of using `GestureDetector`, detecting the long press manually inside `onTouchEvent` would be much more flexible I think? We have much more control in there.
>
Yes, but it's not a clean solution to handle touch events at multiple different places. That makes things more complicated to maintain.

Bnyro:
Thanks to @dhanuarf's work in https://github.com/libre-tube/LibreTube/pull/7913, we can finally implement this now. :tada: 

I've made the changes that were necessary to this PR, so everything works now.

SE-Bilal-Ahmad:
Thank you @Bnyro

brnwlshubh:
@Bnyro Fast Forward is not as smooth as YT

Figim:
> @Bnyro Fast Forward is not as smooth as YT

This is not YT and agree

PR #7526, opened 2025-07-05
Show a blank video when transitioning to a new video, just to inform the user that the new video is being loaded

dhanuarf:
This PR changes the app behaviour on how it loads a new video when the player is already opened.

Previously, when the user clicks on a new video, there is no loading screen or any kind of information to tell the user that that clicked video is currently being loaded, it just load in the background without the user knowing. From ux point of view this feels kinda awkward because when the user click on something, they expect something to happen. On a fast internet connection this is a fine but, on a slow internet, this behaviour become more noticeable (and quite annoying actually). 

On a slow internet,  a video could take up to 2-3 seconds to load and this put the user into awkward position since there is no knowing if the new clicked video is actually being loaded.

This PR also makes the app 'feels' more responsive since again, when the user click on something, they expect something to happen.

Or maybe this is an intentional behaviour?


Bnyro:
> Or maybe this is an intentional behaviour?
>
That behavior change had been caused by a recent refactor that improved the underlying logic of the player when switching videos. Previously we created a completely new player, which took around 500ms longer, so we didn't have the UX issue you described before that change.

TL;DR: No, it's not been intended - Just nobody looked into it because there've been other things with higher priority.

Thanks for the PR, I'll test and review it when I find time :+1:

app/src/main/java/com/github/libretube/services/AbstractPlayerService.kt
```diff
@@ -199,6 +200,11 @@ abstract class AbstractPlayerService : MediaLibraryService(), MediaLibrarySessio
     }
 
     private fun navigateVideo(videoId: String) {
+
+        // show blank video when transitioning just to inform the user that the new video is being loaded
+        exoPlayer?.setMediaItem(MediaItem.fromUri(""))
```
Bnyro: This will probably throw errors in the logs because the `MediaItem` is not playable, right?

Couldn't we just set the media item to `null` or remove all media items instead?

app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
```diff
@@ -307,7 +307,6 @@ class PlayerFragment : Fragment(R.layout.fragment_player), OnlinePlayerOptions {
 
         override fun onMediaMetadataChanged(mediaMetadata: MediaMetadata) {
             super.onMediaMetadataChanged(mediaMetadata)
-
```
Bnyro: Please undo the formatting changes.

app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
```diff
@@ -1089,6 +1095,14 @@ class PlayerFragment : Fragment(R.layout.fragment_player), OnlinePlayerOptions {
             ?.dismiss()
     }
 
+    private fun showVideoInfo(show: Boolean){
+        val visibility = if(show) View.VISIBLE else View.INVISIBLE
+
+        binding.descriptionLayout.visibility = visibility
+        binding.relatedRecView.visibility = visibility
+        binding.playerChannel.visibility = visibility
```
Bnyro:
```suggestion
        binding.descriptionLayout.isInvisible = !show
        binding.relatedRecView.isInvisible = !show
        binding.playerChannel.isInvisible = !show
```

app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
```diff
@@ -1089,6 +1095,14 @@ class PlayerFragment : Fragment(R.layout.fragment_player), OnlinePlayerOptions {
             ?.dismiss()
     }
 
+    private fun showVideoInfo(show: Boolean){
```
Bnyro:
```suggestion
    private fun toggleVideoInfoVisibility(show: Boolean) {
```

app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
```diff
@@ -353,6 +351,14 @@ class PlayerFragment : Fragment(R.layout.fragment_player), OnlinePlayerOptions {
                 e.printStackTrace()
             }
         }
+
+        override fun onMediaItemTransition(mediaItem: MediaItem?, reason: Int) {
+            super.onMediaItemTransition(mediaItem, reason)
+            val id = mediaItem?.mediaMetadata?.extras?.getString(IntentData.videoId)
+            if(id == null){ // unknown video
+                showVideoInfo(false)
+            }
```
Bnyro: I don't think that's the best way to check if the video changed, relying on the `extras` isn't that clean. If applying the suggestions from the comment above, we might be able to just check if `mediaItem != null`.

app/src/main/java/com/github/libretube/services/AbstractPlayerService.kt
dhanuarf: Umm, `setMediaItem()` expects nonnull type, so we cant set it to null. Any other ideas?

app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
dhanuarf: Tbh, the reason I do this is because I just wanted to do a quick hack without editing too many codes to get what I wanted (showing blank video). I also have no experience using android media lib before.  So I just did all this as long as it works lol. Sorry if you dont like it.

Anyway, mediaitem cant be set to null, any other ideas?

app/src/main/java/com/github/libretube/services/AbstractPlayerService.kt
Bnyro: > or remove all media items instead?
>
:)

dhanuarf: I'm not sure how to do that.. ?

Bnyro: `exoPlayer?.removeMediaItems(0, Int.MAX_VALUE)` should do the trick.

According to the documentation, this should cause `onMediaItemTransition(...)` to be called with `mediaItem = null` as argument.

dhanuarf: Oh, thats nice! After fiddling a bit more turns out you can just call `clearMediaItems()`. 

I pushed a new commit, is that okay now?

dhanuarf:
There is still small issue though. If the user minimize the player when it's transitioning to the blank video, the previous video title is still being shown and a replay button also there. But for some reason I cant change their visibility from `toggleVideoInfoVisibility`, are their visibility being tied elsewhere?

Though, this is not a big deal really

Bnyro:
> But for some reason I cant change their visibility from toggleVideoInfoVisibility, are their visibility being tied elsewhere?
>
Yes, when you minimize the player they are made visible, so the changes you do in `toggleVideoInfoVisibility`are just being overriden. Possibly you can just set the title text to an empty string, the other buttons are not that important I guess.

dhanuarf:
ok done

Bnyro:
@FineFindus what's your opinion about this behavior? I'm not 100% sure yet if this is the best UX-way to inform users that a new video is being loaded, but I don't have any better idea.

Possibly we could show a loading indicator (circle) to make it more obvious?

FineFindus:
> 'm not 100% sure yet if this is the best UX-way to inform users that a new video is being loaded, but I don't have any better idea.

From a quick test it is already much better than the current way we handle transitions. I feel like it's generally fine, as it matches with how the first video is loaded when the player is created, though I don't think it would hurt to add another indicator.
Maybe we could force the player to show the loading spinner?

dhanuarf:
>Maybe we could force the player to show the loading spinner?

I've also been looking a way to do that but it seems like the player only show the spinner when it is trying to load something. Since `mediaItem` is empty so the spinner wont show up.

My previous method however, the one where I do `setMediaItem(MediaItem.fromUri(""))`, that method force the player to load a bogus `mediaItem`, so it shows the spinner. But I agree that method is ugly.

Maybe add a spinner on top the `playerView` ? (which is also ugly method)

Bnyro:
> Maybe add a spinner on top the playerView ? (which is also ugly method)
>
Yes, that's what I've done now. We could also add a custom player layout (like https://github.com/google/ExoPlayer/blob/release-v2/library/ui/src/main/res/layout/exo_player_view.xml) to be able to control the progress bar, but that'd be a larger refactor.

Issue #7023, opened 2025-01-25
LibreTube causing bluetooth functionality to repeatedly crash


AMurray-Beep:
### Steps to reproduce

-Have Bluetooth enabled and connected
-Load into LibreTube 
-Play a video

### Expected behavior

-Audio plays through bluetooth-connected device

### Actual behavior

-Bluetooth function rapidly crashes and reboots repeatedly until LibreTube is fully closed

### LibreTube version

v0.27.0

### Android version

Android 9

### Other details

Functionality returns the moment I close the video player. Pausing does not halt the behavior. 

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [x] I have written a short but informative title.
- [x] I will fill out all of the requested information in this form.

alazhar:
issue happens to my device too

ghost-in-the-zsh:
Confirming that this issue is also present in my device, on the same LibreTube version. If it helps, my device is a Galaxy S9+.

Savitor01:
I also experienced it. Bluetooth keeps stopping until leave the tube

Bnyro:
That seems to be a bug in the Androidx media library, we're working around that now.

Please test https://github.com/libre-tube/NightlyBuilds to check if that did the job.

Pittvandewitt:
I have experienced this issue too and updating the media3 library to version 1.5.0 or higher resolves the issue.
However there now seems to be issues regarding insets, probably because of targeting SDK 35, which is needed to upgrade the media3 dependency.

Edit: Above information is false

Bsaubolle:
Can confirm this is still an issue in 0.28.1 (Android 15).

Stez69:
Currently having this issue
Pixel 9 Android 16
0.28.1

Chrigux:
Absolutely still a problem. Pixel 8a/ Grapheen OS (Android15) / Libretube v0.28.1

Pittvandewitt:
This issue happens because a Streams object can be too big to be set as extras here: https://github.com/Bnyro/LibreTube/blob/9d90314137e37b9c43a62a800c1528b06ef4de95/app/src/main/java/com/github/libretube/services/OnlinePlayerService.kt#L224
[...]

PR #7530, opened 2025-07-08
Fix crash when streams size > 1MB

Pittvandewitt:
The maximum parcelable size is 1MB. When this limit is reached, a crash will occur. The metadata is set with a `Streams` parcelable object and is sent over IPC to a place in the Android system where it will make Bluetooth crash. 

This PR reduces the `Streams` object size by setting `audioStreams` and `videoStreams` to empty lists as these lists can grow substantially, reaching over the 1MB size limit.

Fixes https://github.com/libre-tube/LibreTube/issues/7023


FineFindus (changes requested):
Thanks. Please add a short comment explaining why this is necessary, similar to the JSON workaround.

Bnyro (approved):
Thank you!

PR #7539, opened 2025-07-14
feat: remove trending

FineFindus:
Removes the Trending section, layout and all other references to trending, due to YouTube removing the trending page.

Ref: https://support.google.com/youtube/thread/356702168?sjid=13404251444768258109-EU

It is worth noting that, without the Trending section, the Home screen looks rather empty. This could be an opportunity to redesign it more extensively.


app/src/main/java/com/github/libretube/helpers/PreferenceHelper.kt
```diff
@@ -51,6 +58,16 @@ object PreferenceHelper {
...
+        /**
+         * Removes the trending layout options, due to YouTube dropping the trending section.
+         *
+         * See https://support.google.com/youtube/thread/356702168?sjid=13404251444768258109-EU
+         */
+        PreferenceMigration(1, 2) { context ->
+            val defaultItems = context.resources.getStringArray(R.array.homeTabItemsValues);
+            val items = getStringSet(PreferenceKeys.HOME_TAB_CONTENT, defaultItems.toSet())
+            putStringSet(PreferenceKeys.HOME_TAB_CONTENT, items.minus("trending"))
+        },
```
FineFindus, line 70: This migration is technically not necessary, as the settings component seem to handle the removed item fine, however to avoid unexpected surprises and just to clean up, I think it is a good idea to remove them nonetheless.

Figim:
Isn't this the main thing on YouTube? 

Why did they do this? 😲

FineFindus:
> Is there an ETA when they will remove it? As of now, the trends still work normally as it seems and I couldn't find any date in the linked post.

Yes, next [Monday, July 21](https://support.google.com/youtube/answer/7239739?hl=en).


> I think we should keep it as long as it works.

Happy to wait that one week :)

> Yes, that could certainly be done within the M3 Expressive migration. I don't have any real ideas though, so help with design ideas is appreciated.

I don't have anything concrete either. We could ask if there is a designer in the community, who would like to help out.
One option may be to just drop the Home screen entirely, as except for the 'Continue Watching' section, all other sections are already easily available. But I don't think that's an easy decision to remove it.

Bnyro:
> Yes, next [Monday, July 21](https://support.google.com/youtube/answer/7239739?hl=en).
>
> Happy to wait that one week :)
>
Fair enough :)

> I don't have anything concrete either. We could ask if there is a designer in the community, who would like to help out.
One option may be to just drop the Home screen entirely, as except for the 'Continue Watching' section, all other sections are already easily available. But I don't think that's an easy decision to remove it.
>
It would be easier to fill the home tab if the NewPipe Extractor supported YouTube's explore pages, i.e. Gaming, Live, News, ...
I'm not sure what to do with the home tab either, two options
- a very good redesign that allows to view important content at one glance
- removing it: possibly causing lots of complaints from users

I think it makes sense to attempt the first way and try to find somebody with a bit talent on designing.

Bnyro:
Created https://github.com/libre-tube/LibreTube/issues/7542 for tracking the home tab.

FineFindus:
> Two leftovers in the HomeViewModel here: https://github.com/libre-tube/LibreTube/pull/7539/files#diff-db24ef33473dd7cc78366197c1b39e5f2ad5c438b33873c73f694dcb661e9ce2L34-L42

Removed in https://github.com/libre-tube/LibreTube/pull/7539/commits/ff9ea64a3804027676ed9fa46755c010f965a86b. Will squash that into https://github.com/libre-tube/LibreTube/pull/7539/commits/7914250c915ccc7d819ca6562252a1a240d244bc when merging :)

FineFindus:
The trending frontend has now been removed, though the backend still works. Not if we want to merge this now or wait (i.e. until the backend is killed, or we have a replacement in https://github.com/libre-tube/LibreTube/issues/7542)?

Bnyro:
> The trending frontend has now been removed, though the backend still works. Not if we want to merge this now or wait (i.e. until the backend is killed, or we have a replacement in #7542)?
>
I'd wait either until the API for it is fully gone or we have a replacement ready, otherwise the home tab is quite empty.

Figim:
> > The trending frontend has now been removed, though the backend still works. Not if we want to merge this now or wait (i.e. until the backend is killed, or we have a replacement in #7542)?
> 
> I'd wait either until the API for it is fully gone or we have a replacement ready, otherwise the home tab is quite empty.

The trend doesn't get deleted, the name changes. 
It looks like it will be on YouTube charts 

like this: https://charts.youtube.com/

PR #7553, opened 2025-07-19
fix: collapse description when loading new video

FineFindus:
Fixes a regression, where the description layout could still be expanded, when loading a new video, leading to only a blank layout being displayed.

Ref: https://github.com/libre-tube/LibreTube/pull/7526

Blank screen when loading new video with expanded description


Bnyro:
Good catch :+1:

PR #7574, opened 2025-07-27
feat: enable edge-to-edge in bottom sheets

FineFindus:
The BottomSheetDialog internally adds an inset, leading to this inset being displayed instead of the recyclerview for chapters and comments. Whilst it is possible to remove, we need to manually handle the inset padding on the lists.

Ref: https://github.com/libre-tube/LibreTube/pull/7507#pullrequestreview-3058291871


FineFindus:
Considering how many sheets would need the inset to be re-applied, I think it is easier to just removed for those two only. However, that leads to some of the sheets having a pretty big inset, not sure if that is something that we want?

Bottomsheet with bigger bottom inset

Bnyro:
> However, that leads to some of the sheets having a pretty big inset, not sure if that is something that we want?
>
Hmm, for the player option sheet it seems fine. However, in the playing queue sheet, it looks a bit out of place looking at the bottom action bar we got there.

screen

FineFindus:
Should be better now.

Bnyro (approved):

PR #7579, opened 2025-07-28
feat: change 'playback pitch' option to semitone mode

dhanuarf:
This PR adds semitone increment and decrement when changing pitch in playback option.

The UI is slightly modified, padding is added on the left side just to create some kind of "separation" between `Playback speed` and `Pitch`.

Slight downside: `stepSize` in `Pitch slider` is removed to accommodate the changes.

https://github.com/user-attachments/assets/93e2aa8f-52d3-4503-aee4-192fbc1a7ee1


dhanuarf:
**Force pushed changes**: add ability to change semitone via text input

<br>
Screenshot_20250729-100643_LibreTube Debug

dhanuarf:
**Another forced push changes:** slightly improve UI <br>

Screenshot_20250729-134855_LibreTube Debug

dhanuarf:
**Changes** [6a31e64](https://github.com/libre-tube/LibreTube/pull/7579/commits/6a31e641a968fc9fc6d1b4c36a669a9ae443ac9b): <br>

Screenshot_20250730-073606_LibreTube Debug

Bnyro:
I'm not sure if it's a good idea to keep the old pitch setting because it might be confusing to have different ways to set the pitch.

Perhaps using semitones is more intuitive, so at least in my opinion we should drop the old preset values and change the slider to use semitone steps as well instead of using the pitch percentage.

Opinions @dhanuarf @FineFindus ?

app/src/main/java/com/github/libretube/ui/sheets/PlaybackOptionsSheet.kt
```diff
@@ -39,17 +45,57 @@ class PlaybackOptionsSheet(
 
         binding.speed.value = player.playbackParameters.speed
         binding.pitch.value = player.playbackParameters.pitch
+
+        val semitone = playbackPitchToSemitone(binding.pitch.value).round(2)
+        binding.semitoneEditText.setText(semitone.toString())
+        binding.semitoneEditText.setOnEditorActionListener { editText, actionId, event ->
+            if (actionId == EditorInfo.IME_ACTION_DONE) {
+                val enteredValue = editText.text.toString().toFloat()
+                if(enteredValue > 12.0f || enteredValue < -12.0f){
```
Bnyro:
```suggestion
                if (enteredValue.absoluteValue > 12.0f) {
```

app/src/main/java/com/github/libretube/ui/sheets/PlaybackOptionsSheet.kt
```diff
@@ -68,17 +114,55 @@ class PlaybackOptionsSheet(
...
+        onChange()
+    }
+
+    private fun semitoneToPlaybackPitch(semitone: Float): Float{
+        return SEMITONE_RATIO.pow(semitone).toFloat()
+    }
+
+    private fun playbackPitchToSemitone(playbackPitch: Float): Float{
+        return ln(playbackPitch) / ln(SEMITONE_RATIO)
+    }
```
Bnyro: Please add a reference to the source of the formula, it's probably nothing everybody knows out of their mind :)

app/src/main/java/com/github/libretube/ui/sheets/PlaybackOptionsSheet.kt
```diff
@@ -68,17 +114,55 @@ class PlaybackOptionsSheet(
...
+
+    private fun incrementDecrementSemitoneBy(value: Float){
+        val playbackPitch = player.playbackParameters.pitch
+        var semitone = playbackPitchToSemitone(playbackPitch)
+        semitone = semitone + value
+
+        // clamping, cant do more than an octave
+        semitone = semitone.coerceAtMost(12.0f)
+        semitone = semitone.coerceAtLeast(-12.0f)
```
Bnyro:
```suggestion
        semitone = semitone.coerce(-12.0f, 12.0f)
```

app/src/main/res/values/strings.xml
```diff
@@ -304,6 +304,8 @@
     <string name="change_instance">Change instance</string>
     <string name="suggest_change_instance">Loading is taking more than usual. Consider changing instance</string>
     <string name="playback_pitch">Pitch</string>
+    <string name="playback_pitch_semitone">Semitone</string>
+    <string name="playback_pitch_semitone_error_input">Value must be between -12.0 (minimum) and 12.0 (maximum)</string>
```
Bnyro: Please don't put the numbers inside the string, instead use `%1$.2f` and pass the bounds with `getString(R.string.playback_pitch_semitone, 12f)`.

See https://www.w3schools.com/java/ref_string_format.asp.

app/src/main/java/com/github/libretube/ui/sheets/PlaybackOptionsSheet.kt
```diff
@@ -68,17 +114,55 @@ class PlaybackOptionsSheet(
...
+        return ln(playbackPitch) / ln(SEMITONE_RATIO)
+    }
+
+    private fun incrementDecrementSemitoneBy(value: Float){
+        val playbackPitch = player.playbackParameters.pitch
+        var semitone = playbackPitchToSemitone(playbackPitch)
+        semitone = semitone + value
+
+        // clamping, cant do more than an octave
+        semitone = semitone.coerceAtMost(12.0f)
```
Bnyro: Since `12.0f` appears quite often, please put it into a constant instead of using 12.0f everywhere to make it easier to overview.

FineFindus:
> Perhaps using semitones is more intuitive, so at least in my opinion we should drop the old preset values and change the slider to use semitone steps as well instead of using the pitch percentage.

Agreed, having two UI elements for editing the same value is pretty bad UX. I think having the slider is better, considering the value is discrete and bounded (i.e. it cannot be freely edited).

On a more general note, what's the use case for changing the pitch? I assume it would be for manual adjustment when changing speed, but maybe that should better be done automatically?

app/src/main/java/com/github/libretube/ui/sheets/PlaybackOptionsSheet.kt
```diff
@@ -39,17 +45,57 @@ class PlaybackOptionsSheet(
...
+                if(enteredValue > 12.0f || enteredValue < -12.0f){
+                    editText.error = context?.getString(R.string.playback_pitch_semitone_error_input)
+                    return@setOnEditorActionListener false
+                }
+                changePlaybackPitch(semitoneToPlaybackPitch(enteredValue))
+                clearEditTextFocusAndHideKeyboard()
+
+                true
+            }
+            else false
```
FineFindus: As the else is just a `false`, this should be changed to an early return

dhanuarf:
> Perhaps using semitones is more intuitive, so at least in my opinion we should drop the old preset values and change the slider to use semitone steps as well instead of using the pitch percentage.

I do agree of removing the preset. But changing the slider back to discrete with semitone steps, I personally prefer the slider to be continuous. But this is your project so the decision is up to you.

The reason I prefer continuous slider is because for more accurate pitch tuning. There are some songs that are tuned differently than the typical tuning. For example, standard tuning is typically tuned with the note A4 at 440hz. But there are some songs where the note A4 is tuned to 432hz. So if your musical instrument is tuned with A4=440hz, the songs sound a bit out of tune.

With a semitone calculator like https://www.omnicalculator.com/other/semitone, you can calculate the semitone difference and then enter the value in the app.

But this feature probably will only useful for some music nerds, not general people.
So yes, in most app, for general use to most people, changing the semitone in +-1 increment is good enough.

So again, the decision is yours, I'll just follow along.

dhanuarf:
> On a more general note, what's the use case for changing the pitch? I assume it would be for manual adjustment when changing speed, but maybe that should better be done automatically?


I'm not sure about general cases, but at least in my case, as someone who plays musical instrument, I like changing the playback pitch when I want to play along to some songs that have different tuning than my instrument. Or maybe there are others who want to sing along but the song is out of their vocal range so they adjust the playback pitch to fit their vocal range better.

Or maybe there are some others who are sensitive to high-pitched sound, so they tune down the playback pitch for comfortable listening. 

Or maybe there are some people who just like to change the playback pitch just for gigs and giggles, I'm not sure.

But yeah, the ability to change the playback speed is enough for generic usage.

app/src/main/java/com/github/libretube/ui/sheets/PlaybackOptionsSheet.kt
dhanuarf: Umm sorry I'm not sure what to do here. I'm a bit lacking when it comes to programming terms.

Should i just replace `else false` with `return@setOnEditorActionListener false` instead?

Edit: ok nevermind

app/src/main/java/com/github/libretube/ui/sheets/PlaybackOptionsSheet.kt
dhanuarf: Well, I also happened to just copied and pasted them from google's search top result :).

dhanuarf:
[deacd72](https://github.com/libre-tube/LibreTube/pull/7579/commits/deacd725f9086d860af6df1bcc434a676ed83571) update:
The playback pitch slider value is now in semitones, but the slider itself is still in continuous mode. 

Also, should we change back the playback speed to decimal values or keep them at percentages?

Bnyro:
> But changing the slider back to discrete with semitone steps, I personally prefer the slider to be continuous. But this is your project so the decision is up to you.
>
I agree, semitones might be too big steps for some people, so a continuous slider makes sense.

> Also, should we change back the playback speed to decimal values or keep them at percentages?
>
I prefer decimal values because that's the way most other apps do it as well, so it's more intuitive.

app/src/main/java/com/github/libretube/ui/sheets/PlaybackOptionsSheet.kt
FineFindus:
```kotlin
            if (actionId != EditorInfo.IME_ACTION_DONE) {
               return@setOnEditorActionListener false
            }

            val enteredValue = editText.text.toString().toFloat()
            if(enteredValue > 12.0f || enteredValue < -12.0f){
                editText.error = context?.getString(R.string.playback_pitch_semitone_error_input)
                return@setOnEditorActionListener false
            }
            changePlaybackPitch(semitoneToPlaybackPitch(enteredValue))
            clearEditTextFocusAndHideKeyboard()

            true
```

No, something like this. This way, the happy path (the code that actual does the logic) is not as indented and easier to read.

dhanuarf:
[5a364ff](https://github.com/libre-tube/LibreTube/pull/7579/commits/5a364ff4905cb722792157319d73f17b2b707b53) changes the minus&plus button into musical flat&sharp button.

I tried to use the unicode version of these symbols to simplify things, but they look ugly. So I replace them to use icons.

Screenshot_20250802-100443_LibreTube Debug

Bnyro (approved):
Thank you!

dhanuarf:
Actually, just right now I was just looking for a way to turn the slider into centered like in here: https://m3.material.io/components/sliders/overview, but cant seem find a way to do that.

Bnyro:
> Actually, just right now I was just looking for a way to turn the slider into centered like in here: https://m3.material.io/components/sliders/overview, but cant seem find a way to do that.
>
That's only possible after upgrading to MD3 Expressive afaik, see the WIP PR https://github.com/libre-tube/LibreTube/pull/7444

dhanuarf:
> That's only possible after upgrading to MD3 Expressive afaik, see the WIP PR #7444

Ahh, so thats why.

I was so confused because this centered slider update is done in Dec 2023, and the material library we are using is `1.12.0` which is released in [May 2024](https://mvnrepository.com/artifact/com.google.android.material/material), got me wondered why wont this work

PR #7592, opened 2025-08-02
feat: support for trending categories (gaming, trailers, ...)

Bnyro:
- replaces / closes https://github.com/libre-tube/LibreTube/pull/7539

Planned for the homepage redesign with MD3 Expressive:
- a dropdown to let users change the trending category on the homepage, e.g. to gaming

This doesn't yet really fit into the UI imo, that's why I want to incorporate that with the Expressive PR.


Bnyro:
It might make sense to hide the tab layout when there's only one category available (so in case Piped is used)?

app/src/main/res/layout/fragment_trends.xml
```diff
@@ -1,35 +1,17 @@
...
-        android:layout_height="wrap_content"
-        app:layout_constraintBottom_toBottomOf="parent"
-        app:layout_constraintLeft_toLeftOf="parent"
-        app:layout_constraintRight_toRightOf="parent"
-        app:layout_constraintTop_toTopOf="parent" />
-
-    <com.github.libretube.ui.views.CustomSwipeToRefresh
-        android:id="@+id/home_refresh"
+    <com.google.android.material.tabs.TabLayout
+        android:id="@+id/tab_layout"
```
FineFindus, line 9: Tabbar with split text

This should be scrolling, to avoid layout issues (and preferable have the same font size for all items).

app/src/main/java/com/github/libretube/ui/models/TrendsViewModel.kt
```diff
@@ -20,17 +21,25 @@ import retrofit2.HttpException
 import java.io.IOException
 
 class TrendsViewModel : ViewModel() {
-    val trendingVideos = MutableLiveData<List<StreamItem>>()
+    val trendingVideos = MutableLiveData<Map<TrendingCategory, List<StreamItem>>>()
```
FineFindus, line 25: We should initialize this with the trending content with already have from the home page, to avoid the additional request.

app/src/main/java/com/github/libretube/api/NewPipeMediaServiceRepository.kt
```diff
@@ -240,16 +240,26 @@ fun String.toListLinkHandler() = with(JsonHelper.json.decodeFromString<TabData>(
...
 
-    override suspend fun getTrending(region: String): List<StreamItem> {
+    private val trendingCategories = mapOf(
+        TrendingCategory.TRENDING to "Trending",
+        TrendingCategory.GAMING to "trending_gaming",
+        TrendingCategory.TRAILERS to "trending_movies_and_shows",
+        TrendingCategory.PODCASTS to "trending_podcasts_episodes",
+        TrendingCategory.MUSIC to "trending_music",
+        TrendingCategory.LIVE to "live"
+    )
```
FineFindus, line 255:
```suggestion
    private val trendingCategoryIds = mapOf(
        TrendingCategory.TRENDING to "Trending",
        TrendingCategory.GAMING to "trending_gaming",
        TrendingCategory.TRAILERS to "trending_movies_and_shows",
        TrendingCategory.PODCASTS to "trending_podcasts_episodes",
        TrendingCategory.MUSIC to "trending_music",
        TrendingCategory.LIVE to "live"
    )
```

I'm guessing the values come from NewPipe? Probably a good idea to also link to the source of them, in case the change in the future.

FineFindus (changes requested):
When changing tabs by directly jumping to a tab some of the tabs around are loaded as well. It would be great if we could avoid the additional requests.

> It might make sense to hide the tab layout when there's only one category available (so in case Piped is used)?

I think that depends on how Piped intends to deal with them. With how it currently is I would just hide them, considering that the default trending backend will be fully removed at some point and the piped backend hasn't seen any commit in a while. If they update, I agree we should hide the tabs.

Figim:
Newpipe does not support : "music videos"
There is more content here and it is sorted. 
https://charts.youtube.com/charts/TopVideos/am/daily

app/src/main/res/layout/fragment_trends.xml
Bnyro: Good catch, I didn't know it's possible to make tab layouts scrollable

app/src/main/java/com/github/libretube/ui/models/TrendsViewModel.kt
Bnyro: I agree, but that'd be better off in a different PR, since this one focuses only on adding the new kiosk tabs

app/src/main/java/com/github/libretube/api/NewPipeMediaServiceRepository.kt
Bnyro: They're split across multiple files in NewPipe, I just took them by reading the output of `KioskExtractor.kioskList` now.

Bnyro: Added a reference to the directory where they're listed.

Bnyro:
> When changing tabs by directly jumping to a tab some of the tabs around are loaded as well. It would be great if we could avoid the additional requests.
>
We're interrupting the previous requests now, I doubt there's any better way (except if we wait a few 100ms before doing the initial request).

app/src/main/java/com/github/libretube/ui/models/TrendsViewModel.kt
Bnyro: Done in https://github.com/libre-tube/LibreTube/pull/7600

Issue #7594, opened 2025-08-02
Source error on all videos/audio in recent nightly builds


akashsriramganapathy:
### Steps to reproduce

1. Install the latest nightly build after `#16659846103`
2. Open any video or audio
3. Observe that playback fails with a “source error”

### Expected behavior

Video and audio should play normally, as they do in build `#16659846103`.

### Actual behavior

Every video/audio shows a **“source error”** and fails to play.

### LibreTube version

latest nightly

### Android version

Android 15

### Other details

* The issue does **not occur** in this working build:
  [https://github.com/libre-tube/LibreTube/actions/runs/16659846103](https://github.com/libre-tube/LibreTube/actions/runs/16659846103)
* Likely a regression introduced after this version.
* Let me know if logs or reproduction info is needed.

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [x] I have written a short but informative title.
- [x] I will fill out all of the requested information in this form.

Figim:
Is full local mode enabled?

Figim:
[02_08-21-43-21_160.log](https://github.com/user-attachments/files/21559852/02_08-21-43-21_160.log)

it was strange

Akash-Sriram:
> Is full local mode enabled?

Yes

Akash-Sriram:
> [02_08-21-43-21_160.log](https://github.com/user-attachments/files/21559852/02_08-21-43-21_160.log)
> 
> it was strange

you can test the linked action build works fine

Figim:
The Media3 update broke the player.  https://github.com/libre-tube/LibreTube/pull/7585

This is a bug in the media3 repo. The pr is there. Until there is a new version, you can go back to the old one 

https://github.com/androidx/media/issues/2688

PR #7595, opened 2025-08-02
revert: downgrade media3

FineFindus:
Fixes a regression, caused by the recent media3 update, where no video can be played due to a source error. This is only intended as a stop-gap solution until we are able to figure out the root cause behind the issue.

This reverts commit bb46f847da53e826ef710e7cf768e94f68f2e0c2.

Closes: https://github.com/libre-tube/LibreTube/issues/7594


Figim:
They fixed it. It will be in the next version. 
I referenced the issue on libretube. But wouldn't it be scary?

FineFindus:
Thanks, just saw that myself.

> But wouldn't it be scary?

In what sense would that be scary?

Figim:
> Thanks, just saw that myself.
> 
> > But wouldn't it be scary?
> 
> In what sense would that be scary?

Isn't media3 a google-owned repo? 

I found out later.

FineFindus:
> Isn't media3 a google-owned repo?

And? We have linked against their issues multiple times already, it makes things much easier to track. They won't complain because of this.

Figim:
> > Isn't media3 a google-owned repo?
> 
> And? We have linked against their issues multiple times already, it makes things much easier to track. They won't complain because of this.

Okey😊 thanks

PR #7597, opened 2025-08-03
feat: separate 'clone' and 'create' when creating a new playist

dhanuarf:
This PR separate `clone playlist` and `create playlist` into  separate view via spinner.

Having 2 options in the same UI while the user only able to proceed with just one of them is kind of bad UX, I think?  

Before changes:
before-pl-dialog-Screenshot_20250803-065659_LibreTube Debug

After changes:
After-pl-dialog-Screenshot_20250803-064904_LibreTube Debug

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/dialogs/CreatePlaylistDialog.kt
+++ b/app/src/main/java/com/github/libretube/ui/dialogs/CreatePlaylistDialog.kt
@@ -4,6 +4,7 @@ import android.app.Dialog
 import android.os.Bundle
 import android.widget.Toast
 import androidx.core.os.bundleOf
+import androidx.core.view.isVisible
 import androidx.fragment.app.DialogFragment
 import androidx.fragment.app.setFragmentResult
 import androidx.lifecycle.lifecycleScope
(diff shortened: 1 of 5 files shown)


Bnyro (commented):
Perhaps the spinner option to create a new playlist should be renamed to something like `New private playlist`? That probably makes it more clear, "Create playlist" is already in the title and thus we currently have the same text twice.

dhanuarf:
> Perhaps the spinner option to create a new playlist should be renamed to something like `New private playlist`? That probably makes it more clear, "Create playlist" is already in the title and thus we currently have the same text twice.

Yeah I was trying to avoid to create or modify any strings, so I just used whats available.

dhanuarf:
[c86f442](https://github.com/libre-tube/LibreTube/pull/7597/commits/c86f442c3cf83ff0fc9ddc22248ddebe092ef763):

Screenshot_20250803-175201_LibreTube Debug

app/src/main/java/com/github/libretube/ui/dialogs/CreatePlaylistDialog.kt
```diff
@@ -22,6 +23,17 @@ class CreatePlaylistDialog : DialogFragment() {
     override fun onCreateDialog(savedInstanceState: Bundle?): Dialog {
         val binding = DialogCreatePlaylistBinding.inflate(layoutInflater)
 
+        binding.createPlaylistSpinner.items = listOf<String>(
+            getString(R.string.clonePlaylist),
+            getString(R.string.createNewPlaylist))
```
Bnyro:
```suggestion
        binding.createPlaylistSpinner.items = listOf<String>(
            getString(R.string.createNewPlaylist),
            getString(R.string.clonePlaylist)
        )
```
The default value should be the first item, not the second one. It feels a bit confusing otherwise.

PR #7601, opened 2025-08-03
feat: disable video track when minimized while PiP is off

dhanuarf:
This PR disables video when the user has `Settings>Player>Behaviour when minimized` configured to `None`.

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
+++ b/app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
@@ -914,12 +914,12 @@ class PlayerFragment : Fragment(R.layout.fragment_player), OnlinePlayerOptions {
         // check whether the screen is on
         val isInteractive = requireContext().getSystemService<PowerManager>()!!.isInteractive
 
-        // disable video stream since it's not needed when screen off
-        if (!isInteractive) {
-            // disable the autoplay countdown while the screen is off
+        // disable video stream since it's not needed when screen off or when PiP is not enabled
+        if (!isInteractive || !PlayerHelper.pipEnabled) {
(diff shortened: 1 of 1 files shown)


Figim:
> As a note: IIRC SABR doesn't allow audio-only streaming, so this will break when we eventually migrate to it. Though it shouldn't be a blocker for now.

In that case, wouldn't it make more sense to close this PR?

dhanuarf:
> As a note: IIRC SABR doesn't allow audio-only streaming, so this will break when we eventually migrate to it. Though it shouldn't be a blocker for now.

So all features related to audio-only stream will be removed from the app?

Figim:
So how does YT Music work?

FineFindus:
> So all features related to audio-only stream will be removed from the app?

No idea, probably depends on how the situation is when we switch (i.e. if a method is found). But no one is actively working on an implementation.

> So how does YT Music work?

They have their own API.

Figim:
> > So all features related to audio-only stream will be removed from the app?
> 
> No idea, probably depends on how the situation is when we switch (i.e. if a method is found). But no one is actively working on an implementation.
> 
> > So how does YT Music work?
> 
> They have their own API.

ytdlp and freetube works. https://github.com/yt-dlp/yt-dlp/pull/13515

Other: https://github.com/etwoo/youtube-unthrottle/commit/76b48d0a2eba29b7dae21cf185d1ec279361c506

**They removed "video only." "audio only" will remain.** 

https://github.com/LuanRT/googlevideo/commit/3ab569c63724b4efe264b1fe47ff8344e5d85fbe

**fix(sabr): Video-only playback is no longer supported
What a bummer...

At least they can't remove audio-only, since it's still needed for YouTube Music. :)**

Yt music has a function to change yt video to audio. This function is Premium  😂

PR #7603, opened 2025-08-03
feat: disable 'Add to playlist' button and spinner when there is no playlist available

dhanuarf:
Slight UI impovement to avoid confusion.

Dialog before:
beforeAddPlScreenshot_20250803-181107_LibreTube Debug

Dialog after:
afterAddPlScreenshot_20250803-183242_LibreTube Debug

This PR also includes **fix**:
-After the user deleted all playlists, if they try to add a video to a playlist by triggering this dialog, the app crashes.

### DIFF
--- a/app/src/main/res/layout/dialog_add_to_playlist.xml
+++ b/app/src/main/res/layout/dialog_add_to_playlist.xml
@@ -1,7 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:app="http://schemas.android.com/apk/res-auto"
-    xmlns:tools="http://schemas.android.com/tools"
     android:layout_width="wrap_content"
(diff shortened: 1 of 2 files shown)


app/src/main/res/layout/dialog_add_to_playlist.xml
```diff
@@ -14,5 +14,25 @@
         android:layout_marginTop="8dp"
         app:icon="@drawable/ic_playlist_add"
         tools:ignore="RtlSymmetry" />
+    <LinearLayout
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:layout_marginHorizontal="24dp"
+        android:padding="8dp"
+        android:gravity="right|center_vertical"
```
Bnyro:
```suggestion
        android:gravity="end|center_vertical"
```
In order to also support `right to left (rtl)` screens, we have to use `end` instead of `right`.

PR #7615, opened 2025-08-07
fix: player not paused when switching to other apps while pauseOnQuit is on

### LINKED ISSUE #7614: Minimizing Setting Ignored When Switching Apps
### Steps to reproduce

1. Go to Settings -> Player
2. Set `Behavior When Minimized` to 'Pause'
3. Open a Video
4. Switch to another App 

### Expected behavior

The video should pause, as it does when you switch to the Home Screen. Also because you're minimizing LibreTube when switching in between apps.

### Actual behavior

The video does not pause, as indicated by the audio, which keeps playing. 

### LibreTube version

0.28.1

### Android version

15

### Other details

_No response_

### Acknowledgements

dhanuarf:
This PR fixes #7614

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
+++ b/app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
@@ -923,8 +923,12 @@ class PlayerFragment : Fragment(R.layout.fragment_player), OnlinePlayerOptions {
             setVideoTrackTypeDisabled(true)
         }
 
-        // pause player if screen off and setting enabled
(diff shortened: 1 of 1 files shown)


dhanuarf:
> Please move (or integrate this) to the check below that already pauses the screen if the screen is off, to avoid having got sections (in slightly different places) that handle pausing the player.
> 

Would this be ok?
```
        val shouldPausePlayer =
            (isInteractive && PlayerHelper.pauseOnQuit) ||
                    (!isInteractive && PlayerHelper.pausePlayerOnScreenOffEnabled)

        if (shouldPausePlayer) {
            playerController.pause()
        }
```

PR #7623, opened 2025-08-08
fix: appbar and bottom navigation paddings not properly applied

### LINKED ISSUE #7616: App bar sometimes gets cropped after exiting from fullscreen
### Steps to reproduce

1. Watch a video
2. Enter fullscreen
3. Exit fullscreen via back button or via gesture swipe down (if the setting is on)
4. Check the app bar

### Expected behavior

App bar should not be cropped

### Actual behavior

App bar is cropped.

Physical device (Android 13)
Image

Emulator (Android 15)
Image

For some reason this issue does not happen if I exit the fullscreen via the dedicated button (bottom-right of the player).
This issue happens after implementing the `edge-to-edge` on api 35. Also for some other reason, on my physical device only the app bar gets cropped, while on the emulator both app bar and bottom navigation get cropped .
 

### LibreTube version

Latest commit

### Android version

Android 13 (physical device), Android 15 (emulator)

### Other details

I checked the code in `MainActivity`, `onApplyWindowInsets` does get called, the padding also get applied, but the layout does not get "relayout-ed". I tried to add `post{ requestLayout() }` after `setPadding`, the problem doesnt appear anymore. Wrapping `setPadding` inside `Dispatchers.Main` also works.

Maybe the system is busy relayouting the UI while exiting fullscreen (because orientation change, updating window params and stuff)  so it's not ready yet, so we have to wait a little to update the padding?

Any thoughts on this?
Thanks.

### Acknowledgements

dhanuarf:
There is a possibility that the paddings are not properly applied when exiting from player's fullscreen.
Fixes #7616

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/activities/MainActivity.kt
+++ b/app/src/main/java/com/github/libretube/ui/activities/MainActivity.kt
@@ -119,22 +119,31 @@ class MainActivity : BaseActivity() {
         // manually apply additional padding for edge-to-edge compatibility
         // see https://developer.android.com/develop/ui/views/layout/edge-to-edge
         binding.root.onSystemInsets { _, systemBarInsets ->
-            with(binding.appBarLayout) {
-                setPadding(
-                    paddingLeft,
-                    systemBarInsets.top,
-                    paddingRight,
-                    paddingBottom
-                )
-            }
-            with (binding.bottomNav) {
-                setPadding(
-                    paddingLeft,
-                    paddingTop,
-                    paddingRight,
-                    systemBarInsets.bottom
-                )
-            }
+            // there's a possibility that the paddings are not being applied properly when
+            // exiting from player's fullscreen. Adding OnGlobalLayoutListener serves as
+            // a workaround for this issue
+            binding.root.viewTreeObserver.addOnGlobalLayoutListener(object:
+            ViewTreeObserver.OnGlobalLayoutListener{
+                override fun onGlobalLayout() {
+                    with(binding.appBarLayout) {
+                        setPadding(
+                            paddingLeft,
+                            systemBarInsets.top,
+                            paddingRight,
+                            paddingBottom
+                        )
+                    }
+                    with(binding.bottomNav) {
+                        setPadding(
+                            paddingLeft,
+                            paddingTop,
+                            paddingRight,
+                            systemBarInsets.bottom
+                        )
+                    }
+                    binding.root.viewTreeObserver.removeOnGlobalLayoutListener(this)
+                }
+            })
         }
         // manually update the bottom bar height in the mini player transition
         binding.bottomNav.addOnLayoutChangeListener { _, _, _, _, _, _, _, _, _ ->


app/src/main/java/com/github/libretube/ui/activities/MainActivity.kt
```diff
@@ -119,22 +119,31 @@ class MainActivity : BaseActivity() {
...
-                    paddingRight,
-                    systemBarInsets.bottom
-                )
-            }
+            // there's a possibility that the paddings are not being applied properly when
+            // exiting from player's fullscreen. Adding OnGlobalLayoutListener serves as
+            // a workaround for this issue
+            binding.root.viewTreeObserver.addOnGlobalLayoutListener(object:
+            ViewTreeObserver.OnGlobalLayoutListener{
+                override fun onGlobalLayout() {
```
Bnyro, line 127:
```suggestion
                override fun onGlobalLayout() {
                    val insets = getSystemInsets()
```

Please remove the outer `binding.root.onSystemInsets` to avoid nested listeners, see https://github.com/libre-tube/LibreTube/blob/888d95e3328d95c846907ed4b11e8b2b5e0eee17/app/src/main/java/com/github/libretube/ui/extensions/OnSystemInsets.kt#L18 for the `getSystemInsets` method.

Bnyro:
Please format your code using the inbuilt tools in Intellij IDEA / Android Studio if possible, I'd like to keep the formatting consistent everywhere across our source code :)

app/src/main/java/com/github/libretube/ui/activities/MainActivity.kt
dhanuarf: Wouldnt it make this useless since we have `removeOnGlobalLayoutListener` at the end?

And if we dont remove the listener it will be called quite a lot of times. I tried it, paddings become inconsistently applied.

dhanuarf: If we need to avoid nested listener then we can just use `post{ requestLayout() }`. At least it works on my side.

But before that, do you also able to reproduce this issue on your side? Or am I the only one who's experiencing this? lol

Bnyro: > But before that, do you also able to reproduce this issue on your side? Or am I the only one who's experiencing this?
>
Yeah, I was able to reproduce it (although it only happens sometimes).

> Wouldnt it make this useless since we have removeOnGlobalLayoutListener at the end?
>
At least for me it seems to fix the issue (have been trying to reproduce it 5+ times). Please test it as well though.

dhanuarf: I'm sorry, I'm not sure I follow you here.

If we remove `onSystemInsets` listener, `onGlobalLayout` will only be called once in `onCreate` because we have we have `removeOnGlobalLayoutListener` at the end, so the paddings will only be applied once, hence why I said it's useless. I think I'm missing something here :/

Bnyro: Oh right, I thought the activity would be recreated on every rotation, though that"s not the case anymore when entering fullscreen...

Bnyro (approved):

PR #7635, opened 2025-08-15
fix(player ui): seek and skip buttons appear at opposite side in rtl mode

dhanuarf:
I also notice the skip segment button appears at the left side in rtl mode. The video is progressing from left to right so the skip segment button on the left side seems kind of unintuitive,  I think? I wanted to hardcode it on the right side but realized it's been there for quite some time so it might make some people unhappy if it's moved so I just left it untouched.

### DIFF
--- a/app/src/main/res/layout/double_tap_overlay.xml
+++ b/app/src/main/res/layout/double_tap_overlay.xml
@@ -3,7 +3,8 @@
     xmlns:app="http://schemas.android.com/apk/res-auto"
     android:layout_width="match_parent"
     android:layout_height="match_parent"
-    android:baselineAligned="false">
+    android:baselineAligned="false"
+    android:layoutDirection="ltr">
 
     <!-- double tap rewind btn -->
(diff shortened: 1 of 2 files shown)


Figim:
it was visible on the right side . Latest debug

Screenshot_20250815_060326_LibreTube Debug

dhanuarf:
> it was visible on the right side . Latest debug
> 
> ![Screenshot_20250815_060326_LibreTube Debug]

No, I was talking about when the device is in RTL mode

Bnyro:
> I also notice the skip segment button appears at the left side in rtl mode. The video is progressing from left to right so the skip segment button on the left side seems kind of unintuitive, I think? I wanted to hardcode it on the right side but realized it's been there for quite some time so it might make some people unhappy if it's moved so I just left it untouched.
>
Perhaps we should keep it like that while there are no complaints because people probably already got used to it like that.

Bnyro (approved):
Thanks!

PR #7637, opened 2025-08-15
fix(sponsorblock): segment time not properly parsed if exceeds 1 hour

dhanuarf:
Previously when trying to submit a sponsorblock segment, if the time exceed 1 hour, the time conversion resulted in a wrong value and caused the segment to be submitted in the wrong place of the video

### DIFF
--- a/app/src/main/java/com/github/libretube/util/TextUtils.kt
+++ b/app/src/main/java/com/github/libretube/util/TextUtils.kt
@@ -58,7 +58,7 @@ object TextUtils {
     fun String.parseDurationString(): Float? = parseTimeString(this)
 
     private fun parseTimeString(timeString: String): Float? {
-        if (timeString.isDigitsOnly()) return timeString.toLongOrNull()?.toFloat()
+        if (timeString.all { it.isDigit() }) return timeString.toLongOrNull()?.toFloat()
 
         if (timeString.all { it.isDigit() || ",.:".contains(it) }) {
             var secondsTotal = 0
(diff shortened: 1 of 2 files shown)


dhanuarf:
After thinking about it for a while, I dont think this is a refactor. But I'm not sure if it's a fix or a feat

FineFindus:
> After thinking about it for a while, I dont think this is a refactor. But I'm not sure if it's a fix or a feat

It's a `fix` (it's not adding the ability to submit hours, but _fixing_ the wrong parsing).

dhanuarf:
> It's a `fix` (it's not adding the ability to submit hours, but _fixing_ the wrong parsing).

Yeah. I was made confused by my own sentence

app/build.gradle.kts
```diff
@@ -57,6 +57,11 @@ android {
         jvmTarget = "17"
     }
 
+    // ref: https://developer.android.com/training/testing/local-tests#error
+    testOptions {
```
Bnyro, line 61: Why's this needed here? The tests went through before without that change iirc?

app/src/main/java/com/github/libretube/util/TextUtils.kt
```diff
@@ -79,6 +82,12 @@ object TextUtils {
                     secondsScoped *= 10
                     secondsScoped += char.digitToInt()
                 } else if (char == ':') {
+                    if(isInHours){
```
Bnyro, line 85:
```suggestion
                    if (isInHours) {
```
nit: formatting

PR #7649, opened 2025-08-28
fix: image loading issues when scrolling too fast on a list

dhanuarf:
Load images using coil's extension, by passing the `target` image directly. The nice thing about this is they automatically cancel the requests for you somehow. 

Their website says coil automatically cancel/pause requests for you, but doesnt say 'when' it does that. The only answer I got was from google's ai result, which says that it cancels the requests when the view gets detached.

### DIFF
--- a/app/src/main/java/com/github/libretube/helpers/ImageHelper.kt
+++ b/app/src/main/java/com/github/libretube/helpers/ImageHelper.kt
@@ -3,14 +3,13 @@ package com.github.libretube.helpers
 import android.content.Context
 import android.graphics.Bitmap
 import android.graphics.Color
-import android.graphics.drawable.Drawable
 import android.net.Uri
 import android.widget.ImageView
 import androidx.core.net.toUri
 import coil3.ImageLoader
-import coil3.asDrawable
 import coil3.disk.DiskCache
 import coil3.disk.directory
+import coil3.load
 import coil3.network.okhttp.OkHttpNetworkFetcherFactory
 import coil3.request.CachePolicy
 import coil3.request.ImageRequest
@@ -107,11 +106,13 @@ object ImageHelper {
             if (urlToLoad.startsWith(HTTP_SCHEME) && !isCached(urlToLoad)) return
         }
 
-        getImageWithCallback(target.context, urlToLoad) { result ->
-            // set the background to white for transparent images
-            if (whiteBackground) target.setBackgroundColor(Color.WHITE)
-
-            target.setImageDrawable(result)
+        target.load(urlToLoad) {
+            listener(
+                onSuccess = { _, _ ->
+                    // set the background to white for transparent images
+                    if (whiteBackground) target.setBackgroundColor(Color.WHITE)
+                }
+            )
         }
     }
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/github/libretube/helpers/ImageHelper.kt
```diff
@@ -107,12 +106,10 @@ object ImageHelper {
             if (urlToLoad.startsWith(HTTP_SCHEME) && !isCached(urlToLoad)) return
         }
 
-        getImageWithCallback(target.context, urlToLoad) { result ->
-            // set the background to white for transparent images
-            if (whiteBackground) target.setBackgroundColor(Color.WHITE)
+        // set the background to white for transparent images
+        if (whiteBackground) target.setBackgroundColor(Color.WHITE)
```
Bnyro: The problem is that if we set the background color here directly, the channel avatars (e.g. in the results) will have a white background instead of the gray placeholder while the image is still loading.

Bnyro:
> Their website says coil automatically cancel/pause requests for you, but doesnt say 'when' it does that. The only answer I got was from google's ai result, which says that it cancels the requests when the view gets detached.
>
Where did you find that info? Looking at their source code I can't find anything about it though.

The `ImageView.load` method calls `.target(imageView)`, which internally just uses the `GenericViewTarget`: https://github.com/coil-kt/coil/blob/c0d23d4c6c02960fd09210fc2f31c1d6d7b1a990/coil/src/androidMain/kotlin/coil3/singletonImageLoaders.android.kt#L42

It basically uses `GenericViewTarget` form https://github.com/coil-kt/coil/blob/c0d23d4c6c02960fd09210fc2f31c1d6d7b1a990/coil-core/src/androidMain/kotlin/coil3/target/GenericViewTarget.kt#L20, which doesn't react to any view changes.

dhanuarf:
> Where did you find that info? Looking at their source code I can't find anything about it though.

Exactly. But somehow the image loading is not glitching anymore after this changes (at least on my case).  So somehow they cancel the request somewhere.

We can add `CoilUtils.dispose(view)` on every viewholder's `onViewRecycled()` if we wanna make sure though.

dhanuarf:
ok, found it here: https://github.com/coil-kt/coil/blob/main/coil-core/src/androidMain/kotlin/coil3/request/RequestDelegate.android.kt

It checks if the view is attached. If it isnt, it disposes the request. https://github.com/coil-kt/coil/blob/main/coil-core/src/androidMain/kotlin/coil3/request/ViewTargetRequestManager.kt#L93

Bnyro:
> ok, found it here: https://github.com/coil-kt/coil/blob/main/coil-core/src/androidMain/kotlin/coil3/request/RequestDelegate.android.kt
> 
> It checks if the view is attached. If it isnt, it disposes the request. https://github.com/coil-kt/coil/blob/c0d23d4c6c02960fd09210fc2f31c1d6d7b1a990/coil-core/src/androidMain/kotlin/coil3/request/ViewTargetRequestManager.kt#L81
>
Perhaps we can implement an observer that listens for the `onDestroy` lifecycle event and cancels the Coil-Job ourselves? That way we can still properly handle the white background, as mentioned in https://github.com/libre-tube/LibreTube/pull/7649#discussion_r2311884131.

app/src/main/java/com/github/libretube/helpers/ImageHelper.kt
dhanuarf: ok, it's wrapped inside a `onSuccess` listener now

PR #7662, opened 2025-09-06
fix: keyboard showing up in player screen

dhanuarf:
Now the search bar's focus is being cleared every time the player is maximized.


app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
```diff
@@ -617,6 +617,11 @@ class PlayerFragment : Fragment(R.layout.fragment_player), OnlinePlayerOptions {
                     commonPlayerViewModel.setSheetExpand(true)
                     mainMotionLayout.progress = 0F
                     changeOrientationMode()
+
+                    // clear search bar focus
```
FineFindus:
```suggestion
                    // clear search bar focus to avoid keyboard popups
```

nit: the code already shows that we clear the focus, so this comment it is redundant. It would be better to instead explain why this is done :)

PR #7673, opened 2025-09-12
feat: show toolbar when scrolling up in subscriptions feed for better accessibility

dhanuarf:
Now we dont need to scroll all the way up just to filter the list.

~~Other changes:~~
~~- Toolbar stays expanded when the subscription list is opened. The reason I made it like this is because I think the subscription list is still part/menu of the toolbar, not an actual content of this fragment, so the toolbar should stay open with the list ( I think?)~~

~~- Added collapse icon:~~

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/fragments/SubscriptionsFragment.kt
+++ b/app/src/main/java/com/github/libretube/ui/fragments/SubscriptionsFragment.kt
@@ -91,7 +91,7 @@ class SubscriptionsFragment : DynamicLayoutManagerFragment(R.layout.fragment_sub
 
         // Determine if the child can scroll up
         binding.subRefresh.setOnChildScrollUpCallback { _, _ ->
-            !isAppBarFullyExpanded
+            !isAppBarFullyExpanded || binding.subFeed.canScrollVertically(-1)
         }
 
         binding.subRefresh.isEnabled = true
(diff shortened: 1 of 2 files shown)


Bnyro:
Perhaps the changes no longer really makes sense due to the subscriptions redesign with https://github.com/libre-tube/LibreTube/pull/7444?

dhanuarf:
> Perhaps the changes no longer really makes sense due to the subscriptions redesign with #7444?

I think the part where the toolbar is shown when scrolling up (which is the main point of this PR) still makes sense. I can revert the other changes in the bullet points above.

dhanuarf:
Or maybe I can close this PR and let you handle the rest to prevent merge conflict? Since the patch for that is just changing `layout_scrollFlags` to `scroll|enterAlways`, and few more code change to prevent `SwipeRefreshLayout` from glitching.

**Edit**: Oh sorry, I didnt know the branch has been merged already

Bnyro (commented):
I like the idea in general, unfortunately this doesn't seem to work properly yet.

Sometimes the toolbar stays at the top for a short time when scrolling down (which is perhaps not what we want) and it doesn't seem to appear when scrolling up :/

(Tested on Android 16)

Bnyro (approved):
Awesome, thank you!

Issue #7602, opened 2025-08-03
Error if the content country is missing


Figim:
### Steps to reproduce

1. Select Germany as  location. 
2.  Podcast not supported in Germany 
3. Now go to that section. 
4. Error

### Expected behavior

Show me a message instead of an error. 

### Actual behavior

Error

org.schabi.newpipe.extractor.exceptions.ParsingException: Could not get channel name
	at org.schabi.newpipe.extractor.services.youtube.YoutubeChannelHelper.lambda$getChannelName$9(YoutubeChannelHelper.java:546)
	at org.schabi.newpipe.extractor.services.youtube.YoutubeChannelHelper$$ExternalSyntheticLambda10.get(D8$$SyntheticClass:0)
	at java.util.Optional.orElseThrow(Optional.java:403)
	at org.schabi.newpipe.extractor.services.youtube.YoutubeChannelHelper.getChannelName(YoutubeChannelHelper.java:546)
	at org.schabi.newpipe.extractor.services.youtube.extractors.kiosk.YoutubeDesktopBaseKioskExtractor.getName(YoutubeDesktopBaseKioskExtractor.java:65)
	at org.schabi.newpipe.extractor.services.youtube.extractors.kiosk.YoutubeTrendingPodcastsEpisodesExtractor.getName(YoutubeTrendingPodcastsEpisodesExtractor.java:6)
	at org.schabi.newpipe.extractor.kiosk.KioskInfo.getInfo(KioskInfo.java:66)
	at com.github.libretube.api.NewPipeMediaServiceRepository.getTrending(NewPipeMediaServiceRepository.kt:266)
	at com.github.libretube.ui.models.TrendsViewModel$fetchTrending$2$response$1.invokeSuspend(TrendsViewModel.kt:39)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:101)
	at kotlinx.coroutines.internal.LimitedDispatcher$Worker.run(LimitedDispatcher.kt:113)
	at kotlinx.coroutines.scheduling.TaskImpl.run(Tasks.kt:89)
	at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:589)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:823)
[...]

PR #7675, opened 2025-09-12
fix(Home): hide Trending category

FineFindus:
As the `Trending` trending category is finally fully shutdown (i.e. the API doesn't work for me for the last few days anymore), hide the category from users and migrate any existing users to another category.

Also fixes an issue, where the wrong category items could be cached from trending.

Happy to also use another choice for the default category value.

### DIFF
--- a/app/src/main/java/com/github/libretube/api/NewPipeMediaServiceRepository.kt
+++ b/app/src/main/java/com/github/libretube/api/NewPipeMediaServiceRepository.kt
@@ -246,14 +246,14 @@ class NewPipeMediaServiceRepository : MediaServiceRepository {
 
     // see https://github.com/TeamNewPipe/NewPipeExtractor/tree/dev/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/kiosk
     private val trendingCategories = mapOf(
-        TrendingCategory.TRENDING to "Trending",
         TrendingCategory.GAMING to "trending_gaming",
         TrendingCategory.TRAILERS to "trending_movies_and_shows",
         TrendingCategory.PODCASTS to "trending_podcasts_episodes",
         TrendingCategory.MUSIC to "trending_music",
         TrendingCategory.LIVE to "live"
     )
-    override fun getTrendingCategories(): List<TrendingCategory> = trendingCategories.keys.toList()
+    override fun getTrendingCategories(): List<TrendingCategory> =
+        trendingCategories.keys.toList()
 
     override suspend fun getTrending(region: String, category: TrendingCategory): List<StreamItem> {
         val kioskList = NewPipeExtractorInstance.extractor.kioskList
--- a/app/src/main/java/com/github/libretube/helpers/PreferenceHelper.kt
+++ b/app/src/main/java/com/github/libretube/helpers/PreferenceHelper.kt
@@ -7,6 +7,7 @@ import androidx.core.content.edit
 import androidx.preference.PreferenceManager
 import com.github.libretube.LibreTubeApp
 import com.github.libretube.R
+import com.github.libretube.api.TrendingCategory
 import com.github.libretube.constants.PreferenceKeys
 import com.github.libretube.enums.SbSkipOptions
 import com.github.libretube.helpers.LocaleHelper.getDetectedCountry
@@ -51,6 +52,10 @@ object PreferenceHelper {
                         putString(key, SbSkipOptions.MANUAL.name.lowercase())
                     }
                 }
+        },
+        PreferenceMigration(1, 2) {
+            // select a random category as the new value
+            putString(PreferenceKeys.TRENDING_CATEGORY, TrendingCategory.LIVE.name)
         },
     )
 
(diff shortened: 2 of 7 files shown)


Bnyro:
> As the Trending trending category is finally fully shutdown (i.e. the API doesn't work for me for the last few days anymore), hide the category from users and migrate any existing users to another category.
>
Weird, it still works fine for me as of today :eyes:

app/src/main/java/com/github/libretube/api/NewPipeMediaServiceRepository.kt
```diff
@@ -253,7 +253,8 @@ class NewPipeMediaServiceRepository : MediaServiceRepository {
         TrendingCategory.MUSIC to "trending_music",
         TrendingCategory.LIVE to "live"
     )
-    override fun getTrendingCategories(): List<TrendingCategory> = trendingCategories.keys.toList()
+    override fun getTrendingCategories(): List<TrendingCategory> =
+        trendingCategories.keys.drop(1).toList()
```
Bnyro: Perhaps we should fully remove the `TRENDING` enum entry instead of dropping it manually everywhere? Likely, we should just disable trends for Piped (which is the only user of the `TRENDING` category after these changes if I see that correctly)

app/src/main/java/com/github/libretube/helpers/PreferenceHelper.kt
```diff
@@ -52,6 +53,10 @@ object PreferenceHelper {
                     }
                 }
         },
+        PreferenceMigration(1, 2) {
+            // select a random category as the new value
+            putString(PreferenceKeys.TRENDING_CATEGORY, TrendingCategory.entries.drop(1).random().name)
```
Bnyro: I'd default to `LIVE`, but that's just personal preference.

Anyways I'd like to avoid making it random, it should perhaps be consistent for all users.

app/src/main/java/com/github/libretube/ui/fragments/HomeFragment.kt
```diff
@@ -201,12 +201,13 @@ class HomeFragment : Fragment(R.layout.fragment_home) {
...
-    private fun showTrending(streamItems: List<StreamItem>?) {
-        if (streamItems == null) return
+    private fun showTrending(trends: Pair<TrendingCategory, List<StreamItem>>?) {
+        if (trends == null) return
+        val (category, streamItems) = trends
 
         // cache the loaded trends in the [TrendsViewModel] so that the trends don't need to be
         // reloaded there
-        trendsViewModel.setStreamsForCategory(TrendingCategory.TRENDING, streamItems)
+        trendsViewModel.setStreamsForCategory(category, streamItems)
```
Bnyro, line 210: Good catch!

FineFindus:
> Weird, it still works fine for me as of today 👀

Knowing YouTube that's either because it's a gradual roll-out or they're A/B testing whether it can be rolled-out, but I guess in either case it doesn't hurt to have this ready :)

Bnyro:
It makes sense to merge the cache fix already before creating the release, I don't have a preference when we merge the other changes though (I'm fine with merging it now already too).

dhanuarf:
Instead of deleting the category, can we just merge all category (gaming, music, live, etc) into this `Trends` category locally without relying on the extractor? And merge the list in random order if possible?

FineFindus:
No, as that would require us to make `n` requests instead of just `1`. We would also need to deterministically merge the categories (as otherwise they would be shuffled on every reload which is contrary to how a trend section should behave, from the users perspective)

app/src/main/java/com/github/libretube/ui/fragments/HomeFragment.kt
Bnyro, line 210: Related: I noticed we currently only set the first 10 trending streams to the view model (because `showTrending` only gets passed the 10 first items).

Ref: https://github.com/libre-tube/LibreTube/pull/7695

dhanuarf: Sorry for chiming in, I accidentally set the category to podcast, and my podcast is not showing (related: #7602). Since the returned `StreamItems` is null, this section is kept hidden. And the only way to change the category back is via `trendingTV`, and since it's hidden, I cant change it back.

Maybe kept `trendingTV` shown and put a placeholder in `trendingRV` when the returned list is empty?

Bnyro:
Trends are gone for me too now since yesterday, so I agree with merging this now.

app/src/main/java/com/github/libretube/ui/fragments/HomeFragment.kt
Bnyro: I agree, sounds like a good idea.

FineFindus:
Oddly enough they have re-appeared for me, so I guess it's A/B-Test/gradual rollout. Though I don't oppose merging, since it's still something that will be removed in the long-run.

Bnyro:
@dhanuarf
> Maybe kept trendingTV shown and put a placeholder in trendingRV when the returned list is empty?
>
Please open a new issue about that.

PR #7681, opened 2025-09-13
feat: little UI fixes

Elolcol:


app/src/main/java/com/github/libretube/ui/adapters/SearchResultsAdapter.kt
```diff
@@ -99,7 +99,6 @@ class SearchResultsAdapter(
 
             channelContainer.isGone = item.uploaderAvatar.isNullOrEmpty()
             channelName.text = item.uploaderName
-            ImageHelper.loadImage(item.uploaderAvatar, channelImage, true)
```
Bnyro, line 102: What's the reason for removing the uploader avatar? I think that should be undone.

app/src/main/res/layout/fragment_player.xml
```diff
@@ -91,27 +90,24 @@
...
                     app:shapeAppearance="@style/CircleImageView" />
 
                 <LinearLayout
                     android:layout_width="0dp"
-                    android:layout_height="wrap_content"
-                    android:layout_gravity="start|center_vertical"
-                    android:layout_marginStart="5dp"
-                    android:layout_marginEnd="8dp"
+                    android:layout_height="match_parent"
+                    android:layout_marginHorizontal="5dp"
```
Bnyro, line 110: Why did you remove the center alignment here? It looks quite weird if the channel name + subscription count aren't centered vertically.

app/src/main/res/layout/exo_styled_player_control_view.xml
```diff
@@ -309,11 +310,11 @@
                 android:layout_width="match_parent"
                 android:layout_height="match_parent"
                 android:paddingTop="6dp"
-                app:bar_height="2dp"
+                app:bar_height="4dp"
```
Bnyro, line 314: The bar looks a bit strange with 4dp height in my opinion. I already considered changing it before, but I think it looks much better with 2dp.

app/src/main/res/layout/exo_styled_player_control_view.xml
```diff
@@ -309,11 +310,11 @@
...
                 app:played_color="?attr/colorSecondary"
-                app:scrubber_color="?attr/colorSecondary"
+                app:scrubber_color="?attr/colorOnSecondary"
```
Bnyro, line 316: Having a different color here than in the bar looks a bit strange in my opinion.

app/src/main/java/com/github/libretube/ui/adapters/SearchResultsAdapter.kt
Elolcol: because it seems too redundant and overloaded

app/src/main/res/layout/fragment_player.xml
Elolcol: Yes, it may look a little strange, but in the fragment_channel isn't vertically aligned either.

app/src/main/res/layout/exo_styled_player_control_view.xml
Elolcol: maybe, but its more for ux than ui, or stay in 2dp but add something like yt to change the time using all the timeline player for more easy to use.

app/src/main/res/layout/exo_styled_player_control_view.xml
Elolcol: is the same color of the timeline in the videos views

app/src/main/res/layout/fragment_player.xml
Bnyro: Well, in my opinion both should be aligned.

app/src/main/java/com/github/libretube/ui/adapters/SearchResultsAdapter.kt
Bnyro: Nevertheless I'd prefer to keep it :)

Elolcol:
Yes, when I saw that I had added more than 10 commits, I knew it wasn't going to be accepted, but I decided to continue so that I could have the changes in a branch. I really shouldn't have made so many commits in a single pull request.

I'm going to make a PR to gather all the changes in the dialogs, as I think it's a change for the better in the UI. Still, I would like the other changes to be taken into account, and if there are any changes that might be liked, to be able to convert them into a PR.

I still have a few small changes to commit.

Elolcol:
Well, that's all.

There are some issues with the dragHandle in the about and help sections, and with the border radius in the queue and chapters, and i don't know how to change the style in the time selector in the notification settings.

Bnyro:
Thanks for starting to split the commits into separate PRs, that makes it much easier for me! :)

Issue #7699, opened 2025-09-16
Subscription Bottom Sheet: filter list by currently selected group and show total list number


dhanuarf:
### Describe your suggested feature

I'm not sure this should be an issue report or a feature request, since these are things that were previously there.

The changes made in b762f64 removed the ability to filter the subscription list based on what group currently selected. The bottom sheet always show all subscriptions regardless selected group. I'm not sure this was intentional or not, but I prefer it the way before, because currently the only way to see which channel inside the group is by clicking edit button, and then see which channels are being checked.

Maybe also add a title at the top of the sheet to indicate which group the list is currently on. Also it would be nice to have total number of (filtered) subscriptions after the title.

Something like:
<img src="https://github.com/user-attachments/assets/7929237c-09a4-4c4c-a6f1-d78e3843af29" width="auto" height="420" />

--
Thanks.

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [x] I have written a short but informative title.
- [x] I will fill out all of the requested information in this form.

Bnyro:
I agree. It's not been intended to remove that, it's just an oversight.

Perhaps we should just add the filter chips and the edit button from the feed fragment to the channel subscriptions bottom sheet as well.

Bnyro:
Merged your PR, great job!

PR #7702, opened 2025-09-18
feat(subscriptions sheet): filter list by currently selected group

dhanuarf:
Also added edit button for the selected group.

Previously discussed in #7699.
The sheet contains only the selected group, because I dont think duplicating all the group chips here is a good idea. This way we can add edit button for that selected group.

I also thought of adding delete button here, but still not sure about it.

https://github.com/user-attachments/assets/4d28e299-4625-4e2a-9704-64efbd726370


dhanuarf:
Also about the group button chips in `SubscriptionFragment`, wouldn't it make more sense if the `longPress` action is a dropdown menu containing "edit" and "delete" instead of playing video?

**Edit:** Actually never mind. I just thought maybe some people prefer it this way.

PR #7725, opened 2025-09-29
Fix: floating buttons

Elolcol:
Now the floating buttons don't cut off the shadows, and now the history button isn't covered by the mini player.

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/fragments/WatchHistoryFragment.kt
+++ b/app/src/main/java/com/github/libretube/ui/fragments/WatchHistoryFragment.kt
@@ -4,8 +4,10 @@ import android.content.res.Configuration
 import android.os.Bundle
 import android.os.Parcelable
 import android.view.View
+import android.view.ViewGroup.MarginLayoutParams
 import androidx.core.view.isGone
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/github/libretube/ui/fragments/WatchHistoryFragment.kt
```diff
@@ -54,8 +56,10 @@ class WatchHistoryFragment : DynamicLayoutManagerFragment(R.layout.fragment_watc
         _binding = FragmentWatchHistoryBinding.bind(view)
         super.onViewCreated(view, savedInstanceState)
 
-        commonPlayerViewModel.isMiniPlayerVisible.observe(viewLifecycleOwner) {
-            _binding?.watchHistoryRecView?.updatePadding(bottom = if (it) 64f.dpToPx() else 0)
```
Bnyro, line 58: This line also needs to be kept, to reproduce:
- watch a few videos to fill your watch history
- start playing a video and navigate to your watch history
- you can't fully see the last video item now because it's overlapped by the player

screen

Bnyro (approved):
Thanks!

PR #7735, opened 2025-10-03
feat(player): move screenshot writing task to background thread

dhanuarf:
Also added snackbar to notify user.

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
+++ b/app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
@@ -111,6 +111,8 @@ import com.github.libretube.util.OnlineTimeFrameReceiver
 import com.github.libretube.util.PlayingQueue
 import com.github.libretube.util.TextUtils
 import com.github.libretube.util.TextUtils.toTimeInSeconds
+import com.google.android.material.snackbar.Snackbar
+import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.runBlocking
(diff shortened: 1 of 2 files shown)


dhanuarf:
>  * having a snackbar to show that a screenshot is being created isn't really necessary, as that is already obvious when the user is asked to choose a file and generally saving the file is relatively quick

This depends on the quality and/or the content of what is being captured. PNG is quite costly when it comes to complex/detailed picture. It could take up to a few seconds depending on the device. I think it would be awkward if we only notify the user after it's succesfully saved, because the wait time could be several seconds.

But I'm fine removing it though.

>  * we should use a toast here, instead of a snackbar, considering the user does not have any useful interaction, similar to how we show a toast when a SponsorBlock segment is skipped.

The reason I use snackbar is just to match the material design. Toast is displayed by the system framework, so different android version have different looks. So if we really follow the material design I think we need start to use snackbar from now on?

>  Alternatively I'm also open to adding a share button to the snackbar.

I think that's a great idea.

Bnyro (approved):
Thank you both, I agree that a share button justified adding the snackbar here 👍

app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
```diff
@@ -392,11 +394,30 @@ class PlayerFragment : Fragment(R.layout.fragment_player), OnlinePlayerOptions {
...
+                screenshotBitmap = null
 
-            screenshotBitmap = null
+                withContext(Dispatchers.Main) {
+                    Snackbar.make(
+                        requireView(),
+                        R.string.screenshot_is_saved,
+                        Snackbar.LENGTH_LONG
+                    ).apply {
+                        setDuration(2500)
```
FineFindus:
```suggestion
                        2500
                    ).apply {
```

The duration can be set directly.

app/src/main/res/values/strings.xml
```diff
@@ -468,6 +468,7 @@
     <string name="behavior_when_minimized">Behavior when minimized</string>
     <string name="external_player">External player</string>
     <string name="screenshot">Screenshot</string>
+    <string name="screenshot_is_saved">Screenshot is saved</string>
```
FineFindus: This sounds a bit unnatural, maybe "Screenshot has been saved" or just "Screenshot saved"?

FineFindus (approved):
Added two more small nitpicks, but otherwise LGTM.

FineFindus:
Thanks!

PR #7739, opened 2025-10-09
Provide service type via startForeground compat method

Pittvandewitt:
Android API 29 and 34 have different ways of starting a foreground service. They require providing a service type. This PR provides the service type via the compat method.

I think the download services can also be implemented through workmanager, for easier network monitoring and thread management. What are your thoughts on this?

### DIFF
--- a/app/src/main/java/com/github/libretube/services/DownloadService.kt
+++ b/app/src/main/java/com/github/libretube/services/DownloadService.kt
@@ -4,9 +4,11 @@ import android.app.NotificationManager
 import android.app.PendingIntent.FLAG_CANCEL_CURRENT
 import android.app.PendingIntent.FLAG_UPDATE_CURRENT
 import android.content.Intent
+import android.content.pm.ServiceInfo
 import android.net.ConnectivityManager
 import android.net.Network
 import android.os.Binder
+import android.os.Build
 import android.os.IBinder
 import android.util.Log
 import android.util.SparseBooleanArray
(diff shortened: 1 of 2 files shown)


Bnyro:
> I think the download services can also be implemented through workmanager, for easier network monitoring and thread management. What are your thoughts on this?
>
I agree that this would make sense, however are you sure that we can still display the download progress in the downloads fragment this way? Is there also a way to bind a background worker to the UI (similar to the service binder we currently use)?

app/src/main/java/com/github/libretube/services/DownloadService.kt
```diff
@@ -518,7 +522,14 @@ class DownloadService : LifecycleService() {
...
             .setGroupSummary(true)
 
-        startForeground(NotificationId.DOWNLOAD_IN_PROGRESS.id, summaryNotificationBuilder.build())
+        ServiceCompat.startForeground(
+            this, NotificationId.DOWNLOAD_IN_PROGRESS.id, summaryNotificationBuilder.build(),
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.UPSIDE_DOWN_CAKE) {
+                ServiceInfo.FOREGROUND_SERVICE_TYPE_DATA_SYNC
+            } else {
+                0
+            }
```
Bnyro, line 530: `FOREGROUND_SERVICE_TYPE_DATA_SYNC` was added with Android 29, so we should be fine by always using that one.

Also, why does the condition use `UPSIDE_DOWN_CAKE` (SDK 34) instead of SDK 29?

Issue #7740, opened 2025-10-09
Clear video queue confirmation


NannyBo:
### Describe your suggested feature

When a video queue exists, attempting to play a video outside of the queue should show a confirmation dialog such as, "Playing this video will clear the queue. Are you sure you want to continue?"

### Other details

I can't count the amount of times I've had a large queue built up and accidentally clicked on another video while scrolling, completely wiping out my queue without confirmation. This issue isn't related to the clear queue button, but rather a request for a dialog box to prevent accidentally wiping out the queue when scrolling through videos (which happens often and is easy to do).

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [x] I have written a short but informative title.
- [x] I will fill out all of the requested information in this form.

Bnyro:
Since https://github.com/libre-tube/LibreTube/pull/7676, the queue is kept anyways, so no changes needed here.

I've added a confirmation dialog to the playing queue sheet (although not requested here) because I think that the button there is easy to hit by accident

Issue #7659, opened 2025-09-06
Remove "Confirm unsubscribing" option


Bnyro:
### Describe your suggested feature

- Destructive actions like unsubscribing should always require a confirmation.
- to make LibreTube easier to use, we should remove these rather unimportant settings where possible

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [x] I have written a short but informative title.
- [x] I will fill out all of the requested information in this form.

FineFindus:
I totally agree that we should remove this preference, however, I'm not sure if always displaying is a good idea. In general, it's better to give the user an option to undo their choice, rather than first asking for confirmation ([1](https://developer.gnome.org/hig/principles.html#:~:text=Allow%20destructive%20actions%20to%20be%20undone%20instead%20of%20asking%20for%20confirmation), [2](https://design.gitlab.com/patterns/destructive-actions/#low-severity), [3](https://goodui.org/#8)). Considering that the easiest way to undo is just to re-subscribe, I don't think a confirmation dialog is required.

dhanuarf:
> I totally agree that we should remove this preference, however, I'm not sure if always displaying is a good idea. In general, it's better to give the user an option to undo their choice, rather than first asking for confirmation ([1](https://developer.gnome.org/hig/principles.html#:~:text=Allow%20destructive%20actions%20to%20be%20undone%20instead%20of%20asking%20for%20confirmation), [2](https://design.gitlab.com/patterns/destructive-actions/#low-severity), [3](https://goodui.org/#8)). Considering that the easiest way to undo is just to re-subscribe, I don't think a confirmation dialog is required.

To be honest, the reason I have this option always checked  is just to notify myself if I accidentally pressed "unsubscribe". 

Because currently the way this app handle subscribe/unsubsricbe is just by toggling button without any notification (or toast). So if I accidentally touch that button I wouldnt have known, and then missed all of the new videos from that channel.

Having confirmation dialog serves as a good indicator if intentionally clicked the button.

FineFindus:
I think that can be avoided by making the difference between the states (i.e. subscribed and not subscribed) more noticeable, e.g. by adding a toast or using different colors/shapes (as suggested in the M3E guidelines) for the button.

Bnyro:
In my opinion we should for now add a confirmation dialog to the channel fragment and the player fragment because there's always only one channel and it could happen that you accidentially unsubscribe.

[...]

PR #7752, opened 2025-10-10
feat: request user confirmation before clearing playback queue

Bnyro:


FineFindus:
I feel like this might be another case of [adding a confirmation dialog where an undo would be better](https://github.com/libre-tube/LibreTube/issues/7659). Or maybe we could just add the video to the current queue, if it's not already?

Bnyro:
I think it makes sense to add a dialog here because it's easy to accidentially hit the "clear" button as it's in a row with many other similar looking buttons.

> Or maybe we could just add the video to the current queue, if it's not already?
>
What do you mean by that? Currently we already keep the current video when clearing the queue.

Bnyro:
The changes here only affect the playing queue sheet, probably should have mentioned that in the pr title

Bnyro:
So https://github.com/libre-tube/LibreTube/issues/7740 is not really related, yeah :)

FineFindus:
> I think it makes sense to add a dialog here because it's easy to accidentally hit the "clear" button as it's in a row with many other similar looking buttons.

I don't disagree that it makes sense to prevent accidental taps, however having to confirm that users actually want to clear the queue every time is a bit overkill, don't you think? I feel like an undo options is much less intrusive whilst still fulfilling the same goal :)

> What do you mean by that? Currently we already keep the current video when clearing the queue.

Sorry, got it mixed up.

> So https://github.com/libre-tube/LibreTube/issues/7740 is not really related, yeah :)

Related in the sense that we discussed avoiding confirmation dialogs there.

PR #7743, opened 2025-10-09
feat: improve fullscreen button placement and bottom bar UI

101paul:
### Summary
This PR improves the user experience of the video player by adjusting the placement of the fullscreen button and refining the bottom control bar layout.

### Changes
- Moved the fullscreen button slightly away from the progress bar to prevent accidental taps that skip the video.
- Adjusted spacing and alignment of the bottom player controls for a cleaner, more accessible layout.
- Minor UI improvements for better usability in both landscape and portrait modes.

### Motivation
Currently, the fullscreen button is positioned too close to the progress bar, which often causes users to accidentally jump to the end of the video. These changes aim to make video playback more intuitive and reduce accidental interactions.

This is a **small enhancement** but it significantly improves user interaction with the video player.

PR #7753, opened 2025-10-10
Fix initial video player margin issue in PlayerFragment 

101paul:
Hey there ! Thanks for the feedback

This is a follow-up to my previous PR #7743 . 

I’ve updated the implementation so that the margin change for the fullscreen button now occurs only when entering fullscreen mode. In normal mode, the margins remain untouched to keep the layout consistent.
Added a small tooltip for the fullscreen button as a friendly hint.
I’ve also reverted the unintended changes in app/src/main/res/values/languages.xml.
These updates should make the fullscreen layout behave smoothly and keep the UI consistent in both modes.

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
+++ b/app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
@@ -882,7 +882,6 @@ class PlayerFragment : Fragment(R.layout.fragment_player), OnlinePlayerOptions {
         windowInsetsControllerCompat.isAppearanceLightStatusBars = false
 
         commonPlayerViewModel.isFullscreen.value = true
-
         updateFullscreenOrientation()
 
         commonPlayerViewModel.setSheetExpand(null)
--- a/app/src/main/res/values/dimens.xml
+++ b/app/src/main/res/values/dimens.xml
@@ -2,6 +2,9 @@
 <resources>
     <integer name="grid_items">1</integer>
     <integer name="grid_items_landscape">2</integer>
-
+    <dimen name="normal_button_margin_bottom">0dp</dimen>
+    <dimen name="normal_button_margin_end">0dp</dimen>
+    <dimen name="fullscreen_button_margin_bottom">5dp</dimen>
+    <dimen name="fullscreen_button_margin_end">3dp</dimen>
     <item name="landscape_player_constraint_width_percentage" type="dimen" format="float">0.55</item>
 </resources>
--- a/app/src/main/java/com/github/libretube/ui/views/CustomExoPlayerView.kt
+++ b/app/src/main/java/com/github/libretube/ui/views/CustomExoPlayerView.kt
@@ -713,6 +713,18 @@ abstract class CustomExoPlayerView(
                 marginEnd = horizontalMargin
             }
         }
+
+        binding.fullscreen.layoutParams = (binding.fullscreen.layoutParams as MarginLayoutParams).apply {
+            if (isFullscreen()) {
+                // Add extra bottom margin in fullscreen mode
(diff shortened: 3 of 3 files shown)


osamasayed585:
I think if we’re going to change this part, we should apply the adjustment suggested by [krwow](https://github.com/libre-tube/LibreTube/issues/7756)

101paul:
> I think if we’re going to change this part, we should apply the adjustment suggested by [krwow](https://github.com/libre-tube/LibreTube/issues/7756)

I think the existing design works well overall. The only concern is that the icon appears a bit too close to the progress bar — adjusting that spacing should address the issue for now.

app/src/main/res/values/dimens.xml
```diff
@@ -2,6 +2,9 @@
 <resources>
     <integer name="grid_items">1</integer>
     <integer name="grid_items_landscape">2</integer>
-
+    <dimen name="normal_button_margin_bottom">0dp</dimen>
+    <dimen name="normal_button_margin_end">0dp</dimen>
+    <dimen name="fullscreen_button_margin_bottom">11dp</dimen>
```
Bnyro: 11dp is a bit too much imo because that doesn't align anymore with the video duration and progress texts in the left.

Everything up to `5dp` looks still okay, but it shouldn't be much more in my opinion.

app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
```diff
@@ -911,6 +911,24 @@ class PlayerFragment : Fragment(R.layout.fragment_player), OnlinePlayerOptions {
...
+            // Add extra bottom margin in fullscreen
+            layoutParams.bottomMargin = resources.getDimensionPixelSize(R.dimen.fullscreen_button_margin_bottom)
+            layoutParams.marginEnd = resources.getDimensionPixelSize(R.dimen.fullscreen_button_margin_end)
+        } else {
+            // Reset to default margin
+            layoutParams.bottomMargin = resources.getDimensionPixelSize(R.dimen.normal_button_margin_bottom)
+            layoutParams.marginEnd = resources.getDimensionPixelSize(R.dimen.normal_button_margin_end)
+        }
+
+        playerControlsBinding.fullscreen.layoutParams = layoutParams
```
Bnyro: Please put this logic into `CustomExoPlayerView.kt` and call it from there from within `onFullscreenChange`, see https://github.com/libre-tube/LibreTube/blob/master/app/src/main/java/com/github/libretube/ui/views/CustomExoPlayerView.kt#L824.

That way it also works if playing downloaded videos offline :)

Bnyro:
> I think if we’re going to change this part, we should apply the adjustment suggested by [krwow](https://github.com/libre-tube/LibreTube/issues/7756)
>
I think it's fine to merge both, but perhaps it makes sense if you could test https://github.com/libre-tube/LibreTube/pull/7757 @101paul to see if it makes sense to apply these changes here as well after #7757 in your opinion.

app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
101paul: Hii @Bnyro 
I’ve moved the fullscreen button margin logic into CustomExoPlayerView.kt and it’s now called inside onFullscreenChange() like you suggested. It should now work properly even for downloaded videos offline :)

All changes are pushed to the fix-video-margins branch. Excited to see it merged 

Thanks for the guidance !

app/src/main/res/values/dimens.xml
101paul: Got it done 🎉

101paul:
I just went ahead and tested #7757 together with my PR by merging them and everything is working flawlessly :) 
It looks great and I think applying these changes after #7757 will work really smoothly!

Issue #7742, opened 2025-10-09
Fullscreen icon too close to video progress bar


krwow:
### Describe your suggested feature

Hello,

I think that placement of fullscreen icon during video playback is not optimal and it's too close to video progress bar. It often makes me accidentally go to the end of video instead of making player full screen. Because of that I often have to scroll back and try to locate the last seen part of video. It is most annoying in a landscape view.

I believe that placing this icon next to the progress bar, in the right bottom corner (with some empty space between icon and bar) would make it harder to touch progress bar instead of icon by accident.

<img width="454" height="204" alt="Image" src="https://github.com/user-attachments/assets/7c29a989-b6db-4ab0-8389-ad11bb820c10" />

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [x] I have written a short but informative title.
- [x] I will fill out all of the requested information in this form.

osamasayed585:
This is the same behavior on YouTube

krwow:
You are right, so maybe the issue is behavior of video progress bar, not the icon placement.

osamasayed585:
Good point 👍 
I totally agree and support this idea. The progress bar should work like YouTube’s by dragging the dot, not by tapping

Bnyro:
https://github.com/libre-tube/LibreTube/pull/7753

PR #7757, opened 2025-10-12
feat: allow drag-to-seek like YouTube

### LINKED ISSUE #7756: Changing behaviour of video progress bar to match the other apps

In a YT app or website you have to drag the dot located on progress bar to change currently watching part of a video, while in LibreTube it is just a matter of touching another part of progress bar and that's troublesome for me.

In my case I often touch the progress bar instead of full screen icon in the bottom right corner, so I have to try to recall and go back to the last seen part of video. As I have already mentioned, it is usually the way I suggested in the other apps.

### Other details

_No response_

### Acknowledgements

osamasayed585:
Summary
-------
This PR changes the seekbar behavior so that users must drag the scrubber (thumb) to change the video position, instead of jumping by tapping anywhere on the bar. This reduces accidental taps when trying to tap the full-screen button.

Changes
-------
- Updated DismissableTimeBar.onTouchEvent to ignore taps distant from the current scrubber position.
- Kept existing preview behavior (SeekbarPreviewListener) unchanged.

Related issues
--------------
- closes #7756
- Related to #7742

Notes
-----
- This PR only changes how the seekbar works (drag-to-seek).
- The full-screen icon margins stay the same to keep the current layout.
- The main issue was caused by the tap-to-seek behavior, which this fix solves.

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/views/DismissableTimeBar.kt
+++ b/app/src/main/java/com/github/libretube/ui/views/DismissableTimeBar.kt
@@ -22,6 +22,8 @@ open class DismissableTimeBar(
     private var shouldAddListener = false
     var exoPlayer: Player? = null
 
+    private val listeners = mutableListOf<OnScrubListener>()
+
     // Drag-only seeking state
     private var initialX: Float = 0f
     private var initialY: Float = 0f
@@ -30,15 +32,22 @@ open class DismissableTimeBar(
     private val touchSlopPx: Int = ViewConfiguration.get(context).scaledTouchSlop
 
     init {
-        addSeekBarListener(object : OnScrubListener {
-            override fun onScrubStart(timeBar: TimeBar, position: Long) = Unit
+        super.addListener(object : OnScrubListener {
+            override fun onScrubStart(timeBar: TimeBar, position: Long) {
+                listeners.forEach { it.onScrubStart(timeBar, position) }
+            }
 
-            override fun onScrubMove(timeBar: TimeBar, position: Long) = Unit
+            override fun onScrubMove(timeBar: TimeBar, position: Long) {
+                listeners.forEach { it.onScrubMove(timeBar, position) }
+            }
 
             override fun onScrubStop(timeBar: TimeBar, position: Long, canceled: Boolean) {
+                listeners.forEach { it.onScrubStop(timeBar, position, canceled) }
+
                 if (canceled) return
                 // Ignore if gesture started too far above the bar (keep original behavior)
                 if (initialY <= MINIMUM_ACCEPTED_HEIGHT.dpToPx()) return
+
                 exoPlayer?.seekTo(position)
             }
         })
@@ -109,17 +118,16 @@ open class DismissableTimeBar(
     /**
      * DO NOT CALL THIS METHOD DIRECTLY. Use [addSeekBarListener] instead!
      */
+    @Deprecated("Use addSeekBarListener instead")
     override fun addListener(listener: OnScrubListener) {
-        if (shouldAddListener) super.addListener(listener)
+        // do nothing, see below on how listeners should be set
     }
 
     /**
      * Wrapper to circumvent adding the listener created by [PlayerControlView]
      */
     fun addSeekBarListener(listener: OnScrubListener) {
-        shouldAddListener = true
-        addListener(listener)
-        shouldAddListener = false
+        listeners.add(listener)
     }
 
     fun setPlayer(player: Player) {


app/src/main/java/com/github/libretube/ui/views/DismissableTimeBar.kt
```diff
@@ -37,6 +38,23 @@ open class DismissableTimeBar(
     override fun onTouchEvent(event: MotionEvent): Boolean {
         lastYPosition = event.y
 
+        if (exoPlayer == null) return super.onTouchEvent(event)
```
Bnyro:
```suggestion
        val exoPlayer = exoPlayer ?: return super.onTouchEvent(event)
```
This prevents the null-assertions later on

app/src/main/java/com/github/libretube/ui/views/DismissableTimeBar.kt
```diff
@@ -37,6 +38,23 @@ open class DismissableTimeBar(
...
+                val duration = exoPlayer!!.duration
+                val currentPosition = exoPlayer!!.currentPosition
+
+                // We calculate the current point position relative to the length of the bar
+                val thumbX = (currentPosition.toFloat() / duration) * width
+                val touchX = event.x
+                val distance = abs(thumbX - touchX)
+
+                // Ignore touches that are far from the point
+                if (distance > 40) return false
```
Bnyro: Please move `40` to a constant in the companion object (line 82)

Bnyro:
Thanks for the PR @osamasayed585 

I personally find it a bit inconvenient to not be able to seek by clicking directly on the timebar, but I guess I'll get used to it.

Let's however keep this open for discussion for some days.

app/src/main/java/com/github/libretube/ui/views/DismissableTimeBar.kt
osamasayed585: Good point, I’ll update it

osamasayed585:
Thanks for your feedback @Bnyro 

This behavior is similar to YouTube. Users are usually used to using the handle (thumb) rather than by random taps.

dhanuarf:
If you have system navigation gesture enabled, and you're trying to drag from the side, it can conflicts with the system gesture. So I think if we really want to implement this we should either: 1. increase horizontal margin, or 2. add [system gesture exclusion](https://developer.android.com/develop/ui/views/touch-and-input/gestures/gesturenav) around the sides.

I think we could also increase the ~~thumb size~~ drag treshold for better accesibility

**Edit**:
 I just tried youtube official app and the dragging behavior is different than this implementation.
In the official app, you can drag anywhere on the timebar, not just from the thumb area.

Bnyro:
Perhaps we should not seek the player when a user does a single tap (possibly a misclick), but still allow starting to drag anywhere on the seekbar if they do more than just a single tap?

Bnyro:
I've added a commit on top to implement my suggestion https://github.com/libre-tube/LibreTube/pull/7757/commits/b4c364ea9bb2fe3c89183130e454274e83a756e5.

I tried to implement this in `onTouchEvent`, but it seemed to be far too complicated because we would basically rewrite all logic in `DefaultTimeBar` for triggering listeners.

The side effect of my implementation that tapping the seekbar once still moves the thumb there for a short time until you release the tap, so it's a small visual glitch.

Please let me know if you find a better way to handle that!

osamasayed585:
I tried many ways, but it’s not working as we expect.

I’ll keep trying until we get the result we want.

osamasayed585:
Feature Complete 🎉
The drag-only seek bar with YouTube-style behavior is now fully implemented and ready. Users can now drag to seek from the current position

Bnyro:
> Feature Complete 🎉 The drag-only seek bar with YouTube-style behavior is now fully implemented and ready. Users can now drag to seek from the current position
>
Great job, I can confirm it works as we expect it to now, definitely in a merge-able state now! Thanks for working on that, looks much better than my initial implementation :)

What I'm still unsure about is whether it's better to start moving the thumb at the current video progress or at the actual position where the touch event is detected. In my opinion it's more intuitive to move the progress to the actual position where the user placed their finger instead of changing the position relatively to the current video progress.

That's very subjective though, modifying that behavior should be fairly simple anyways depending on what we decide to in the end.

@FineFindus , @dhanuarf Could you please test these changes as well to tell which option you prefer?

dhanuarf:
> What I'm still unsure about is whether it's better to start moving the thumb at the current video progress or at the actual position where the touch event is detected. In my opinion it's more intuitive to move the progress to the actual position where the user placed their finger instead of changing the position relatively to the current video progress.
> 
> @FineFindus , @dhanuarf Could you please test these changes as well to tell which option you prefer?

To be honest, I personally still prefer the default seekbar behavior. I think the only purpose for this behavior is that it prevents accidental tap. But I think the probability of that is fairly low because you have to tap the screen before you can see the seekbar. 

I would prefer to drag to seek anywhere on the screen (instead of just on the seekbar) like most modern mobile video player rather than this though.

Or maybe we can create a fake view at the bottom video frame to trigger this drag to seek when the controls are hidden, but still keep the default seekbar behavior when the controls are visible. So basically instead of 'drag anywhere', we implement 'drag at the bottom'

osamasayed585:
> > What I'm still unsure about is whether it's better to start moving the thumb at the current video progress or at the actual position where the touch event is detected. In my opinion it's more intuitive to move the progress to the actual position where the user placed their finger instead of changing the position relatively to the current video progress.
> > @FineFindus , @dhanuarf Could you please test these changes as well to tell which option you prefer?
> 
> To be honest, I personally still prefer the default seekbar behavior. I think the only purpose for this behavior is that it prevents accidental tap. But I think the probability of that is fairly low because you have to tap the screen before you can see the seekbar.
> 
> I would prefer to drag to seek anywhere on the screen (instead of just on the seekbar) like most modern mobile video player rather than this though.
> 
> Or maybe we can create a fake view at the bottom video frame to trigger this drag to seek when the controls are hidden, but still keep the default seekbar behavior when the controls are visible. So basically instead of 'drag anywhere', we implement 'drag at the bottom'

@dhanuarf Actually, we have three possible behaviors for the seek bar -> 

1. **YouTube-style** (current implementation): 
    - Drag starts from current thumb position
   - Thumb moves relative to finger movement

2. **TikTok-style** ( we can do this):
   - On drag start: thumb instantly jumps to finger position  
   - Then continues dragging from there

3. **Facebook-style** (old behavior of this PR):
   - On drag start: thumb stays at current position
   - But dragging moves thumb directly to finger position (not relative)

So all require dragging, but with different starting behaviors
It's not a new concept - it's what billions of users are familiar with and already expect from modern video apps

Bnyro:
As I stated before (and if I understand your list correctly), I'd vote for the Tiktok-Style because it feels unnatural that dragging starts somewhere else than at the position of your own finger.

dhanuarf:
Well the 'tiktok-style' is literally the default behavior is it not?  I dont use tiktok but judging from the description mentioned above it's just how regular seekbar works.

Bnyro:
> Well the 'tiktok-style' is literally the default behavior is it not? I dont use tiktok but judging from the description mentioned above it's just how regular seekbar works.
>
I think the difference is that doing a single tap doesn't trigger seeking to that position, only when dragging. If not, I probably misunderstood the TikTok style :shrug:

osamasayed585:
You're both right! Let me clarify:

**TikTok-style vs Default behavior:**
- **Old default**: Tap anywhere → immediate seek (no dragging needed)
- **TikTok-style**: Drag anywhere → thumb jumps to finger position, then continues dragging
- **Current (YouTube-style)**: Drag anywhere → thumb moves relative from current position

The key difference: TikTok makes the thumb "jump" to your finger on drag start, while YouTube keeps it at current position and moves relatively.

All three require dragging except the old default which allowed tap-to-seek.

dhanuarf:
Ohh so the difference is that in that style it only starts seeking when the finger moves. Well, it's a slight difference but I think it's going to take a while for me to adapt to that :)

osamasayed585:
Exactly, most users adapt within a few uses since it follows YouTube's pattern they already know. thanks for giving it a try 😊

Bnyro:
Thank you @osamasayed585! I've refactored the code to the TikTok-behavior, but I'm leaving your whole code for the YouTube-style here if we decide to ever change the behavior in the future (and because it's clean code):

```kotlin
package com.github.libretube.ui.views

import android.annotation.SuppressLint
import android.content.Context
import android.util.AttributeSet
import android.view.MotionEvent
import android.view.ViewConfiguration
import androidx.media3.common.Player
import androidx.media3.common.util.UnstableApi
import androidx.media3.ui.DefaultTimeBar
import androidx.media3.ui.PlayerControlView
import androidx.media3.ui.TimeBar
import androidx.media3.ui.TimeBar.OnScrubListener
import com.github.libretube.extensions.dpToPx
import kotlin.math.abs

@UnstableApi
open class DismissableTimeBar(
    context: Context,
    attributeSet: AttributeSet? = null
): DefaultTimeBar(context, attributeSet) {
    private var shouldAddListener = false
    var exoPlayer: Player? = null

    // Drag-only seeking state
    private var initialX: Float = 0f
    private var initialY: Float = 0f
    private var initialDownTime: Long = 0L
    private var waitingForDrag: Boolean = false
    private var dragStarted: Boolean = false
    private val touchSlopPx: Int = ViewConfiguration.get(context).scaledTouchSlop
    private var startXAtCurrentProgress: Float = 0f

    init {
        addSeekBarListener(object : OnScrubListener {
            override fun onScrubStart(timeBar: TimeBar, position: Long) = Unit

            override fun onScrubMove(timeBar: TimeBar, position: Long) = Unit

            override fun onScrubStop(timeBar: TimeBar, position: Long, canceled: Boolean) {
                if (canceled) return
                // Ignore if gesture started too far above the bar (keep original behavior)
                if (initialY <= MINIMUM_ACCEPTED_HEIGHT.dpToPx()) return
                exoPlayer?.seekTo(position)
            }
        })
    }

    @SuppressLint("ClickableViewAccessibility")
    override fun onTouchEvent(event: MotionEvent): Boolean {
        when (event.actionMasked) {
            MotionEvent.ACTION_DOWN -> {
                initialX = event.x
                initialY = event.y
                initialDownTime = event.downTime
                waitingForDrag = true
                dragStarted = false
                // Consume without forwarding to prevent tap-to-seek or thumb jump
                return true
            }

            MotionEvent.ACTION_MOVE -> {
                if (waitingForDrag) {
                    val dx = abs(event.x - initialX)
                    val dy = abs(event.y - initialY)
                    if (dx > touchSlopPx || dy > touchSlopPx) {
                        // Compute starting X based on current player progress so the thumb
                        // starts from the current playback position (YouTube-like behavior)
                        startXAtCurrentProgress = computeXForCurrentProgress()

                        // Begin scrubbing now by synthesizing a DOWN at the current progress X
                        val fakeDown = MotionEvent.obtain(event)
                        fakeDown.action = MotionEvent.ACTION_DOWN
                        fakeDown.setLocation(startXAtCurrentProgress, event.y)
                        super.onTouchEvent(fakeDown)
                        fakeDown.recycle()

                        waitingForDrag = false
                        dragStarted = true
                    } else {
                        // Still not considered a drag; consume
                        return true
                    }
                }

                // Forward MOVE only after we've started dragging, with relative X
                if (dragStarted) {
                    val forwardX = clampToTrack(startXAtCurrentProgress + (event.x - initialX))
                    val fakeMove = MotionEvent.obtain(event)
                    fakeMove.setLocation(forwardX, event.y)
                    val handled = super.onTouchEvent(fakeMove)
                    fakeMove.recycle()
                    return handled
                }
                return true
            }

            MotionEvent.ACTION_UP, MotionEvent.ACTION_CANCEL -> {
                // If we started a drag, forward the terminal event to finish scrubbing
                if (dragStarted) {
                    val forwardX = clampToTrack(startXAtCurrentProgress + (event.x - initialX))
                    val fakeUp = MotionEvent.obtain(event)
                    fakeUp.setLocation(forwardX, event.y)
                    val handled = super.onTouchEvent(fakeUp)
                    fakeUp.recycle()
                    waitingForDrag = false
                    dragStarted = false
                    return handled
                }

                // It's a tap without drag: do nothing (prevent seek and thumb jump)
                waitingForDrag = false
                dragStarted = false
                // Optionally report click for accessibility without side effects
                performClick()
                return true
            }
        }

        return super.onTouchEvent(event)
    }

    private fun computeXForCurrentProgress(): Float {
        val durationMs = exoPlayer?.duration?.takeIf { it > 0 } ?: 0L
        val positionMs = exoPlayer?.currentPosition ?: 0L
        val fraction = if (durationMs > 0) positionMs.toFloat() / durationMs.toFloat() else 0f
        val left = paddingLeft.toFloat()
        val right = (width - paddingRight).toFloat()
        val barWidth = (right - left).coerceAtLeast(1f)
        val unclampedX = left + fraction * barWidth
        return clampToTrack(unclampedX)
    }

    private fun clampToTrack(x: Float): Float {
        val left = paddingLeft.toFloat()
        val right = (width - paddingRight).toFloat()
        return x.coerceIn(left, right)
    }

    /**
     * DO NOT CALL THIS METHOD DIRECTLY. Use [addSeekBarListener] instead!
     */
    override fun addListener(listener: OnScrubListener) {
        if (shouldAddListener) super.addListener(listener)
    }

    /**
     * Wrapper to circumvent adding the listener created by [PlayerControlView]
     */
    fun addSeekBarListener(listener: OnScrubListener) {
        shouldAddListener = true
        addListener(listener)
        shouldAddListener = false
    }

    fun setPlayer(player: Player) {
        this.exoPlayer = player
    }

    companion object {
        private const val MINIMUM_ACCEPTED_HEIGHT = -70f
    }
}
```

PR #7758, opened 2025-10-12
fix(trending): 'Trailers' and 'Podcasts' content got swapped

dhanuarf:
~~I understand we can just swap the enum, but~~ this way we dont need to match the exact order of the enum.

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/fragments/HomeFragment.kt
+++ b/app/src/main/java/com/github/libretube/ui/fragments/HomeFragment.kt
@@ -143,7 +143,7 @@ class HomeFragment : Fragment(R.layout.fragment_home) {
             ).let { categoryName -> trendingCategories.first { it.name == categoryName } }
 
             val categories = trendingCategories.map { category ->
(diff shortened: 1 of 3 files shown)


dhanuarf:
~~Added new commit [bde1b7e](https://github.com/libre-tube/LibreTube/pull/7758/commits/bde1b7e71dfb0270e76d49fa508684b7c875ed8a)~~

app/src/main/java/com/github/libretube/ui/fragments/HomeFragment.kt
```diff
@@ -140,20 +140,20 @@ class HomeFragment : Fragment(R.layout.fragment_home) {
             val currentTrendingCategoryPref = PreferenceHelper.getString(
                 PreferenceKeys.TRENDING_CATEGORY,
                 TrendingCategory.LIVE.name
-            ).let { TrendingCategory.valueOf(it) }
+            ).let { categoryName -> trendingCategories.first {it.name == categoryName } }
```
Bnyro:
```suggestion
            ).let { categoryName -> trendingCategories.first { it.name == categoryName } }
```
nit: formatting

PR #7762, opened 2025-10-14
refactor: use enum for top-level destinations

osamasayed585:
I replaced the AppShortcut data class with a TopLevelDestination enum.
This enum defines the main destinations (Home, Trends, Subscriptions, Library) with their routes, icons, and labels in one place.

This makes the code cleaner and easier to manage when creating shortcuts or handling navigation from intents.

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/activities/MainActivity.kt
+++ b/app/src/main/java/com/github/libretube/ui/activities/MainActivity.kt
@@ -480,10 +480,18 @@ class MainActivity : BaseActivity() {
         // navigate to (temporary) playlist or channel if available
         if (navigateToMediaByIntent(intent)) return
 
+        // Get saved search query if available
         intent?.getStringExtra(IntentData.query)?.let {
             savedSearchQuery = it
         }
 
+        // Open the Downloads screen if requested
+        if (intent?.getBooleanExtra(IntentData.OPEN_DOWNLOADS, false) == true) {
+            navController.navigate(R.id.downloadsFragment)
+            return
+        }
+
+        // Handle navigation from app shortcuts (Home, Trends, etc.)
         intent?.getStringExtra(IntentData.fragmentToOpen)?.let {
             ShortcutManagerCompat.reportShortcutUsed(this, it)
             when (it) {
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/github/libretube/ui/activities/MainActivity.kt
```diff
@@ -483,17 +484,13 @@ class MainActivity : BaseActivity() {
...
-            }
-
+        intent?.getStringExtra(IntentData.fragmentToOpen)?.let {
+            ShortcutManagerCompat.reportShortcutUsed(this, it)
             when (it) {
-                "home" -> navController.navigate(R.id.homeFragment)
-                "trends" -> navController.navigate(R.id.trendsFragment)
-                "subscriptions" -> navController.navigate(R.id.subscriptionsFragment)
-                "library" -> navController.navigate(R.id.libraryFragment)
-                "downloads" -> navController.navigate(R.id.downloadsFragment)
```
Bnyro, line 496: We still need to keep the `downloads` case here, the `downloads` action is triggered when users press the notification while downloading a video.

osamasayed585: I’ll move the downloads logic to use a separate intent instead of mixing it with the top-level destinations.

#### For example

**In `DownloadService`:**
```kotlin
private fun getNotificationBuilder(item: DownloadItem): Builder {
    val intent = Intent(this, MainActivity::class.java)
        .putExtra("openDownloads", true) // this key will be in the `IntentData` class with the same name 
    val activityIntent = PendingIntentCompat.getActivity(
        this, 0, intent, FLAG_CANCEL_CURRENT, false
    )

    return Builder(this, DOWNLOAD_CHANNEL_NAME)
        .setContentTitle("[${item.type}] ${item.fileName}")
        .setProgress(0, 0, true)
        .setOngoing(true)
        .setContentIntent(activityIntent)
}
```
**And in `MainActivity`:** 

```kotlin 
if (intent?.getBooleanExtra("openDownloads", false) == true) {
    navController.navigate(R.id.downloadsFragment)
    return
}
```

This keeps the downloads logic separate from top-level destinations and makes the intent handling clearer.

Bnyro: I agree, that's a good idea. Perhaps we should even move `"openDownloads"` to a constant variable so we can't misspell it :)

PR #7766, opened 2025-10-15
feat: playlist layout

Elolcol:


app/src/main/res/layout/fragment_playlist.xml
```diff
@@ -99,13 +99,13 @@
                         android:layout_weight="1"
                         android:layout_height="wrap_content"
                         android:orientation="vertical"
-                        android:layout_marginHorizontal="10dp">
+                        android:layout_marginStart="12dp"
+                        android:layout_marginEnd="8dp">
 
-                        <com.github.libretube.ui.views.ExpandableTextView
+                        <TextView
```
Bnyro: Why did you remove the expandable text view here? If the description is very long, that allows us to only display the first 2 lines and allow the user to expand the description by clicking it.

Perhaps we should redesign the ExpandableTextView instead of removing it here.

Elolcol: The description or the title?.

Bnyro: Sorry - I meant the title, not the description.

Mixed it up with the channel page.

Some playlists have very long titles, especially since users can set the name of their own private playlists themselves.

PR #7767, opened 2025-10-15
refactor: improve sizes in home layout

Elolcol:


app/src/main/res/layout/fragment_home.xml
```diff
@@ -130,20 +131,19 @@
                         android:id="@+id/trendingRV"
                         android:layout_width="match_parent"
                         android:layout_height="wrap_content"
-                        android:layout_marginHorizontal="10dp"
                         android:nestedScrollingEnabled="false"
                         android:visibility="gone"
-                        app:layoutManager="androidx.recyclerview.widget.GridLayoutManager"
-                        app:spanCount="2" />
```
Bnyro, line 137: I personally think that 2 columns looked better here.

Elolcol: I think it looks better with a single column; it's easier to read and more similar to the subscription page.

Bnyro (approved):

PR #7768, opened 2025-10-15
feat: download playlist name

Elolcol:


app/src/main/java/com/github/libretube/ui/dialogs/DownloadPlaylistDialog.kt
```diff
@@ -44,6 +44,7 @@ class DownloadPlaylistDialog : DialogFragment() {
         val possibleAudioQualities = resources.getStringArray(R.array.audioQualityBitrates)
             .toList()
 
+        binding.playlistTitle.text = playlistName
```
Bnyro, line 47: Maybe we should put the playlist name inside `"`
```suggestion
        binding.playlistTitle.text = "\"$playlistName\""
```
or do something else to visually highlight it, because it looks very similar to the TextView below at the moment, so it's hard to spot the playlist name at first glance

Elolcol: I forgot to add weight to the title so that it would look good.

PR #7774, opened 2025-10-16
feat: comments row, replies and sheet styles and layout

Elolcol:


Bnyro:
I think it'd be better to keep the button background transparent and not make them filled, somehow this doesn't look really good in my opinion.

What's your opinion @FineFindus ?

Apart from that I very much like the other changes, thanks for the PR!

FineFindus (commented):
I agree, I think in general color should be used to nudge the user towards important actions available to them (e.g. the play button for a playlist) and neither the back button nor the close button is that important for the color.

Elolcol:
So, how about we make it paler?

`android:backgroundTint="?colorSurface"`

image
image

PR #7776, opened 2025-10-16
fix: home screen error not shown if homepage fails to load

Bnyro:
That's a revert of changes you made @FineFindus.

On the current master branch, I can't seem to trigger the "Change instance" page when it fails to load.

If you have a better idea, we'll take that :)


FineFindus:
I don't fully remember why I changed that, but I think it fixed the issue where the screen would always be in a loading state when no section has any items (i.e. fresh install and trending not responding). Considering that we've removed anyway, maybe that's not an issue anymore?

Bnyro:
> Considering that we've removed anyway, maybe that's not an issue anymore?
>
Yeah, perhaps.

PR #7782, opened 2025-10-17
feat: improve description style and layout

Elolcol:


Bnyro (commented):
I'm personally not sure about changing the font weight for the like and dislike count because the view count and date in the same row are not bold either, so in my opinion we should keep the font weight normal there.

Apart from that lgtm.

Bnyro (approved):
Thanks!

PR #7795, opened 2025-10-18
feat: fetch dearrow via background thread in VideoCardsAdapter

dhanuarf:
Correcting a  typo.


FineFindus:
Thanks!

(As a nitpick, this is rather a `fix` than a `feat`, since it fixes a bug)

PR #7816, opened 2025-10-26
feat: add 'change  region' button in TrendsFragment

dhanuarf:
For simpler and easier access.

Screenshot_20251026-133423_LibreTube Debug
<br>
<br>
---------
<br>
Also, loading indicators are moved to the top most of the view. And the content's alpha is reduced when it's loading just to create some sort of 'focus' to the loading indicator.

### DIFF
--- a/app/src/main/res/layout/fragment_home.xml
+++ b/app/src/main/res/layout/fragment_home.xml
@@ -102,7 +102,8 @@
                         android:layout_width="wrap_content"
                         android:layout_height="wrap_content"
                         android:layout_gravity="center"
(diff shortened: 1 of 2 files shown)


Cubillas7:
Hello, does it make sense put a tooltip after long press for this new button? I think that there is no now? Thanksss

Bnyro:
> Hello, does it make sense put a tooltip after long press for this new button? I think that there is no now? Thanksss
>
Good catch, added with https://github.com/libre-tube/LibreTube/pull/7816/commits/dd241f7a1f168cda3268760b100f2e516d2ac0f6

Issue #7828, opened 2025-10-27
share no longer works with 0.29.0


ghost:
### Steps to reproduce

1. Open a video
2. Choose "share" button
3. Select an appropriate app (Fennec, Brave, Seal...)

### Expected behavior

The link is opened in the selected app

### Actual behavior

The link is not opened, and nothing appears in the app. Eg: Fennec shows an empty page

### LibreTube version

0.29.0

### Android version

14

### Other details

It does not matter if you share a YouTube or piped links.
Copying the link still works

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [x] I have written a short but informative title.
- [x] I will fill out all of the requested information in this form.

Figim:
Can't really launch downloader apps with the LibreTube share button. The main issue

wuffi1977:
hello where is the download folder of libre tube i can t find my downloads pls help

PR #7830, opened 2025-10-27
fix: shared url not being retrieved properly

dhanuarf:
In my case it fixes the issue described here #7828, but I'm still not sure why it's fixed. Cant find any documentation or any forum regarding this though.

Before the change, opening the shared url in Cromite crash that app, the log doesnt seem much of a help either. And some other apps just doesnt retrieve the url at all.

The documentation on [Intent.EXTRA_TEXT](https://developer.android.com/reference/android/content/Intent#EXTRA_TEXT) says that if you send `CharSequence`, the recipient must use `getCharSequence` to retrieve it, and so I thought "well, maybe that's the reason".
But then I tried to hardcode an url using a char sequence like:
```kotlin
val url: CharSequence = "https://youtu.be/something"
```
The recipient received that text just fine....

Any clue on this?

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/dialogs/ShareDialog.kt
+++ b/app/src/main/java/com/github/libretube/ui/dialogs/ShareDialog.kt
@@ -122,7 +122,7 @@ class ShareDialog : DialogFragment() {
             .setView(binding.root)
             .setPositiveButton(R.string.share) { _, _ ->
                 val intent = Intent(Intent.ACTION_SEND)
-                    .putExtra(Intent.EXTRA_TEXT, binding.linkPreview.text)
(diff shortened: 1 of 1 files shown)


dhanuarf:
Oh I found something, apparently if you make a string using interpolation or concatenation like:
```kotlin
val a = "a"
val b = "b"
val url = "https://youtu.be/$a$b"
```
you need to call `url.toString()` before passing it to intent for some reason.

dhanuarf:
> Thanks for looking into this and finding a fix, do you have a source on it?

No source sorry, just fiddling around myself. At least in my case it behaves like that, not sure :/.

PR #7835, opened 2025-10-29
feat: generate content-bound poToken

FineFindus:
YouTube switched to using content-bound poToken, i.e. the poToken is generated for each video using the video's ID. This fixes issues, where the server would respond with 403, as the previously generated poToken was no longer valid.

Ref: https://github.com/FreeTubeApp/FreeTube/pull/8137

Marked as a draft, as this is being rolled out as a experiment, so it may not be required for every request yet.


Figim:
I  saw a similar issue in Extractor. Can't this be done inside NewPipe extractor? If yt requires it in some videos, use it

FineFindus:
> I saw a similar issue in Extractor. Can't this be done inside NewPipe extractor? If yt requires it in some videos, use it

The issue is misplaced and should be moved to the NewPipe repo. The extractor cannot generate potokens, as they're platform-dependent, whilst the extractor is a platform-independent library.

PR #7841, opened 2025-10-31
fix(TimeBar): listener not being properly removed causing them to accumulate

dhanuarf:
Since we deprecated the use of `addListener` and using our own list of
listeners, we should also deprecate the use of `removeListener` and add
our own method to remove the listeners from our own list.

closes https://github.com/libre-tube/LibreTube/issues/7866

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/activities/OfflinePlayerActivity.kt
+++ b/app/src/main/java/com/github/libretube/ui/activities/OfflinePlayerActivity.kt
@@ -18,6 +18,7 @@ import androidx.media3.common.MediaMetadata
 import androidx.media3.common.Player
 import androidx.media3.session.MediaController
 import androidx.media3.ui.PlayerView
+import androidx.media3.ui.TimeBar
 import com.github.libretube.compat.PictureInPictureCompat
 import com.github.libretube.compat.PictureInPictureParamsCompat
 import com.github.libretube.constants.IntentData
@@ -55,6 +56,7 @@ class OfflinePlayerActivity : BaseActivity() {
     private lateinit var playerController: MediaController
     private lateinit var playerView: PlayerView
     private var timeFrameReceiver: TimeFrameReceiver? = null
+    private var seekBarPreviewListener: TimeBar.OnScrubListener? = null
 
     private lateinit var playerBinding: ExoStyledPlayerControlViewBinding
     private val commonPlayerViewModel: CommonPlayerViewModel by viewModels()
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/github/libretube/ui/activities/OfflinePlayerActivity.kt
```diff
@@ -91,13 +93,11 @@ class OfflinePlayerActivity : BaseActivity() {
...
             if (playbackState == Player.STATE_READY) {
-                binding.player.binding.exoProgress.addSeekBarListener(
-                    SeekbarPreviewListener(
-                        timeFrameReceiver ?: return,
-                        binding.player.binding,
-                        playerController.duration
-                    )
-                )
+                with(binding.player.binding.exoProgress){
+                    seekBarPreviewListener?.also { removeSeekBarListener(it) }
```
Bnyro:
```suggestion
                    seekBarPreviewListener?.let { removeSeekBarListener(it) }
```

app/src/main/java/com/github/libretube/ui/activities/OfflinePlayerActivity.kt
```diff
@@ -91,13 +93,11 @@ class OfflinePlayerActivity : BaseActivity() {
...
+                    seekBarPreviewListener =
+                        createSeekBarPreviewListener()?.also { addSeekBarListener(it) }
```
Bnyro:
```suggestion
           createSeekBarPreviewListener()?.let {
                 addSeekBarListener(it) 
                 seekBarPreviewListener = it
            }
```

app/src/main/java/com/github/libretube/ui/activities/OfflinePlayerActivity.kt
```diff
@@ -202,6 +202,14 @@ class OfflinePlayerActivity : BaseActivity() {
         binding.player.initialize(chaptersViewModel)
     }
 
+    private fun createSeekBarPreviewListener(): TimeBar.OnScrubListener?{
```
Bnyro:
```suggestion
    private fun createSeekBarPreviewListener(): TimeBar.OnScrubListener? {
```
nit: formatting

app/src/main/java/com/github/libretube/ui/activities/OfflinePlayerActivity.kt
```diff
@@ -91,13 +93,11 @@ class OfflinePlayerActivity : BaseActivity() {
...
             // setup seekbar preview
             if (playbackState == Player.STATE_READY) {
-                binding.player.binding.exoProgress.addSeekBarListener(
-                    SeekbarPreviewListener(
-                        timeFrameReceiver ?: return,
-                        binding.player.binding,
-                        playerController.duration
-                    )
-                )
+                with(binding.player.binding.exoProgress){
```
Bnyro: nit: formatting
```suggestion
                with(binding.player.binding.exoProgress) {
```

Bnyro:
Good catch, thanks for the PR!

app/src/main/java/com/github/libretube/ui/activities/OfflinePlayerActivity.kt
Bnyro: This is just for consistency, as we're using `let` everywhere else in the code for doing such operations on nullable operators .

PR #7844, opened 2025-11-01
 feat: drop LBRY HLS support 

FineFindus:
Removes the support for LBRY HLS, as it
- relies on the Piped server injecting an additional stream
- is only available for very few streams
- DASH generally offers superior video streaming over HLS

Additionally, the original company behind LBRY has since closed and there doesn't appear to be any new development in the last few years.

Removing this allows us to remove another quite confusing preference and is generally consistent with the removal of other HLS related preferences.


Bnyro:
That's rather a `chore` than a `feat` I think?

Bnyro (approved):

FineFindus:
I think all directly code-related/ feature changes are a `feat`. Not having to maintain code anymore is also a feature :)

PR #7869, opened 2025-11-03
revert: update media3 to v1.8.0

Bnyro:
Reverts libre-tube/LibreTube#7862

The next good version will be 1.9.0 (which is in alpha), 1.8.0 is broken.


FineFindus:
I love that this is now the second time we've had to revert v1.8.0 :smile:

Bnyro:
> I love that this is now the second time we've had to revert v1.8.0 😄
>
Yeah, I didn't expect `renovate` to open the same update PR a second time - I thought this would finally be a new version containing the fix we discussed...

PR #7900, opened 2025-11-09
feat: remove release action

FineFindus:
Removes the release action, responsible for sending a notification about the release to the Telegram channel. As the action hasn't run successfully in over a year, with no complaints, it is safe to assume, that there is no demand for it.

Saves us a bit of CI time and the environment, since we don't need to needlessly run the action on every release anymore :)

GitHub actions for the Telegram release


Bnyro:
I already forgot it existed :)

Issue #7904, opened 2025-11-10
Some texts are hardcoded and cannot be translated via Weblate


ULTRA-LEGENDS:
### Steps to reproduce

1. Open the LibreTube application.

2. Go to Settings.

3. Navigate to Notifications.

4. Look at the texts inside the “Checking every …” option.

5. Notice that these texts do not appear in Weblate and cannot be translated.

### Expected behavior

All texts in the application should be available in Weblate for translation.

### Actual behavior

Some texts are hardcoded in the source code and cannot be translated.

![Image](https://github.com/user-attachments/assets/f33446ef-1166-4e75-ab10-25442a7b1287)

### LibreTube version

0.29.1

### Android version

Android 15

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [x] I have written a short but informative title.
- [x] I will fill out all of the requested information in this form.

PR #7905, opened 2025-11-11
chore: mark notification frequency values as translatable

FineFindus:
Closes: https://github.com/libre-tube/LibreTube/issues/7904

### DIFF
--- a/app/src/main/res/values/array.xml
+++ b/app/src/main/res/values/array.xml
@@ -145,13 +145,13 @@
     </string-array>
 
     <string-array name="checkingFrequency">
-        <item>15 minutes</item>
-        <item>30 minutes</item>
-        <item>60 minutes</item>
-        <item>2 hours</item>
-        <item>6 hours</item>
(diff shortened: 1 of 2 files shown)


Bnyro:
Wouldn't it be better to let Android translate them? Android definitely also has a way to format time strings inbuilt.

Bnyro:
Ref https://stackoverflow.com/questions/31182821/how-can-i-format-a-time-duration-to-a-string-with-localized-time-units-on-androi

FineFindus:
> Android definitely also has a way to format time strings inbuilt.

Only since API 24 (which is too high for us) with a class that is soft-deprecated. The replacement class that is recommended only works since API 30 :(

Bnyro:
Our minSdk is 26, so this should work? (https://github.com/libre-tube/LibreTube/blob/b75dd6e48239f2e6b85ee0a4471139de15f4bd4f/app/build.gradle.kts#L16)

PR #7908, opened 2025-11-11
fix(LocalFeedRepository): update subscribed channels

FineFindus:
Fixes an issue, where local subscriptions could display outdated information (i.e. having the wrong avatar/name), as the where never updated after being inserted into the database.
To avoid this, we use the channel info, which is fetched when refreshing the local feed to also update the local subscriptions.

### DIFF
--- a/app/src/main/java/com/github/libretube/api/SubscriptionHelper.kt
+++ b/app/src/main/java/com/github/libretube/api/SubscriptionHelper.kt
@@ -1,5 +1,6 @@
 package com.github.libretube.api
 
+import com.github.libretube.api.obj.Subscription
 import com.github.libretube.constants.PreferenceKeys
 import com.github.libretube.db.obj.SubscriptionsFeedItem
 import com.github.libretube.helpers.PreferenceHelper
@@ -57,4 +58,7 @@ object SubscriptionHelper {
 
     suspend fun submitFeedItemChange(feedItem: SubscriptionsFeedItem) =
         feedRepository.submitFeedItemChange(feedItem)
+
+    suspend fun submitSubscriptionChannelInfosChanged(subscriptions: List<Subscription>) =
+        subscriptionsRepository.submitSubscriptionChannelInfosChanged(subscriptions)
 }
--- a/app/src/main/java/com/github/libretube/repo/LocalFeedRepository.kt
+++ b/app/src/main/java/com/github/libretube/repo/LocalFeedRepository.kt
@@ -3,6 +3,7 @@ package com.github.libretube.repo
 import android.util.Log
 import com.github.libretube.api.SubscriptionHelper
 import com.github.libretube.api.obj.StreamItem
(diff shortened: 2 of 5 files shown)


app/src/main/java/com/github/libretube/api/SubscriptionHelper.kt
```diff
@@ -57,4 +58,7 @@ object SubscriptionHelper {
 
     suspend fun submitFeedItemChange(feedItem: SubscriptionsFeedItem) =
         feedRepository.submitFeedItemChange(feedItem)
+
+    suspend fun submitSubscriptionChanges(subscriptions: List<Subscription>) =
```
Bnyro: nit: `submitSubscriptionChanges` sounds a bit like it's triggered when channels are subscribed/unsubscribed, I'd rather call it something like `submitSubscriptionChannelInfosChanged` or similarly

app/src/main/java/com/github/libretube/repo/LocalFeedRepository.kt
```diff
@@ -156,15 +162,14 @@ class LocalFeedRepository : FeedRepository {
                 )
             }
 
-        val channelAvatar = channelInfo.avatars.maxByOrNull { it.height }?.url
-        return related.map { item ->
+        return Pair(subscription, related.map { item ->
```
Bnyro: Please put that `related.map` in a temporary variable to declutter things a bit

app/src/main/java/com/github/libretube/repo/LocalSubscriptionsRepository.kt
```diff
@@ -92,4 +92,15 @@ class LocalSubscriptionsRepository : SubscriptionsRepository {
     override suspend fun getSubscriptionChannelIds(): List<String> {
         return Database.localSubscriptionDao().getAll().map { it.channelId }
     }
+
+    override suspend fun submitSubscriptionChanges(subscriptions: List<Subscription>) {
+        Database.localSubscriptionDao().updateAll(subscriptions.map {
```
Bnyro, line 97: Please add a small comment here that we're updating the database entries here for the case that a channel avatar/name/verification status changes, perhaps helps when re-viewing the code again in a year :)

FineFindus: Yeah, I meant to do this, but I forgot :)

FineFindus:
Hmm, weirdly it only worked for one out of the two outdated channels I have, the other one is still using the old avatar. No idea what's going on there :shrug:

Bnyro (approved):

Bnyro:
> Hmm, weirdly it only worked for one out of the two outdated channels I have, the other one is still using the old avatar. No idea what's going on there 🤷
>
Maybe for that channel only the RSS feed was loaded and not the whole channel info (i.e. return in L133 at LocalFeedRepository).

PR #7913, opened 2025-11-12
feat(video player): improve touch gesture handling

dhanuarf:
- Gesture detector now works properly.
- Toggling player controls now feels more responsive and more consistent
- Center gesture detection area is slightly increased to give more room
- `Swipe up` detection in `SingleViewTouchableMotionLayout` is now implemented in`PlayerGestureController` for consistency, since you cant detect touch event from parent view (without intercepting) if any of the child is consuming it.

And I think now we can implement  #7523, since gesture detection is now properly canceled when the motion layout is swiping.

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/listeners/PlayerGestureController.kt
+++ b/app/src/main/java/com/github/libretube/ui/listeners/PlayerGestureController.kt
@@ -141,8 +141,14 @@ class PlayerGestureController(activity: BaseActivity, private val listener: Play
                     val eventPositionPercentageX = e.x / width
 
                     when {
-                        eventPositionPercentageX < 0.35 -> listener.onDoubleTapLeftScreen()
-                        eventPositionPercentageX > 0.65 -> listener.onDoubleTapRightScreen()
+                        eventPositionPercentageX < LEFT_AREA_VIEW_PERCENTAGE -> {
+                            listener.onDoubleTapLeftScreen()
+                        }
+
+                        eventPositionPercentageX > RIGHT_AREA_VIEW_PERCENTAGE -> {
+                            listener.onDoubleTapRightScreen()
+                        }
+
                         else -> listener.onDoubleTapCenterScreen()
                     }
                 }
@@ -176,8 +182,8 @@ class PlayerGestureController(activity: BaseActivity, private val listener: Play
             isMoving = true
 
             when {
-                e1.x < width * 0.35 -> if (isFullscreen) listener.onSwipeLeftScreen(distanceY)
-                e1.x > width * 0.65 -> if (isFullscreen) listener.onSwipeRightScreen(distanceY)
+                e1.x < width * LEFT_AREA_VIEW_PERCENTAGE -> if (isFullscreen) listener.onSwipeLeftScreen(distanceY)
+                e1.x > width * RIGHT_AREA_VIEW_PERCENTAGE -> if (isFullscreen) listener.onSwipeRightScreen(distanceY)
                 else -> listener.onSwipeCenterScreen(distanceY)
             }
 
@@ -188,5 +194,7 @@ class PlayerGestureController(activity: BaseActivity, private val listener: Play
     companion object {
         private const val MOVEMENT_THRESHOLD = 30
         private const val BORDER_THRESHOLD = 90
+        private const val LEFT_AREA_VIEW_PERCENTAGE = 0.35f
+        private const val RIGHT_AREA_VIEW_PERCENTAGE = 0.65f
     }
 }


dhanuarf:
Oh no, just found a bug. If you have description opened you cant scroll properly since the touch is intercepted by the parent.

Converting this to draft, gotta check again tomorrow

Bnyro:
> Oh no, just found a bug. If you have description opened you cant scroll properly since the touch is intercepted by the parent.
>
That's why I haven't touched the code in a while, it's just very frustrating...

dhanuarf:
It should work properly now. Feel free to tell me if there are any more regressions :) 
(I hope there isn't any... )

dhanuarf:
> That's why I haven't touched the code in a while, it's just very frustrating...

Frustrating yet kinda fun actually :)

Figim:
I tested it. It looks very good. Thanks!

app/src/main/java/com/github/libretube/ui/views/CustomExoPlayerView.kt
```diff
@@ -131,18 +131,20 @@ abstract class CustomExoPlayerView(
...
-    private fun toggleController() {
-        if (isControllerFullyVisible) hideController() else showController()
+    private fun toggleController(show: Boolean? = null) {
+        val isShow = show?: !isControllerFullyVisible
+        if (isShow) showController() else hideController()
+    }
+
+    init {
+        playerGestureController = PlayerGestureController(context as BaseActivity, this)
+        brightnessHelper = BrightnessHelper(context as Activity)
```
Bnyro:
```suggestion
        playerGestureController = PlayerGestureController(activity, this)
        brightnessHelper = BrightnessHelper(activity)
```
nit

app/src/main/java/com/github/libretube/ui/views/SingleViewTouchableMotionLayout.kt
```diff
@@ -76,46 +74,75 @@ class SingleViewTouchableMotionLayout(context: Context, attributeSet: AttributeS
...
     }
 
-    private val gestureDetector = GestureDetector(context, Listener())
+    override fun onInterceptTouchEvent(event: MotionEvent?): Boolean {
+        when (event?.action) {
+            MotionEvent.ACTION_DOWN -> {
+                shouldInterceptTouchEvent = false
+
+                // never intercept the event if we're currently in audio player mode
+                if (viewToDetectTouch.id == R.id.audio_player_container){
```
Bnyro: Please make that a class variable, i.e. `private val isAudioPlayer by lazy { viewToDetectTouch.id == R.id.audio_player_container }`

app/src/main/java/com/github/libretube/ui/listeners/PlayerGestureController.kt
```diff
@@ -94,41 +116,44 @@ class PlayerGestureController(activity: BaseActivity, private val listener: Play
...
+                        // not considered a double tap
+                        return false
+                    }
+
+                    val (width, _) = listener.getViewMeasures()
+                    val eventPositionPercentageX = e.x / width
+
+                    when {
+                        eventPositionPercentageX < 0.35 -> listener.onDoubleTapLeftScreen()
+                        eventPositionPercentageX > 0.65 -> listener.onDoubleTapRightScreen()
```
Bnyro: 1. I think we should move that to a constant, i.e. `SIDE_AREA_SCREEN_PERCENTAGE`
2. You perhaps should have mentioned that you're changing the sizes of the touch area in the PR description :) (I'm fine with it though, but it'd be nice to mention such behavior changes)

app/src/main/java/com/github/libretube/ui/listeners/PlayerGestureController.kt
```diff
@@ -151,29 +174,18 @@ class PlayerGestureController(activity: BaseActivity, private val listener: Play
             }
 
             isMoving = true
-            wasClick = false
 
             when {
-                e1.x < width * 0.4 -> listener.onSwipeLeftScreen(distanceY)
-                e1.x > width * 0.6 -> listener.onSwipeRightScreen(distanceY)
+                e1.x < width * 0.35 -> if (isFullscreen) listener.onSwipeLeftScreen(distanceY)
```
Bnyro: use constant created above, se https://github.com/libre-tube/LibreTube/pull/7913/files#r2528658581

app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
```diff
@@ -896,12 +888,13 @@ class PlayerFragment : Fragment(R.layout.fragment_player), OnlinePlayerOptions {
     }
 
     /**
-     * Enable or disable fullscreen depending on the current state
+     * Enter/exit fullscreen or toggle it depending on the current state
      */
-    fun toggleFullscreen() {
+    fun toggleFullscreen(isFullscreen: Boolean? = null) {
```
Bnyro:
```suggestion
        fun toggleFullscreen(isFullscreen: Boolean = commonPlayerViewModel.isFullscreen.value == true) {
        binding.player.hideController()

        if (!isFullscreen) {
```

Bnyro:
That's definitely an improvement, thanks for working on that!

I couldn't find any issues while testing (and I don't know 100% how the gesture detection code worked), so the review is only about code quality.

app/src/main/java/com/github/libretube/ui/listeners/PlayerGestureController.kt
dhanuarf: oh yeah sorry about that. Because I made it into a separate commit with (description too) so I thought that was unnecessary to be mentioned in the pr description

dhanuarf:
> I couldn't find any issues while testing (and I don't know 100% how the gesture detection code worked), so the review is only about code quality.  

Feel free to ask me if you have any question about how the code works ;)

app/src/main/java/com/github/libretube/ui/views/SingleViewTouchableMotionLayout.kt
dhanuarf: Now that I looked deeper into this, the '_swipe down to close_' gesture when the player minimized doesnt work in audio player mode, even in the original too, since I just follow what was written.

After thinking about it for a while, do we really need this gesture though? Since we have a big  _'close'_ button on the right side. More touch gesture means more headache, and I think people mostly use the close button to stop the player instead of swiping down. So I'm thinking of removing it, what do you think?

dhanuarf: Umm actually, never mind. I'll just keep it there

Bnyro: > Now that I looked deeper into this, the 'swipe down to close' gesture when the player minimized doesnt work in audio player mode, even in the original too, since I just follow what was written.
>
Yes, it's been disabled because it was very buggy. We perhaps want to re-add in the future though.

Bnyro (approved):
Thank you!

Bnyro:
> Feel free to ask me if you have any question about how the code works ;)
>
It's pretty good to understand after your changes, I just can't tell what the code did before your changes...

I might come back to that offer when working on that part of the code next time. :)

brnwlshubh:
@Bnyro  Fast forward is not as smooth as Youtube

PR #7914, opened 2025-11-12
refactor: remove local feed videos when unsubscribing

FineFindus:
Moves the removal of local feed videos from refreshing the feed to when the user unsubscribes. This is more inline if how the users may expect the channel to be removed (as the videos could still be visible until the next refresh), as well as avoiding the database call on every refresh. I don't think there is any case, where unsubscribe function doesn't get called when unsubscribing from a channel, but I appreciate a second set of eyes taking a look.

### DIFF
--- a/app/src/main/java/com/github/libretube/services/OnlinePlayerService.kt
+++ b/app/src/main/java/com/github/libretube/services/OnlinePlayerService.kt
@@ -114,7 +114,7 @@ open class OnlinePlayerService : AbstractPlayerService() {
         isAudioOnlyPlayer = args.getBoolean(IntentData.audioOnly)
 
         // get the intent arguments
-        this.videoId = playerData.videoId
+        videoId = playerData.videoId
         playlistId = playerData.playlistId
         channelId = playerData.channelId
         startTimestampSeconds = playerData.timestamp


app/src/main/java/com/github/libretube/db/dao/SubscriptionsFeedDao.kt
```diff
@@ -24,8 +24,9 @@ interface SubscriptionsFeedDao {
     @Query("DELETE FROM feedItem WHERE (uploaded < :olderThan AND uploaded != -1)")
     suspend fun cleanUpOlderThan(olderThan: Long)
 
-    @Query("DELETE FROM feedItem WHERE uploaderUrl NOT IN (:channelUrls)")
-    suspend fun deleteAllExcept(channelUrls: List<String>)
+    // TODO: the /channel/ prefix is allowed for compatibility reasons and will be removed in the future
+    @Query("DELETE FROM feedItem WHERE uploaderUrl = :channelUrl OR uploaderUrl LIKE '/channel/' || :channelUrl")
```
FineFindus: Considering that the `/channel/` prefix has been deprecated for quite a while now (at least since the introduction of the local feed), maybe it's time to remove fully remove it?

Bnyro: We should likely write a database migration if we want to remove it, because I don't think anything actually changed the syntax from `/channel/:id` to `:id` in the database automatically so far (except if the user manually (un)subscribed).

FineFindus: Hmm, is that actually necessary? Unless I'm missing something, all the new items should already the version without the prefix, and the old ones should be removed by the `cleanUpOlderThan` function.

Bnyro: My bad you're right actually. I mixed sth up and thought we also stored the /channel prefix in the subscriptions dao (which we didn't)...

PR #7915, opened 2025-11-12
chore: remove dead piped instances linking

FineFindus:
Removes support for opening links from the dead piped instances. All of these instance have been dead for a while now, most of them completely are offline (including the website itself), whilst some still host other services or explicitly announced the shutdown of their piped instance.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -382,18 +382,8 @@
                 <data android:scheme="https" />
                 <data android:pathPrefix="/" />
 
-                <data android:host="piped.adminforge.de" />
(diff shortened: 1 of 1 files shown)


Bnyro (approved):

ULTRA-LEGENDS:
Are you planning to clean up the Piped service options soon? Most of its sites have become useless and no longer work the only option that runs smoothly is the Full Local Mode. Any new user of the app will see many useless

ULTRA-LEGENDS:
This is a picture of these settings.

FineFindus:
We're aware of the current issues surrounding piped, it's on my to-do list, but there hasn't been a decision made yet what will be done.

Issue #7831, opened 2025-10-27
The UI disappears after attempting to close the Chapters menu


dairfod:
### Steps to reproduce

1. Open video
2. Press on Chapters
3. Close it by swiping up like on video example

### Expected behavior

Expected: The menu should close normally without the interface disappearing.

### Actual behavior

Video example:

https://github.com/user-attachments/assets/953d8e27-17a5-4818-b3bb-c757cbe49c53

### LibreTube version

0.29.0

### Android version

Galaxy A52. Android 14. OneUI 6.1

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [x] I have written a short but informative title.
- [x] I will fill out all of the requested information in this form.

alihrhera:
okay i will try to reproduce the bug and fix it.

demeirgen:
@alihrhera thank you

PR #7916, opened 2025-11-12
Fix/UI disappears after attempting to close the chapters menu

alihrhera:
Summary

This pull request improves the behavior of the Chapters Bottom Sheet in the player view by fixing issues related to its dismissal and lifecycle handling.

- Dismiss chapters sheet when hidden
- Introduces a BottomSheetListener to enable communication between the player view and the chapters sheet.
- When the chapters sheet is manually swiped down, it now properly dismisses itself and updates the player view state.
- Fixes issue #7831.
- Relies on the framework to handle dialog recreation correctly during configuration changes.

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/sheets/ExpandablePlayerSheet.kt
+++ b/app/src/main/java/com/github/libretube/ui/sheets/ExpandablePlayerSheet.kt
@@ -1,14 +1,44 @@
 package com.github.libretube.ui.sheets
 
+import android.app.Dialog
 import android.os.Bundle
 import android.view.View
+import android.widget.FrameLayout
 import androidx.annotation.LayoutRes
 import androidx.fragment.app.activityViewModels
 import com.github.libretube.ui.models.CommonPlayerViewModel
+import com.google.android.material.bottomsheet.BottomSheetBehavior
+import com.google.android.material.bottomsheet.BottomSheetDialog
 
 abstract class ExpandablePlayerSheet(@LayoutRes layoutResId: Int) :
     UndimmedBottomSheet(layoutResId) {
     private val commonPlayerViewModel: CommonPlayerViewModel by activityViewModels()
+
+    override fun onCreateDialog(savedInstanceState: Bundle?): Dialog {
+        val dialog = super.onCreateDialog(savedInstanceState)
+        if (commonPlayerViewModel.isFullscreen.value == true) {
+            // prevent an issue where swiping outside of the bottom sheet would make
+            // the app unresponsive by disabling slide actions to dismiss the bottom sheet in fullscreen
+            dialog.setOnShowListener { dialogInterface ->
+                val bottomSheetDialog = dialogInterface as BottomSheetDialog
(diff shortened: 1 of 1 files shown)


Bnyro (changes requested):
This patch should perhaps only be applied when in fullscreen, otherwise this introduces the following bug:
- Open a video (not in fullscreen)
- Click the chapters menu
- You can no longer interact with the video, i.e. you can't pause it

alihrhera:
@Bnyro 
is there any update?

Bnyro:
I'll look into it during the weekend if I find the time for it, life is busy at the moment

app/src/main/java/com/github/libretube/ui/sheets/ChaptersBottomSheet.kt
```diff
@@ -15,20 +17,59 @@ import com.github.libretube.ui.extensions.onSystemInsets
...
+                }
+            }
+        }
+        return dialog
+    }
+
+    fun show(fragmentManager: FragmentManager, isFullScreenPlayer: Boolean) {
+        isFullScreen = isFullScreenPlayer
+        super.show(fragmentManager, null)
+    }
```
Bnyro: Instead, you should add
```kotlin
val playerViewModel: CommonPlayerViewModel by activityViewModels()
```
and listen for fullscreen changes with
```kotlin
commonPlayerViewModel.isFullscreen.observe(...) 
```

app/src/main/java/com/github/libretube/ui/sheets/ChaptersBottomSheet.kt
```diff
@@ -15,20 +17,59 @@ import com.github.libretube.ui.extensions.onSystemInsets
...
+                bottomSheet?.let {
+                    val behavior = BottomSheetBehavior.from(it)
+
+                    behavior.addBottomSheetCallback(object :
+                        BottomSheetBehavior.BottomSheetCallback() {
+                        override fun onStateChanged(bottomSheet: View, newState: Int) {
+                            when (newState) {
+                                BottomSheetBehavior.STATE_HIDDEN -> {
+                                    dismissAllowingStateLoss()
+                                    (activity as? BottomSheetListener)?.onBottomSheetDismissed()
```
Bnyro: That over-complicated listener logic is really unnecessary, you  can just call `dismiss` here instead of doing that roundtrip :)

Bnyro:
With the most recent commit from your PR, the changes no longer seem to work on my device unfortunately.

Btw, please move all the code into `ExpandablePlayerSheet` instead of `ChaptersBottomSheet`, so that other fragments that inherit from that class can also benefit from the changes.

alihrhera:
@Bnyro is there any updates?

ULTRA-LEGENDS:
@Bnyro could you please take a look at this pull request and see whether it’s ready to be merged into the app or not

Bnyro (approved):
I've rebased and cleaned up the code a bit, thanks for your contribution!

PR #7920, opened 2025-11-14
chore: remove public piped instances

### LINKED ISSUE #7833: Can't register for a piped account due to chain validation/unexpected JSON errors
### Steps to reproduce

When navigating yo settings/instance, getting chain validation failed error displayed. After that, When trying to register to any of the piped instances, I am getting an unexpected JSON token at offset 0 error.

Steps to reproduce:
0. Without a piped account already configured... 
1. Open app and navigate to settings/instance. Once there you will see a `Chain validation failed` error
2. Continue to try to register for a piped account with any instance. You will get an unexpected JSON token at offset 0 error. 

### Expected behavior

- You should be able to navigate to settings instance without the chain validation error
- You should be able to register for a piped account with any of the displayed servers

### Actual behavior

When navigating yo settings/instance, getting chain validation failed error displayed. After that, When trying to register to any of the piped instances, I am getting an unexpected JSON token at offset 0 error

### LibreTube version

0.29.0(63)

### Android version

Android 15

### Other details

_No response_

### Acknowledgements

### LINKED ISSUE #8224: Deprecate Piped support

The piped code adds unneeded attack surface and piped is barely maintaned.

### Other details

_No response_

### Acknowledgements

### LINKED ISSUE #8251: Make setup screen consistent

Remove the option at setup to pick between local and piped for streaming. Or add more options at the setup, because important settings like sponsorblock are in the settings menu. It feels out of place to ask for this specific setting only.

### Other details

_No response_

### Acknowledgements

FineFindus:
Drops support for public Piped instances. This is due to
-  only a few (2) public instances still being online, neither of them supports streaming
-  Piped development has stalled and due to increasing blocking from YouTube it's doubtful the situation will improve in the near future
- the instances API has been offline for months and the fallback list of active instances is unmaintained
- users encounter cryptic errors and failing signups due to this

This does not remove support for piped instances overall, custom instances are still supported, with the assumption that self-hosting users are knowledgeable enough to handle the issues themselves.

Future steps may include to merge the fetch instance with auth instance, as self-hosting users are unlikely to host two instance, and use one just for fetching channels.

Ref: https://github.com/libre-tube/LibreTube/pull/7915
Closes: https://github.com/libre-tube/LibreTube/issues/7833
Closes: https://github.com/libre-tube/LibreTube/issues/8224
Closes: https://github.com/libre-tube/LibreTube/issues/8251


ULTRA-LEGENDS:
"I have a small note: to make the Full Local Mode option the default instead of the Local Stream Extraction option where The app's home page will appear like this to any new user and from my point of view this is a mistake because a simple user doesn’t know much about the settings."

FineFindus:
Thinking about it, we probably should do a migration (to avoid users having the app not working anymore).

Bnyro (commented):
> only a few (2) public instances still being online, neither of them supports streaming
>
The thing is that these two still allow you to search and load other data more anonymously - so using the "local streams extraction" with a public Piped instance is still a valid use case imo.

ULTRA-LEGENDS:
But it may stop working soon or in the future and we do not wait until it stops or until we receive new issue reports to start making fixes again

FineFindus:
I need to correct myself, it's only one instance, piped.syncpundit.io, doesn't provide an API.

> still allow you to search and load other data more anonymously

Only partially, as the Piped backend uses an older extractor version, and as such doesn't have the fixes for videos with multiple creators, doesn't support content availability (thus still displays members-only videos), etc …

I'm also not opposed to someone using Piped for such requests, however, I would prefer if users had to add such instances themselves, to provide a natural barrier from stopping new users using Piped and having issues with it.

Figim:
We can continue this work in the future by using a site like Peertube. Peertube needs a user-friendly interface like LibreTube. There is also a concept called Peertube instances. This does not solve the YT problem but also simply changes the purpose and content of LibreTube. So there is no need for an alternative frontend title just for YT.  "An interface created for your favorite stream platform". You mention in the readme that you are using Extractor. 

Invidious did more work 
1. With the support of "Invidious Companion"  
https://github.com/iv-org/invidious/pull/4985
2. You then need to integrate it into your android application and configure the instance.

https://github.com/iv-org/invidious/releases/tag/v2.20250913.0

app/src/main/java/com/github/libretube/helpers/PreferenceHelper.kt
```diff
@@ -144,6 +144,9 @@ object PreferenceHelper {
             val speed = (currentSpeed * 4f).roundToInt() / 4f
             putString(PreferenceKeys.PLAYBACK_SPEED, speed.toString())
         },
+        PreferenceMigration(6, 7) {
+            remove("disable_video_image_proxy")
+        },
```
FineFindus, line 149: As discussed, I'm fairly certain that disabling the [proxy breaks](https://github.com/libre-tube/LibreTube/pull/7920/changes/0b8c65e91c97a26b0bea4d667ab29587a7ddb2c8), but it might be a good idea to test it anyway.

FineFindus:
I'm not sure if/how we should migrate existing users, we could either
- do nothing, everything will still work, but they will have to manually add the piped instance if they ever want to change to it again
- migrate them to full local mode, even if the currently have a working instance
- migrate the selected fetch instance to a custom instance (we're currently only storing the API URL, but not the Name or the frontend URL).

We may also want to redesign the instance settings page (it's not really for instances any more).

app/src/main/java/com/github/libretube/ui/preferences/InstanceSettings.kt
```diff
@@ -183,11 +176,13 @@ class InstanceSettings : BasePreferenceFragment() {
         MaterialAlertDialogBuilder(requireContext())
             .setTitle(preference.title)
             .setView(binding.root)
+            .setNeutralButton(R.string.addInstance) { _, _ ->
+                CreateCustomInstanceDialog().show(childFragmentManager, null)
```
Bnyro, line 180: Good idea 👍

Bnyro (approved):

Bnyro:
> - do nothing, everything will still work, but they will have to manually add the piped instance if they ever want to change to it again
> - migrate them to full local mode, even if the currently have a working instance
> - migrate the selected fetch instance to a custom instance (we're currently only storing the API URL, but not the Name or the frontend URL).
>
We probably shouldn't yet change the settings for users that have a Piped instance configured. I think there still are some people that use Piped for everything (except video streaming, i.e. `Local stream extraction` enabled) or for authentication / syncing. So I think it's the best to just not do any migration and keep their preferences as they were.

FineFindus:
I wanted to avoid situations, where a user had a public piped instance selected, and is now no longer able to select that instance again. But I guess it's also easy enough to add it as a custom one.

Refocus537:
Can we please revert this PR?

Piped is still massively useful even though streaming isn't possible.

It can still be used to fetch any channel's details without making a connection to YouTube. That is a huge plus! Also you can store your subscriptions and playlists on the Piped instance.

One public instance works really well (until local extractions were broken as well. #8443 this fix will come to Piped as well) and frankly removing this is a net loss for all of us. @FineFindus

FineFindus:
As noted in the description, it's still fully possible to use Piped instances, this merely removes the list of public instances. If you have a working instance, you can add it as a custom instance in the app.

Refocus537:
Thanks for the quick response.

I must say that I fail to see the justification for removing the list of public instances since the only instance available actually works. Why are you cutting quick access to instance(s)? Convenience is key here. With this change it's frustrating to find instances manually while knowing you could previously set it up in 3 seconds. This change may seem trivial to ponder about but it really isn't. The change has a cascading effect. The way I see it, this change will impact the already struggling Piped project because as we know only LibreTube enables a hybrid-use approach with local stream extraction (video data from YouTube) and rest of the information fetched from Piped which works wonders. This use case scenario reminds me of the earlier days of Invidious where Invidious proxied everything but video data and you explicitly had to enable video proxying in Invidious settings to proxy the video through Invidious instance as well.

Distinguishing factor was always Piped and it still works great. Instead of removing the public instance list I propose reorganizing, UI overhaul, simplification and overall better UX. 

* Change Settings>'Instance' name to something more clear and appropriate for example Piped or Piped Proxy
* Clarify wording
  - Mode of operation to Fetch source settings
  - Instead of explaining the Full local mode option using the below text, change "Full local mode" to something clearer like Fetch everything from YouTube
  - Local feed extraction to something more intuitive
* Remove "Custom" under Instance heading as you can already add and/or select instances with "Choose..."
* Simplified onboarding screen for first time installs

These are just a few basic thoughts off the top of my head, but I'm sure you get the gist.
I hope you consider my sincere thoughts.

FineFindus:
Thanks for your feedback. The reasons for removal are also listed in the description. If you want to see improved Piped support in LibreTube, I would suggest to start by improving Piped itself first.

> Distinguishing factor was always Piped and it still works great. Instead of removing the public instance list I propose reorganizing, UI overhaul, simplification and overall better UX.

The page is already being reworked as part of https://github.com/libre-tube/LibreTube/pull/8367.

PR #7923, opened 2025-11-16
Update README.md

Figim:
I actually explained how LibreTube handles requests based on information read on Newpipe. We should have mentioned using Extractor and Internal API management even if there is a direct connection to "You Tube". In fact, Extractor does everything and LibreTube uses it.  

https://github.com/TeamNewPipe/NewPipe?tab=readme-ov-file#description

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,8 +151,7 @@ Contributions in any form are welcome!
 Differences to NewPipe
 </h2>
 
-With LibreTube, you have the choice to either send all requests directly to YouTube or proxy them via Piped for better privacy. Piped acts as a middleman server between you and YouTube in this case, which prevents YouTube from accessing personal information such as your IP address. 
-Additionally, using Piped allows you to sync your subscriptions between LibreTube and Piped, which can be used on desktop too.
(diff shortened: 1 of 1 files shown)


FineFindus (commented):
Thanks for the changes, however I don't see why a user should (have to) know how the application accesses information. It is entirly irrelevant to them how the extractor works, the only important information is whether the requests are proxied through Piped, which is already explained in the current version.

Figim:
> Thanks for the changes, however I don't see why a user should (have to) know how the application accesses information. It is entirly irrelevant to them how the extractor works, the only important information is whether the requests are proxied through Piped, which is already explained in the current version.

This is just wrong. LibreTube does not connect to Yt directly with Innertube APi.

FineFindus:
I'm not sure what you think Innertube is, but [it's the internal YouTube API ](https://www.fastcompany.com/3044995/to-take-on-hbo-and-netflix-youtube-had-to-rewire-itself), used by the official clients/frontends to display data, not any third-party API. That is, it literally _is_ YouTube.

PR #7928, opened 2025-11-17
feat: Use online player for offline playback

xlash123:
This patch allows for using the standard online player `PlayerFragment` for both online and offline video playback. This will replace the need for the `OfflinePlayerActivity` and allow for a single method of video playback.

The way I did this was by merging the functionality of the `OfflinePlayerService` into the `OnlinePlayerService`. I also used the metadata from the `StreamItem` in the `PlayerFragment` to act as a the common interface for accessing media metadata, since `Download` already supported a `.toStreamItem()` function.

I've tested the following related parts to ensure working functionality:
- `PlayOfflineDialog` still works for selecting which version of the video to play
- Switching between audio and video still works as seamlessly as before
- Selecting the video/audio from the downloads page will not prompt the `PlayOfflineDialog`

This is my first time really getting into Android development (and first time using Kotlin), so if you'd like me to make changes to clean this up, I am happy to do so and am open to suggestions on this. I hope you find this useful!

### DIFF
--- a/app/src/main/java/com/github/libretube/helpers/BackgroundHelper.kt
+++ b/app/src/main/java/com/github/libretube/helpers/BackgroundHelper.kt
@@ -50,7 +50,9 @@ object BackgroundHelper {
         startMediaService(
             context,
             OnlinePlayerService::class.java,
-            bundleOf(IntentData.playerData to playerData, IntentData.audioOnly to true)
+            bundleOf(
+                IntentData.playerData to playerData, IntentData.audioOnly to true
+            )
         )
     }
 
@@ -79,8 +81,12 @@ object BackgroundHelper {
         downloadTab: DownloadTab,
         shuffle: Boolean = false
     ) {
-        // whether the service is started from the MainActivity or NoInternetActivity
-        val noInternet = ContextHelper.tryUnwrapActivity<NoInternetActivity>(context) != null
+        // close the previous video player if open
+        val fragmentManager =
+            ContextHelper.unwrapActivity<MainActivity>(context).supportFragmentManager
+        fragmentManager.fragments.firstOrNull { it is PlayerFragment }?.let {
+            fragmentManager.commit { remove(it) }
+        }
 
         val playerData = PlayerData(videoId!!)
 
@@ -88,7 +94,6 @@ object BackgroundHelper {
             IntentData.playerData to playerData,
             IntentData.shuffle to shuffle,
             IntentData.downloadTab to downloadTab,
-            IntentData.noInternet to noInternet,
             IntentData.audioOnly to true,
             IntentData.isPlayingOffline to true,
         )
--- a/app/src/main/java/com/github/libretube/services/OnlinePlayerService.kt
+++ b/app/src/main/java/com/github/libretube/services/OnlinePlayerService.kt
@@ -28,6 +28,7 @@ import com.github.libretube.constants.IntentData
 import com.github.libretube.db.DatabaseHelper
 import com.github.libretube.db.DatabaseHolder.Database
 import com.github.libretube.db.obj.DownloadWithItems
+import com.github.libretube.db.obj.filterByTab
 import com.github.libretube.enums.FileType
 import com.github.libretube.enums.PlayerCommand
 import com.github.libretube.enums.SbSkipOptions
@@ -48,11 +49,13 @@ import com.github.libretube.parcelable.PlayerData
 import com.github.libretube.util.DeArrowUtil
 import com.github.libretube.util.PlayingQueue
 import com.github.libretube.util.YoutubeHlsPlaylistParser
+import com.github.libretube.ui.fragments.DownloadTab
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.cancelAndJoin
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.runBlocking
 import kotlinx.coroutines.withContext
 import kotlinx.serialization.encodeToString
 import kotlin.io.path.exists
@@ -83,6 +86,12 @@ open class OnlinePlayerService : AbstractPlayerService() {
     private var downloadWithItems: DownloadWithItems? = null
 
     private var isOnline: Boolean = true
+
+    /** Shuffle offline audio tracks */
+    private var shuffle: Boolean = false
+    /** The download tab we came from */
+    private var downloadTab: DownloadTab? = null
+
 
     // SponsorBlock Segment data
     private var sponsorBlockAutoSkip = true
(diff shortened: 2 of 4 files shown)


Figim:
**Problems and Errors:** 

1. **You have turned off your internet connection. The app crashes when you play the downloaded video.** 

**Also, if there is no internet, you may need to gray out some buttons to avoid errors.** 
**Quality, Audio track, Captions, Download, Comments, Subscribe**


https://github.com/user-attachments/assets/2cb46668-6d95-4241-81d9-1e9766fe5b1c

**Crash log:**

    
[Crashlog.docx](https://github.com/user-attachments/files/23593593/Crashlog.docx)


2. **If there is internet, YouTube-like description (currently not clickable) , subscription and view counts, and related videos should load properly. But it is not done.**


https://github.com/user-attachments/assets/ef798a7b-0ebe-4b69-b7a1-9bba364e493b




3. ** **If the internet is open, play the downloaded video. Error when clicking the Caption button**** 


https://github.com/user-attachments/assets/6db576e3-7f22-4db8-91d5-90d46195dfba

[Crashlog2.docx](https://github.com/user-attachments/files/23593611/Crashlog2.docx)


4. **The player does not work when switching from offline player to online player** 


https://github.com/user-attachments/assets/2f528e96-f136-4e7c-8230-71d833d3087e

5. **Cannot switch from notification to Offline audio player if there is no internet connection**

https://github.com/user-attachments/assets/adb9563f-3575-4377-ad0e-f65e2afda195

xlash123:
I believe I have fixed those problems. It should prefer to use the metadata from the online `Streams` object, else fallback on the downloaded metadata.

The `PlayerFragment` (and likely other components) contain many references to the `MainActivity`, so instead of trying to make the `NoInternetActivity` work with this, I had the `NoInternetFragment` load into the nav if no internet was detected at launch.

I also found that offline shuffle was crashing and fixed that too.

Figim:
1. **Video description is not clickable ( Not loading). This is problem for videos that are downloaded or not.**


https://github.com/user-attachments/assets/b87fea61-81fd-4dbc-9ebf-fb1fe3a09b3a



2. **If I play the currently downloaded audio ( (You are playing it now.)  from the watch history or switch to the "switch to video player" button in the audio player, either the player is dark or the player is dark and related videos etc. content are not loading. (internet is open). If there is internet, even if this audio is the case, when switching to a video player, it should play the online video instead of the audio and load the content.**


https://github.com/user-attachments/assets/350b8807-55ed-4eb3-9bf3-5c87e6788c1d


https://github.com/user-attachments/assets/d77d0441-5779-4e04-90f7-4125cc34931b





3. **When switching from offline video player to offline audio player, the audio player does not load if the internet is on or off.**


https://github.com/user-attachments/assets/359da424-6c59-4d50-8d7f-73c1e4272aa4




4. **Error when clicking the Caption button in the offline video player while not connected to the internet**


https://github.com/user-attachments/assets/d47a050d-30aa-473d-8e4f-eb85229cac42



5. **If you not connected to the internet, you played a video that was not downloaded from your watch history. Click the "Background" button. Then, when you click the "Switch to online video player" button, an error occurred.** 


https://github.com/user-attachments/assets/ac678b9f-0a2e-4f44-a59d-049d4ca0bd1c



6. **Videos are not added to the queue when playing online videos from an offline player**


https://github.com/user-attachments/assets/a0121663-bc17-4931-8d15-a20badcf6948

7. **When the internet closed and the downloaded video is played as "play in the background" and then switched to the video player with the "video switch" button, the offline video does not play and the loading is endless.**



https://github.com/user-attachments/assets/ddaeb2a1-5f04-4027-8b50-e569bf3ec149

8. **If the "Audio only mode" option is enabled, the audio player loads endlessly when playing a video downloaded without internet.** No problem playing with the internet 


https://github.com/user-attachments/assets/7cb18177-f466-4a58-ad1a-1f7ca85bb7bd


9. **It's not your fault. But it would be great if issues #7936 and #7937 were also resolved with this PR?**
**(Bnyro fixed #7936 and #7937)**

Bnyro (commented):
Hey, thanks for the PR! I think this is a definitely a good idea user experience-wise.

> The way I did this was by merging the functionality of the OfflinePlayerService into the OnlinePlayerService
>
I see why you did that, but that will make it very complicated to maintain that code in the future.

First, the `StreamItem` type is not the one perfect fit here, we should instead use the `Streams` type because it already contains that information so you don't have to edit `StreamItem.kt`. The information in `StreamItem.kt` is only meant for video objects from search results or similar places, which don't contain any information like license, metaInfo, ...

The better approach would be the following:
- Create some more abstract methods in `AbstractPlayerService` that work with the `Streams` type
- Implement them in both the `OnlinePlayerService` and the `OfflinePlayerService`
- Make `PlayerFragment.kt` work with the `AbstractPlayerService` class instead of the `OnlinePlayerService` class

Let me know if you have some questions about how that would work, don't hesitate to ask :)

xlash123:
I appreciate the feedback; always good to know the right way to implement this to enable good maintainability. I went into this just trying to get it working so I could show off the vision and get insight on how to better implement this.

I like the idea of implementing through the `AbstractPlayerService` and can see how that would make things easier for maintaining. For conveying the metadata of these downloaded streams, is the idea to convert the data within `DownloadWithItems` into the `Streams` format, and then use that to implement any new `Streams`-based methods to be created in the `AbstractPlayerService` by the `OfflinePlayerService`?

Figim:
> I appreciate the feedback; always good to know the right way to implement this to enable good maintainability. I went into this just trying to get it working so I could show off the vision and get insight on how to better implement this.
> 
> I like the idea of implementing through the `AbstractPlayerService` and can see how that would make things easier for maintaining. For conveying the metadata of these downloaded streams, is the idea to convert the data within `DownloadWithItems` into the `Streams` format, and then use that to implement any new `Streams`-based methods to be created in the `AbstractPlayerService` by the `OfflinePlayerService`?

I have tested again and would you please take a look at the above?

Bnyro:
> I like the idea of implementing through the `AbstractPlayerService` and can see how that would make things easier for maintaining. For conveying the metadata of these downloaded streams, is the idea to convert the data within `DownloadWithItems` into the `Streams` format, and then use that to implement any new `Streams`-based methods to be created in the `AbstractPlayerService` by the `OfflinePlayerService`?
>
Yes, exactly. We could also create a different data class instead of using `Streams` because `Streams` additionally contains `audioStreams` and `videoStreams` which we don't need here, but I think that would just overcomplicate things here.

xlash123:
@Bnyro understood, thanks for the guidance!

@Figim I see your tests; thanks for finding those! I'll tackle those after Bynro's refactoring suggestions.

Figim:
@xlash123 Is this PR still active?

xlash123:
Yes, I just haven't had much free time recently to devote to this unfortunately.

Bnyro:
Inspired by your PR, I've started a completely new implementation of your idea with https://github.com/libre-tube/LibreTube/pull/8206. Help with testing would be very appreciated :)

Because of that, I'm closing your PR now. Thanks for your contribution!

PR #7938, opened 2025-11-19
fix(video player): preview frame jumping around the screen when seeking

dhanuarf:
Fix seekbar preview frame jumping around the screen due to previous coroutines are not being canceled when scrubbing the timebar.

UI changes:
The preview frame now moves as soon as the finger moves (without waiting for the frame image to finish fetching) and a little progress indicator is added below the preview frame to serve as a ui feedback when the frame is being loaded

https://github.com/user-attachments/assets/936395ab-3960-400f-aa94-912ae591f0f0

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/listeners/SeekbarPreviewListener.kt
+++ b/app/src/main/java/com/github/libretube/ui/listeners/SeekbarPreviewListener.kt
@@ -4,6 +4,7 @@ import android.text.format.DateUtils
 import android.view.View
 import android.view.ViewGroup.MarginLayoutParams
 import androidx.core.view.isGone
+import androidx.core.view.isInvisible
 import androidx.core.view.isVisible
 import androidx.core.view.updateLayoutParams
 import androidx.media3.common.util.UnstableApi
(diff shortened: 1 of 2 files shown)


Bnyro (approved):
Thank you!

We definitely also have some more places where we still have coroutines doing stuff in the background instead of canceling them - I probably already fixed 3 or 4 of such occurences recently, so there has been a bit of an issue with handling coroutines in the past :)

I've been a bit concerned about whether the progress bar matches the UI well and feels intuitive at first, but from testing it I can agree it's better than not having one and it's not disturbing in any way :rocket:.

Again, thank you!

FineFindus:
> a little progress indicator is added below the preview frame to serve as a ui feedback when the frame is being loaded

Maybe it would better to (also) display a lower-quality preview frame first and load the high-quality one in the background?
The extractor already provides multiple sets of preview frames, however we currently always default to the highest quality.
 
https://github.com/libre-tube/LibreTube/blob/67d82c28ac08fc43d1b544fd5f213ae0c1fb3ae2/app/src/main/java/com/github/libretube/util/OnlineTimeFrameReceiver.kt#L43-L44

Bnyro:
I'm not sure if that's worth it because this also slows down loading previews you actually want to see and makes the code more complicated, but just my opinion.

PR #7952, opened 2025-11-22
refactor: add minimum of distance in pixels in `SeekbarPreviewListener`

dhanuarf:
Add minimum seeking distance in pixels required to load another preview frame.

Reasoning:
If we only rely on minimum distance of 5 seconds, if the video duration is quite long (hours), 5 seconds is really a short distance, especially on small screen devices. 

Currently I set the distance to 8dp, let me know if I should decrease or increase it.

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/listeners/SeekbarPreviewListener.kt
+++ b/app/src/main/java/com/github/libretube/ui/listeners/SeekbarPreviewListener.kt
@@ -10,14 +10,14 @@ import androidx.core.view.updateLayoutParams
 import androidx.media3.common.util.UnstableApi
 import androidx.media3.ui.TimeBar
 import com.github.libretube.databinding.ExoStyledPlayerControlViewBinding
+import com.github.libretube.extensions.dpToPx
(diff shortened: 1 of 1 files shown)


dhanuarf:
> I don't think this really reduces network workload because YouTube always provides multiple preview frames in a single image (e.g. https://i.ytimg.com/sb/d4JL1IDupYI/storyboard3_L2/M5.jpg?sqp=-oaymwENSDfyq4qpAwVwAcABBqLzl_8DBgjAx4LJBg==&sigh=rs$AOn4CLAYD06EM13HZ5U5La83BD5vy3SF9Q).
> 
> This means most of the image frames we load are already cached and it doesn't make sense to not display them to the user as we already loaded them anyways?

Oh yeah sorry, I didnt check what `getFrameAtTime()` actually does. I thought it fetched a preview from the server every time.
Sorry for wasting your time, I think I'll close this.

Bnyro:
> Sorry for wasting your time
>
No worries, it also took me some time to realize that, I was already on my way to approve the PR :)

Issue #7832, opened 2025-10-28
Translated titles in video overview


dennisfrijlink:
### Describe your suggested feature

For some reason when scrolling through the videos, for example in the subscriptions, the titles of the videos are always English (my language settings are Dutch - NL), but when opening a video it shows the translated title.

I think this is because the API of YouTube itself(?) Maybe a nice feature  for the future to implement a translated overview instead of only a translated title when opening the video.

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [x] I have written a short but informative title.
- [x] I will fill out all of the requested information in this form.

Bnyro:
We use the NewPipeExtractor for loading video titles, so we don't have any control about it.

Tracking issue: https://github.com/TeamNewPipe/NewPipe/issues/3089#issuecomment-3417197776

D3SOX:
This is still an issue when viewing playlists. Couldn't you integrate something similar to what the [YouTube-No-Translation add-on](https://github.com/YouG-o/YouTube-No-Translation) does? Perhaps also with a setting for providing my API key

Issue #7343, opened 2025-04-25
Some video titles, descriptions, or chapters are not shown in their original language.


luckkmaxx:
### Steps to reproduce

I'm Spanish and I use the app in Spanish, with a feed full of Spanish channels. However, some channels (always the same ones) show the video titles in English in the feed, even though the original titles are in Spanish.
When I open one of these videos, the player displays the correct title in Spanish, but the description and chapter titles are in English.
This doesn’t happen all the time. Sometimes, if I close and reopen the app, the same videos then show the correct Spanish title — but the description and chapter titles remain in English.

### Expected behavior

Always show titles, descriptions and chapters in their original language. 

### Actual behavior

The content is translated to english. 

### LibreTube version

0.28.0 and previous too

### Android version

Android 14

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [x] I have written a short but informative title.
- [x] I will fill out all of the requested information in this form.

Bnyro:
That's a NewPipe Extractor issue and nothing LibreTube can fix per se, please open an issue at their repository instead.

ghost:
On the channel video's don't show the original, but in the search results they do for me. E.g. the YouTube channel Harry Vermeegen.

ghost:
Should be re-opend since it seems a client issue: https://github.com/TeamNewPipe/NewPipe/issues/3185#issuecomment-2888879828. Which makes sense since PipePipe already defaults to original text with a option by default.

Issue #3208, opened 2023-02-28
The video title is displayed in English


Qaz-7:
### Steps to reproduce

1.The language and location of the device is set to Turkey
2.The video title is in Turkish.(on YouTube)
3.LibreTube translates the title into English
4.The player is in Turkish

### Expected behavior

The video title should be displayed in Turkish on the home page

### Actual behavior

Translated into English.

### LibreTube version

0.12.0 Latest debug

### Android version

Android 12

### Other details

https://user-images.githubusercontent.com/124310688/221776952-951a4853-8147-4114-9dfb-a2607bbca904.mp4

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

Bnyro:
Titles are based on the instance's location.

Issue #2224, opened 2022-12-04
the video titles are in the wrong language


Darkempire78:
### Steps to reproduce

1. Set your language fo French for example
2. Go to a french channel like "Inoxtag"
3. See that video titles are in english
4. Go to the "misterbest" channel
5. See that titles are in english too

### Expected behavior

French videos should displays the French title and the english videos that have french titles should be displayed in french too (on YouTube, misterbeast videos have a translated title)

### Actual behavior

Titles are always in english even if there is a translation in your device language

### LibreTube version

0.9.0

### Android version

Android 13

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

Bnyro:
The language of the titles is based on the region of the instance you're using, without support by the NewPipe Extractor for it there's no way to get localized titles.

PR #7957, opened 2025-11-22
feat(LocalFeed): prefer untranslated video title if available

FineFindus:
Fixes an issue, where the local subscription feed could show automatically translated video titles. As the RSS feed seems to contain the untranslated title, prefer that over the translated Innertube title.

Note that this only affects the (local) subscription feed, where we already fetch the RSS feed, and not the individual channel pages. 

Closes: https://github.com/libre-tube/LibreTube/issues/7832

### DIFF
--- a/app/src/main/java/com/github/libretube/api/NewPipeMediaServiceRepository.kt
+++ b/app/src/main/java/com/github/libretube/api/NewPipeMediaServiceRepository.kt
@@ -102,7 +102,8 @@ fun StreamInfoItem.toStreamItem(
     return StreamItem(
         type = TYPE_STREAM,
         url = url.toID(),
-        title = name,
+        // if available prefer the RSS feed title, since it's untranslated
+        title = feedInfo?.name ?: name,
         uploaded = uploadDate?.offsetDateTime()?.toEpochSecond()?.times(1000) ?: -1,
         uploadedDate = textualUploadDate ?: uploadDate?.offsetDateTime()?.toLocalDateTime()
             ?.toLocalDate()


FineFindus:
Would be great if someone else could also confirms that this fixes it, and I didn't just hit an edge-case/a-b test (cc @gewuagewua, @dennisfrijlink)

ULTRA-LEGENDS:
"I tried it, but for me, the video title still appears in English instead of Arabic, especially when you go to the channel or the video itself-the video title is in English instead of Arabic. However, when searching for the video, it appears in Arabic in the search results.

FineFindus:
Thanks for testing. As noted in the description this only applies to the local subscription feed (and only for new videos). The video, channel, and search pages are handled differently and cannot be (easily) fixed.

ULTRA-LEGENDS:
Hi, thanks for your response. I understand that the current fix only applies to the local subscription feed.

I’d like to suggest an enhancement: it would be very useful if the app could automatically translate only videos that aren’t yet translated into Arabic. This way, any new video or video without an existing Arabic title would show in Arabic across all pages — channel pages, individual video pages, and subscriptions — while videos already translated remain unchanged.
A simple approach could be:
1. Check if a video already has an Arabic title stored locally.
2. If not, translate the title using a reliable service (like Google Translate or DeepL).
3. Cache the translation to avoid repeated requests.
4. Display the translated title consistently in all parts of the app.
This would improve the user experience significantly without affecting performance or requiring translation for videos that are already in Arabic.              Thank you for your work on the app and its improvements I understand the difficulties but if this idea is feasible it would be a fantastic solution

dhanuarf:
Maybe add an option to toggle it? 

Currently I subbed to some channels that I dont understand the language, but they provide english translated title, description, and  video subtitle. With this change I wont understand what the video is about at a glance unless I click on it.

Actually, I want the opposite of this change. I want the video title in player screen also to be translated (if it's possible). And then add an option to toggle if we want translated or untranslated title on videos.

Bnyro:
> Maybe add an option to toggle it?
>
I don't feel like this is important enough to add yet another (unnecessary) preference, so we should agree on a behavior here.

I personally don't have any opinion about that.

FineFindus:
I think considering how many issues have been opened about this (https://github.com/libre-tube/LibreTube/discussions/7958, https://github.com/libre-tube/LibreTube/issues/7832, https://github.com/libre-tube/LibreTube/issues/7343, https://github.com/libre-tube/LibreTube/issues/3208, https://github.com/libre-tube/LibreTube/issues/2224, …) and are likely to still get as this isn't fixed it would be better to use the original titles. This would also be consistent with how the video is displayed when opened, having the video title change when clicking on it is quite unexpected.

> Actually, I want the opposite of this change. I want the video title in player screen also to be translated (if it's possible). And then add an option to toggle if we want translated or untranslated title on videos.

Having the application support auto-translations would be great, however it needs to be fixed on the extractor side.

Bnyro (approved):

Issue #7850, opened 2025-11-02
Support hdr video as well as av1


m3talg3ar:
### Describe your suggested feature

Support av1 and hdr vids

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [x] I have written a short but informative title.
- [x] I will fill out all of the requested information in this form.

Figim:
No Newpipe extractor support

PR #7959, opened 2025-11-23
feat(Preferences): remove codec preferences

FineFindus:
Removes support for setting the preferred audio/video codec. This preferences can be quite confusing to new, as well experienced users, since it's not immediately clear which choice is optimal (not that there necessarily is one). It also prohibited more modern codecs (such as AV1) being chosen automatically, when the preference was not set to `all`, leading to a worse user experience

Overall it's better to let the player, which has more information, decide what format to use.

Ref: https://github.com/libre-tube/LibreTube/issues/7850

### DIFF
--- a/app/src/main/java/com/github/libretube/constants/PreferenceKeys.kt
+++ b/app/src/main/java/com/github/libretube/constants/PreferenceKeys.kt
@@ -73,8 +73,6 @@ object PreferenceKeys {
     const val DOUBLE_TAP_TO_SEEK = "double_tap_seek"
     const val ALTERNATIVE_PIP_CONTROLS = "alternative_pip_controls"
     const val SKIP_SILENCE = "skip_silence"
-    const val ENABLED_VIDEO_CODECS = "video_codecs"
-    const val ENABLED_AUDIO_CODECS = "audio_codecs"
     const val AUTOPLAY_COUNTDOWN = "autoplay_countdown"
     const val AUTO_FULLSCREEN_SHORTS = "auto_fullscreen_shorts"
     const val PLAY_AUTOMATICALLY = "play_automatically"
--- a/app/src/main/java/com/github/libretube/helpers/PlayerHelper.kt
+++ b/app/src/main/java/com/github/libretube/helpers/PlayerHelper.kt
@@ -331,18 +331,6 @@ object PlayerHelper {
             false
         )
 
-    private val enabledVideoCodecs: String
-        get() = PreferenceHelper.getString(
-            PreferenceKeys.ENABLED_VIDEO_CODECS,
-            "all"
-        )
-
-    private val enabledAudioCodecs: String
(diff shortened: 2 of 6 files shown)


ULTRA-LEGENDS:
I suggest merging the Audio and Video section with the History section or vice versa as the Audio and Video section is nearly empty and combining them would streamline the settings and make better use of the available space in the app Settings

Bnyro:
> I suggest merging the Audio and Video section with the History section or vice versa as the Audio and Video section is nearly empty and combining them would streamline the settings and make better use of the available space in the app Settings
>
Wouldn't that only make sense if these categories were actually related to each other? Audio/video and history settings don't have much in common, so this is probably not intuitive to users. Instead, the audio/video section likely fits into the player section (which is however already very full of settings)

ULTRA-LEGENDS:
Yes, I actually considered merging the audio and video section with the player section, but I thought you would not agree to that because that section is full of settings. That is why I suggested merging it with the history section. And I do agree with merging it with the player
section if that is possible

ULTRA-LEGENDS:
Hi FineFindus I have a suggestion regarding AV1 support in the app:
Why don’t we integrate the new dav1d library introduced with Android 15 to provide an enhanced experience for devices running Android 11 and above?

Key benefits of adding dav1d:
1. Wider support for older and mid-range devices:
Currently (with codec options removed) AV1 only works on devices with hardware support, limiting the number of users who can watch high-quality content.
Adding dav1d allows the app to decode AV1 via software decoding, significantly expanding the range of supported devices

2. Improved user experience without complexity:
Users won’t need to manually select codecs or face performance issues with older formats like VP9 or H.264.
The player will automatically choose the best available codec for each device, ensuring a smooth and consistent experience

3 Enhanced performance and efficiency:
dav1d is known for its fast performance and lower CPU usage compared to older software solutions
It ensures smooth playback of high-resolution AV1 videos even on low- and mid-range devices
Supporting AV1 now ensures the app remains compatible with modern technologies and reduces the need for urgent updates in the future

FineFindus:
Thanks for the suggestion, please open new issues/discussion for this is the future, comments on a PR isn't really the right place for it.

>  Why don’t we integrate the new dav1d library introduced with Android 15 to provide an enhanced experience for devices running Android 11 and above?

That would require us taking on a lot of complexity for little to no benefit, while `dav1d` is an incredible library, software decoders are much, much slower than hardware decoders (and nearly every phone supports H.264/AVC hardware decoding, with many also offering hardware decoding of VP9), so using software decoding would most likely be a downgrade, both in terms of speed, but also battery life.
Additionally, the decoders that are shipped with Android by default are sandboxed, whilst ours wouldn't be.

>  The player will automatically choose the best available codec for each device, ensuring a smooth and consistent experience

This is already the case, we merely set the resolution and the player decides what format is best suited.

Issue #7907, opened 2025-11-11
Add progress indicator for import actions


FossLover0:
### Describe your suggested feature

Importing huge amounts of subscription or of playlist is currently a bit painfully because there is no progress indicator.
I would suggest to add a simple loading dialog or an ongoing notification.

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [x] I have written a short but informative title.
- [x] I will fill out all of the requested information in this form.

SE-Bilal-Ahmad:
See pull request #7973

PR #7973, opened 2025-11-29
feat: Added coroutine worker for importing watch history 

SE-Bilal-Ahmad:
1- Previously a coroutine was executed which imported watch history from youtube but there was no way to know the status of the import history i have a added a coroutine worker whose job is to import the watch history with a foreground notification.

2- I have also added dagger hilt in order to inject dependencies. I could have used other approaches but that wouldnt have been a clean solution

3- For now only watch history is being imported through coroutine worker. The reason is that we are inserting watch item one by one but for subscriptions one complete list is inserted which makes it nearly impossible to know the status of those imports however i will still look for a way to achieve that and for playlists too

Bug #7907

### DIFF
--- a/app/src/main/res/values-ar/strings.xml
+++ b/app/src/main/res/values-ar/strings.xml
@@ -525,12 +525,12 @@
     <string name="export_playlist">صدّر قائمة التشغيل</string>
     <string name="import_format_list_of_urls">قائمة عناوين URL أو معرّفات الفيديو</string>
     <string name="local_feed_extraction">استخراج موجز المحلية</string>
-    <string name="local_feed_extraction_summary">قم بجلب الخلاصة مباشرةً من YouTube قد يكون هذا أبطأ بشكل ملحوظ.</string>
+    <string name="local_feed_extraction_summary">اجلب الخلاصة مباشرةً من YouTube قد يكون هذا أبطأ بشكل ملحوظ.</string>
     <string name="update_information">تحديث المعلومات</string>
     <string name="never_show_again">لا تعرض هذا مجدداً</string>
     <string name="show_upcoming_videos">أظهِر الفيديوهات القادمة</string>
     <string name="full_local_mode">وضع الإتصال المحلي</string>
-    <string name="authentication">المصادقة</string>
+    <string name="authentication">الاستيثاق</string>
     <string name="mode_of_operation">طريقة التشغيل</string>
     <string name="full_local_mode_desc">جلب كل شيء مباشرة من YouTube دون استخدام Piped.</string>
     <string name="updating_feed">تحديث التغذية …</string>


SE-Bilal-Ahmad:
If someone wants i can add an ongoing notification for subscription and playlist imports however that will not be able to tell how much content to import has been left

app/src/main/java/com/github/libretube/factory/NotificationFactory.kt
```diff
@@ -0,0 +1,134 @@
...
+            .setContentText(context.getString(R.string.preparing))
+            .setOngoing(true)
+            .setOnlyAlertOnce(true)
+            .setSmallIcon(R.drawable.ic_launcher_lockscreen)
+            .setProgress(0, 0, false)
+        return builder.build()
+    }
+
+
+    private val notificationChannelId = "DemoNotificationChannelId"
```
Bnyro: ?

app/src/main/java/com/github/libretube/factory/NotificationFactory.kt
```diff
@@ -0,0 +1,134 @@
...
+            context.getString(R.string.resume),
+            ImportReceiver.createResumePendingIntent(context,uuid),
+        )
+    }
+
+
+    private fun createNotificationChannel() {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+
+            val notificationChannel = NotificationChannel(
```
Bnyro: Please use `NotificationChannelCompat` instead

app/src/main/java/com/github/libretube/helpers/ImportHelper.kt
Bnyro, line 1: why did you rename this file?

app/src/main/java/com/github/libretube/handler/ImportHandler.kt
```diff
@@ -0,0 +1,43 @@
...
+    private var cancel = MutableStateFlow(false)
+
+    val isPaused: Boolean
+        get() = paused.value
+
+    val isCancelled: Boolean
+        get() = cancel.value
+
+    fun pause() {
+        paused.value = true
```
Bnyro:
```suggestion
        paused.emit(true)
```
same below

app/src/main/java/com/github/libretube/LibreTubeApp.kt
```diff
@@ -94,13 +117,21 @@ class LibreTubeApp : Application() {
             .setDescription(getString(R.string.push_channel_description))
             .build()
 
+        val importChannel = NotificationChannelCompat.Builder(
+            IMPORT_CHANNEL_NAME,
+            NotificationManagerCompat.IMPORTANCE_DEFAULT
+        ).setName("Import Worker")
+            .setDescription("Shows a notification when importing")
```
Bnyro: These should be translatable

app/src/main/java/com/github/libretube/db/Converters.kt
```diff
@@ -18,11 +19,12 @@ object Converters {
     fun pathToString(path: Path?) = path?.toString()
 
     @TypeConverter
-    fun stringToPath(string: String?) = string?.let { Path(it) }
+    fun stringToPath(string: String?) = string?.let { Paths.get(it) }
```
Bnyro: why?

app/src/main/java/com/github/libretube/enums/ImportState.kt
```diff
@@ -0,0 +1,7 @@
+package com.github.libretube.enums
+
+enum class ImportState {
+    PAUSED,
+    RESUME,
+    CANCEL
```
Bnyro: if this should be a state, `resume` and `cancel` should not be an option since it doesn't represent any kind of state.

Instead you probably meant something like `PAUSED, RUNNING, CANCELED`

app/src/main/java/com/github/libretube/factory/NotificationFactory.kt
```diff
@@ -0,0 +1,134 @@
...
+class NotificationFactory @AssistedInject constructor(
+    @ApplicationContext val context: Context,
+    workManager: WorkManager,
+    @Assisted private val uuid: UUID
+) {
+
+    private val builder = NotificationCompat.Builder(context, IMPORT_CHANNEL_NAME)
+    private val mutex = Mutex()
+
+    fun createNotification(uuid: UUID): Notification {
```
Bnyro: `uuid` unused? Why would that parameter be needed anyways?

app/src/main/java/com/github/libretube/factory/NotificationFactory.kt
```diff
@@ -0,0 +1,134 @@
...
+import com.github.libretube.receivers.ImportReceiver
+import dagger.assisted.Assisted
+import dagger.assisted.AssistedFactory
+import dagger.assisted.AssistedInject
+import dagger.hilt.android.qualifiers.ApplicationContext
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
+import java.util.UUID
+
+private const val CHANNEL_ID_DEFAULT = "import_history"
```
Bnyro: This constant should be put inside `NotificationHelper`

app/src/main/java/com/github/libretube/factory/NotificationFactory.kt
```diff
@@ -0,0 +1,134 @@
...
+
+
+    suspend fun updateState(
+        currentState: Int = 0,
+        finalState: Int = 0,
+        importState: ImportState = ImportState.RESUME
+    ): Notification = mutex.withLock {
+        val percent =
+            if (finalState == 0) 0 else ((currentState.toFloat() / finalState) * 100).toInt()
+        builder.setContentText("$percent% completed")
```
Bnyro: This should be translatable

app/src/main/java/com/github/libretube/factory/NotificationFactory.kt
```diff
@@ -0,0 +1,134 @@
...
+        builder.setProgress(finalState, currentState, false)
+        builder.clearActions()
```
Bnyro: that's a noop?

app/src/main/java/com/github/libretube/receivers/ImportReceiver.kt
```diff
@@ -0,0 +1,89 @@
...
+import org.schabi.newpipe.extractor.timeago.patterns.id
+import java.util.UUID
+
+class ImportReceiver(
+    private val uuid: UUID,
+    private val importHandler: ImportHandler,
+    ) : BroadcastReceiver() {
+    override fun onReceive(context: Context?, intent: Intent?) {
+        val id = intent?.getStringExtra(EXTRA_UUID)
+        val parsedId: UUID? = try {
```
Bnyro: Why do you use `UUID`s instead of incrementing integer IDs like everywhere else?

app/src/main/java/com/github/libretube/receivers/ImportReceiver.kt
```diff
@@ -0,0 +1,89 @@
...
+
+        private fun createIntent(context: Context, action: String,uuid: UUID) = Intent(action)
+            .setData("$SCHEME://$uuid".toUri())
+            .setPackage(context.packageName)
+            .putExtra(EXTRA_UUID, uuid.toString())
+
+        const val ACTION_IMPORT_RESUME =
+            "com.github.libretube.receivers.ImportReceiver.ACTION_IMPORT_RESUME"
+        const val ACTION_IMPORT_PAUSE =
+            "com.github.libretube.receivers.ImportReceiver.ACTION_IMPORT_PAUSE"
```
Bnyro: These should use `context.packageName + ".receivers.ImportReceiver.ACTION_IMPORT_PAUSE"` instead of hardcoding the package name

gradle.properties
```diff
@@ -16,4 +16,5 @@ kotlin.code.style=official
 android.useAndroidX=true
 android.enableJetifier=true
 android.nonTransitiveRClass=true
-android.nonFinalResIds=true
\ No newline at end of file
+android.nonFinalResIds=true
+# Add this line if missing
```
Bnyro: ?

Bnyro (changes requested):
I generally think that the idea of putting the import work into a dedicated worker goes into the right direction, thanks for proposing.

However, we decided to not add Dagger/Hilt as a dependency, so please refactor your PR to work without it.

Apart from that I left some comments, please have a look at them. Without willing to attack you personally, this code looks a bit AI-generated .

We don't accept contributions that add much additional code that was  AI-generated  here. So please clarify that this is not the case or rewrite the AI-generated parts as we don't have any permission to license code that was stolen from other projects as GPL3.

app/src/main/java/com/github/libretube/factory/NotificationFactory.kt
SE-Bilal-Ahmad: I was working on another project called Kotatsu the approach of pausing and resuming the Work manager is where i got this from.

If you want i can write my own logic for pausing and resuming the work manager i am just afraid that my solution might now be better than theirs

app/src/main/java/com/github/libretube/factory/NotificationFactory.kt
SE-Bilal-Ahmad: I will refactor it to use NotificationChannelCompat

app/src/main/java/com/github/libretube/helpers/ImportHelper.kt
SE-Bilal-Ahmad: Okay that was be mistake i will rename it

app/src/main/java/com/github/libretube/factory/NotificationFactory.kt
Bnyro: The question is about whether you call it `DemoNotificationChannelId`, that looks like it's copied from somewhere else without adapting it to our app.

SE-Bilal-Ahmad: Yeah the name was actually copied 
Sorry about that

SE-Bilal-Ahmad:
@Bnyro do i need to rewrite the other code as well by that i meant the pausing handler since thats not AI generated the only Ai generated one is the notification part that too is in much smaller portion not all of it was Ai generated so please clarify

Bnyro:
Only the AI generated code.

SE-Bilal-Ahmad:
okay i will be rewriting it

gradle.properties
SE-Bilal-Ahmad: Comment removed

app/src/main/java/com/github/libretube/receivers/ImportReceiver.kt
SE-Bilal-Ahmad: Problem is that i am inside the companion object therefore i can't get the context there

app/src/main/java/com/github/libretube/factory/NotificationFactory.kt
SE-Bilal-Ahmad: It shows the percentage of watch history that has been imported

app/src/main/java/com/github/libretube/factory/NotificationFactory.kt
SE-Bilal-Ahmad: Has now been moved to string.xml

app/src/main/java/com/github/libretube/factory/NotificationFactory.kt
SE-Bilal-Ahmad: uuid now removed

app/src/main/java/com/github/libretube/enums/ImportState.kt
SE-Bilal-Ahmad: States now renamed

app/src/main/java/com/github/libretube/db/Converters.kt
SE-Bilal-Ahmad: Change was made by mistake change reverted

app/src/main/java/com/github/libretube/LibreTubeApp.kt
SE-Bilal-Ahmad: Now moved to string.xml

app/src/main/java/com/github/libretube/helpers/ImportHelper.kt
SE-Bilal-Ahmad: Change reverted

app/src/main/java/com/github/libretube/factory/NotificationFactory.kt
SE-Bilal-Ahmad: Change has now been made to use NotificationChannelCompat

app/src/main/java/com/github/libretube/factory/NotificationFactory.kt
SE-Bilal-Ahmad: AI code removed, Dagger hilt removed

app/src/main/java/com/github/libretube/receivers/ImportReceiver.kt
SE-Bilal-Ahmad: Suppose a user imports one watch history file and closes the app than again open up the app to import another watch history file than in thjat case there will be no way of knowing what the previous id was thats why i have used UUID

app/src/main/java/com/github/libretube/factory/NotificationFactory.kt
SE-Bilal-Ahmad: Since the contant is not being used so now has been removed

app/src/main/java/com/github/libretube/handler/ImportHandler.kt
SE-Bilal-Ahmad: Problem here is that emit is a suspending function which must be called from another suspending function or a coroutine and i am calling it from a broadcast reciever  method OnRecieve which is not a suspending function

app/src/main/java/com/github/libretube/receivers/ImportReceiver.kt
dhanuarf: Do we really need to define the whole package though? We only use this action only for this specific case, and only used inside our own app so I dont think we need that? Just a simple string is enough no?

Anyway, you can use `BuildConfig.APPLICATION_ID` to get package name in a constant

ULTRA-LEGENDS:
@Bnyro could you please take a look at this pull request and see whether it’s ready to be merged into the app or not

app/src/main/java/com/github/libretube/helpers/ImportHelper.kt
```diff
@@ -23,19 +21,20 @@ import com.github.libretube.obj.NewPipeSubscription
...
 import kotlinx.serialization.json.encodeToStream
 import java.util.stream.Collectors
 
 object ImportHelper {
-    private const val IMPORT_THUMBNAIL_QUALITY = "mqdefault"
-    private const val VIDEO_ID_LENGTH = 11
-    private const val YOUTUBE_IMG_URL = "https://img.youtube.com"
+    private val IMPORT_THUMBNAIL_QUALITY = "mqdefault"
+    private val VIDEO_ID_LENGTH = 11
+    private val YOUTUBE_IMG_URL = "https://img.youtube.com"
```
Bnyro: Why remove `const` here?

build.gradle.kts
```diff
@@ -10,7 +10,7 @@ buildscript {
         classpath(libs.kotlin.gradle.plugin)
         classpath(libs.kotlin.serialization)
         classpath(libs.androidx.navigation.safeargs)
-
+        classpath("com.google.dagger:hilt-android-gradle-plugin:2.41")
```
Bnyro: remove this please, we don't want to use dagger hilt

app/src/main/res/drawable/ic_action_pause.xml
Bnyro, line 1: We already have an `ic_pause`.

app/src/main/res/drawable/ic_action_resume.xml
Bnyro, line 1: We already have an `ic_play`.

app/src/main/java/com/github/libretube/receivers/ImportReceiver.kt
```diff
@@ -0,0 +1,89 @@
...
+    private val uuid: UUID,
+    private val importHandler: ImportHandler,
+    ) : BroadcastReceiver() {
+    override fun onReceive(context: Context?, intent: Intent?) {
+        val id = intent?.getStringExtra(EXTRA_UUID)
+        val parsedId: UUID? = try {
+            UUID.fromString(id)
+        } catch (e: Exception) {
+            null
+        }
```
Bnyro:
```suggestion
        val parsedId runCatching {
            UUID.fromString(id)
        }.getOrNull()
```

app/src/main/java/com/github/libretube/workers/ImportCoroutineWorker.kt
```diff
@@ -0,0 +1,176 @@
...
+    appContext: Context,
+    workerParams: WorkerParameters,
+) : CoroutineWorker(appContext, workerParams) {
+
+    private val notificationFactory: NotificationProvider = NotificationHandler(id,appContext);
+    private val notificationManager =
+        appContext.getSystemService(Context.NOTIFICATION_SERVICE) as NotificationManager
+    private val IMPORT_THUMBNAIL_QUALITY = "mqdefault"
+    private val VIDEO_ID_LENGTH = 11
+    private val YOUTUBE_IMG_URL = "https://img.youtube.com"
```
Bnyro: Please make all these values `const` and move them to the companion object.

app/src/main/java/com/github/libretube/workers/ImportCoroutineWorker.kt
```diff
@@ -0,0 +1,176 @@
...
+                                )
+                            }
+                    } ?: emptyList()
+                }
+
+                else -> emptyList()
+            }
+        }?.flatten().orEmpty()
+
+        registerReceiver()
```
Bnyro: Why register the receiver again here? It's already registered, right?

app/src/main/java/com/github/libretube/workers/ImportCoroutineWorker.kt
```diff
@@ -0,0 +1,176 @@
...
+
+
+    private suspend fun ImportWatchHistory(): Unit {
+        registerReceiver()
+
+        val fileUrisString = inputData.getStringArray(WorkersData.FILES)
+        val importFormatString =
+            inputData.getString(WorkersData.IMPORT_FORMAT) ?: ImportFormat.YOUTUBEJSON.toString()
+        val importFormat = enumValueOf<ImportFormat>(importFormatString)
+        val videos = fileUrisString?.map { a ->
```
Bnyro:
```suggestion
        val videos = fileUrisString?.flatMap { a ->
```
and remove `flatten` at the end.

app/src/main/java/com/github/libretube/workers/ImportCoroutineWorker.kt
```diff
@@ -0,0 +1,176 @@
...
+            val fileUrisDeserialized = Uri.parse(a)
```
Bnyro:
```suggestion
            val fileUri = a.toUri()
```

app/src/main/java/com/github/libretube/workers/ImportCoroutineWorker.kt
Bnyro: Please use a more descriptive name instead of `a`.

app/src/main/java/com/github/libretube/db/obj/StoredItemState.kt
Bnyro, line 1: remove: unused

app/src/main/java/com/github/libretube/workers/ImportCoroutineWorker.kt
```diff
@@ -0,0 +1,176 @@
...
+    private val notificationManager =
+        appContext.getSystemService(Context.NOTIFICATION_SERVICE) as NotificationManager
+    private val IMPORT_THUMBNAIL_QUALITY = "mqdefault"
+    private val VIDEO_ID_LENGTH = 11
+    private val YOUTUBE_IMG_URL = "https://img.youtube.com"
+    override suspend fun doWork(): Result {
+        setForeground(getForegroundInfo())
+        val pausingHandle = ImportHandler()
+        val importType = inputData.getString(WorkersData.IMPORT_TYPE)
+            ?: ImportType.IMPORT_WATCH_HISTORY.toString()
```
Bnyro: If no import type is provided, we should probably throw an error instead of just assuming that it's `IMPORT_WATCH_HISTORY`.

app/src/main/java/com/github/libretube/workers/ImportCoroutineWorker.kt
```diff
@@ -0,0 +1,176 @@
...
+                }
+            }
+
+            else -> Unit
+        }
+        return Result.success()
+    }
+
+
+    private suspend fun ImportWatchHistory(): Unit {
```
Bnyro:
```suggestion
    private suspend fun importWatchHistory(): Unit {
```
method names should be lowercase

app/src/main/java/com/github/libretube/workers/ImportCoroutineWorker.kt
```diff
@@ -0,0 +1,176 @@
...
+        } else if (!importHandler.isPaused) {
+            dispatch(true)
+            if (now - lastUpdateTime >= updateInterval) {
+                notifyNotification(notificationFactory.updateState(currentState, finalState, ImportState.RUNNING))
+                lastUpdateTime = now
+            }
+        }
+    }
+
+    private fun notifyNotification(notification: Notification) {
```
Bnyro:
```suggestion
    private fun updateNotification(notification: Notification) {
```

app/src/main/java/com/github/libretube/workers/ImportCoroutineWorker.kt
```diff
@@ -0,0 +1,176 @@
...
+        ContextCompat.registerReceiver(
+            applicationContext,
+            importReceiver,
+            ImportReceiver.createIntentFilter(),
+            ContextCompat.RECEIVER_NOT_EXPORTED
+        )
+    }
+
+    companion object {
+        fun importWatchHistory(context: Context,uris: List<Uri>, importFormat: ImportFormat) {
```
Bnyro:
```suggestion
        fun startImportWatchHistoryWorker(context: Context,uris: List<Uri>, importFormat: ImportFormat) {
```

app/src/main/java/com/github/libretube/workers/ImportCoroutineWorker.kt
```diff
@@ -0,0 +1,176 @@
...
+            notifyNotification(notificationFactory.updateState(currentState, finalState, ImportState.PAUSED))
+            importHandler.awaitResumed()
+        } else if (!importHandler.isPaused) {
+            dispatch(true)
+            if (now - lastUpdateTime >= updateInterval) {
+                notifyNotification(notificationFactory.updateState(currentState, finalState, ImportState.RUNNING))
+                lastUpdateTime = now
+            }
+        }
+    }
```
Bnyro: I don't really understand what this method is doing, in any case its name is not very descriptive.

As I understand it, the method is supposed to only execute the work from `dispatch` if `ImportHandler#paused` is `false`. However, in this case, why don't you just do
```suggestion
    private suspend fun runActionAndUpdateNotificationIfNotPaused(
        currentState: Int,
        finalState: Int,
        action: suspend () -> Unit
    ) {
        val importHandler = ImportHandler.current()
        val now = System.currentTimeMillis()
        if (importHandler.isPaused) {
            notifyNotification(notificationFactory.updateState(currentState, finalState, ImportState.PAUSED))
            importHandler.awaitResumed()
        } else if (!importHandler.isPaused) {
            action()
            if (now - lastUpdateTime >= updateInterval) {
                notifyNotification(notificationFactory.updateState(currentState, finalState, ImportState.RUNNING))
                lastUpdateTime = now
            }
        }
    }
```

app/src/main/java/com/github/libretube/workers/ImportCoroutineWorker.kt
```diff
@@ -0,0 +1,176 @@
...
+            checkIfPausedOrCancelled(index, videos.size) {
+                if (it) {
+                    DatabaseHelper.addToWatchHistory(videos.get(index))
+                    index++
+                }
+            }
+        }
+
+        if (videos.isEmpty()) {
+            applicationContext.toastFromMainDispatcher(R.string.emptyList)
```
Bnyro: Please do this check before you execute the `while` loop for readability.

app/src/main/java/com/github/libretube/workers/ImportCoroutineWorker.kt
```diff
@@ -0,0 +1,176 @@
...
+        } else {
+            applicationContext.toastFromMainDispatcher(R.string.success)
+        }
+    }
+
+
+    var lastUpdateTime = 0L
+    val updateInterval = 500L
```
Bnyro: Please make update interval a constant in the companion object.

app/src/main/java/com/github/libretube/ui/preferences/BackupRestoreSettings.kt
```diff
@@ -33,6 +33,9 @@ class BackupRestoreSettings : BasePreferenceFragment() {
     private var backupFile = BackupFile()
     private var importFormat: ImportFormat = ImportFormat.NEWPIPE
 
+
+    override val titleResourceId: Int = R.string.backup_restore
```
Bnyro: Please undo this change, perhaps this was by accident?

Issue #7966, opened 2025-11-27
Player in full screen doesn't recognize gestures


deckstose:
### Steps to reproduce

1. Open full screen player with swipe-up gesture (this happens in ~1/3 cases I would say)
2. Leave full screen and repeat

Sadly this is not really reliably reproducible.

### Expected behavior

Player is able to show it's UI with a single tap, double tap pauses video, etc...

### Actual behavior

The player is in full screen and plays the video, but gestures and interaction won't work. The only way to get back to portrait mode is to use the system back gesture.

### LibreTube version

v0.29.1+c558255 (nightly build)

### Android version

Android 16

### Other details

I update nightly builds regularly and I think this is a regression from https://github.com/libre-tube/LibreTube/pull/7913, although I've not bisected the issue. Stable v0.29.1 works just fine.

What I also observe is that _when_ you manage to exit full screen a decent amount of times the portrait player is not snapped to the top edge of the device but floats around in the middle of the screen. You then need to snap the player to the top edge before any interaction works in portrait mode as well (I include this since this maybe related, is it possible that the player in full screen (landscape) _thinks_ it's also not snapped to any edge, although it is, and thus doesn't allow any interaction until you fix it in portrait mode?). The other times _I think_ the player is still not snapped but it's barely visible. You sometimes still need to swipe the portrait player when it looks like it's attached to the top edge before you can interact with it.

This is a lot of guessing from my side, but I hope this helps to interpret this correctly.

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [x] I have written a short but informative title.
- [x] I will fill out all of the requested information in this form.

dhanuarf:
Sorry probably I should've clarified in #7913 that after this change, you have to swipe up at the center area of the video to enter fullscreen, and since you also need to swipe down at the center area to exit fullscreen so I think this change makes sense just fine, although it needs some proper clarification to inform this behavior change in the next version update.
Maybe I can find a way to implement the original swipe up behavior back later.

Other than this, I cant reproduce the part where the touch doesnt work in fullscreen, it works consistently on my device.

shantanu1k:
> Maybe I can find a way to implement the original swipe up behavior back later.

Till that time, is it possible to add an FTUX?
So, that users will know to swipe up and down from the centre?

Bnyro:
[...]

PR #7974, opened 2025-11-30
feat(video player): add visual feedback when fullscreen gesture is in progress

dhanuarf:
Add scale animation to the player view when entering/exiting player via gesture swipe.

This PR adds a new package called `controllers` in `com.github.libretube.ui` to put the `FullscreenGestureAnimationController` class that handle the scale animation.

ref: https://github.com/libre-tube/LibreTube/issues/7966#issuecomment-3591035665

https://github.com/user-attachments/assets/c06e7cbe-56d8-4708-a8a5-d8e5ddcec0da

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/listeners/PlayerGestureController.kt
+++ b/app/src/main/java/com/github/libretube/ui/listeners/PlayerGestureController.kt
@@ -192,6 +192,12 @@ class PlayerGestureController(activity: BaseActivity, private val listener: Play
 
             isMoving = true
 
+            if (!isFullscreen && distanceY > 0) {
+                // Allow swipe up on the entire area if the player is not currently in fullscreen
+                listener.onSwipeCenterScreen(distanceY, e2.y)
+                return true
+            }
+
             when {
                 e1.x < width * LEFT_AREA_VIEW_PERCENTAGE -> {
                     if (isFullscreen) listener.onSwipeLeftScreen(distanceY, e2.y)


app/src/main/java/com/github/libretube/ui/controllers/FullscreenGestureAnimationController.kt
```diff
@@ -0,0 +1,96 @@
...
+    private val onSwipeUpCompleted: () -> Unit,
+    private val onSwipeDownCompleted: () -> Unit,
+) {
+    private var isSwipeInProgress = false
+    private var shouldHandleSwipe = false
+    /**
+     * - `value > 0` = Swiping up
+     * - `value < 0` = Swiping down
+     */
+    private var swipeDirection = 0
```
Bnyro: Since you're not using the actual value of `swipeDirection`, this could also be an enum type with the values `up, none, down` for better readability

app/src/main/java/com/github/libretube/ui/controllers/FullscreenGestureAnimationController.kt
```diff
@@ -0,0 +1,96 @@
...
+    private var swipeStartY = 0f
+
+    fun onSwipe(distanceY: Float, e2: MotionEvent) {
```
Bnyro:
```suggestion
    fun onSwipe(distanceY: Float, positionY: Float) {
```
Since we only use `y` from `e2`, we should use the information hiding principle and only pass the position `e2.y` here instead of the full event

app/src/main/java/com/github/libretube/ui/controllers/FullscreenGestureAnimationController.kt
```diff
@@ -0,0 +1,96 @@
...
+        if (!isSwipeInProgress) {
+            isSwipeInProgress = true
+            swipeDirection = distanceY.toInt()
+            shouldHandleSwipe =
+                (!playerView.isFullscreen() && swipeDirection > 0) ||
+                        (playerView.isFullscreen() && swipeDirection < 0)
```
Bnyro:
```suggestion
            shouldHandleSwipe = if (player.isFullscreen()) swipeDirection == SwipeDirection.Up else swipeDirection == SwipeDirection.Down
```

app/src/main/java/com/github/libretube/ui/controllers/FullscreenGestureAnimationController.kt
```diff
@@ -0,0 +1,96 @@
...
+            .start()
+
+        isSwipeInProgress = false
+        swipeDirection = 0
+    }
+
+    companion object {
+        private const val SCALE_DIFF_TARGET = 0.12f
+        private val SWIPE_DISTANCE_THRESHOLD = 50f.dpToPx()
+        private val SCALE_FACTOR = SCALE_DIFF_TARGET / SWIPE_DISTANCE_THRESHOLD
```
Bnyro: Please also add a short comment about what `SCALE_DIFF_TARGET` does, i.e. I don't get why `SCALE_FACTOR` is calculated this way

Bnyro (commented):
Looks really cool, thanks!

I've left some small comments concerning code quality for simplifying and clarifying some places

app/src/main/java/com/github/libretube/ui/controllers/FullscreenGestureAnimationController.kt
dhanuarf: It basically defines the maximum difference of the scale, so when scalling up it caps the scale at `1f + SCALE_DIFF_TARGET` and when scalling down it caps at `1f - SCALE_DIFF_TARGET`. 
I just coulnt come up with a better name lol.
Any idea what should I name it?

Bnyro: `MAXIMUM_SCALE_DIFF_PERCENTAGE` maybe - and please  also add a documentary comment to it with what you just said :)

dhanuarf:
[4c46319](https://github.com/libre-tube/LibreTube/pull/7974/commits/4c4631965c6b6177ec9b4f7f8d493b3078b18697): The suggested changes. Also added a debouncer to make sure `onSwipeCompleted()` only called once.

Added a new commit [e03cbd6](https://github.com/libre-tube/LibreTube/pull/7974/commits/e03cbd6860a40a74f85398ceee2bc5d516b21f09) to allow swipe up not just from the center.

PR #7975, opened 2025-12-01
fix: Fix keyboard hotkeys not working

jcsepulvedar:
This commit fixes tablets with attached keyboard not registering input

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
+++ b/app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
@@ -1566,7 +1566,7 @@ class PlayerFragment : Fragment(R.layout.fragment_player), OnlinePlayerOptions {
 
             viewModel.isOrientationChangeInProgress = true
 
-            // detatch player view from player to stop surface rendering
+            // detach player view from player to stop surface rendering
             binding.player.player = null
 
             if (::playerController.isInitialized) playerController.release()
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
```diff
@@ -437,6 +437,15 @@ class PlayerFragment : Fragment(R.layout.fragment_player), OnlinePlayerOptions {
...
         super.onViewCreated(view, savedInstanceState)
 
+        binding.player.apply {
+            isFocusable = true
+            isFocusableInTouchMode = true
+            requestFocus()
+            setOnKeyListener { _, keyCode, _ ->
+                onKeyUp(keyCode)
+            }
+        }
```
Bnyro: In any case, this should be moved to `CustomExoPlayerView.kt`

jcsepulvedar: I'll look into it :+1:

jcsepulvedar:
> For me, keyboard events already work fine, so I don't see why these changes are needed?

On my and my wifes tablet (Lenovo  Pro Tab 11 and Lenovo Pro Tab 12 respectivaly) is a hit or miss. Sometimes it pauses the video on the first try, then the focus automatically changes to a random element on screen without me tapping anywhere. So if I press spacebar again to resume the video most of the time it exits fullscreen and or plays the next video. 
With my changes the focus does not go away. Have not had the time to follow up on your last comment sadly.

Bnyro:
> On my and my wifes tablet (Lenovo Pro Tab 11 and Lenovo Pro Tab 12 respectivaly) is a hit or miss. Sometimes it pauses the video on the first try, then the focus automatically changes to a random element on screen without me tapping anywhere. So if I press spacebar again to resume the video most of the time it exits fullscreen and or plays the next video.
> With my changes the focus does not go away
>
Thanks for the explanation! (I'm not really using LibreTube with a keyboard, so I can't tell how it performs in the real world except from a short test)

In that case it definitely makes sense to include these changes. It'd be great if you could move the changes you made to the `CustomExoPlayerView` and test it again (no time pressure), then it's good to merge from my side.

Issue #7972, opened 2025-11-29
Error when importing watch history


DavinSimpson:
### Steps to reproduce

1. Open Settings
2. Click "Backup and restore"
3. Select "Import watch history"
4. Import watch-history.json

### Expected behavior

Watch history should be imported into app

### Actual behavior

App crashes and restarts then goes to homepage while throwing an error.

### LibreTube version

0.29.1

### Android version

Android 13

### Other details

Attached is screenshot of error and a copy of the log, and watch history JSON file

<img width="720" height="1451" alt="Image" src="https://github.com/user-attachments/assets/227c00de-e36d-480f-a0e1-85c7fa0ef619" />
[watch-history.json](https://github.com/user-attachments/files/23834698/watch-history.json)

**LOG:**
kotlinx.serialization.MissingFieldException: Field 'url' is required for type with serial name 'com.github.libretube.obj.YouTubeWatchHistoryChannelInfo', but it was missing at path: $[38670].subtitles[0]
	at kotlinx.serialization.json.internal.StreamingJsonDecoder.decodeSerializableValue$1(Unknown Source:160)
	at kotlin.io.CloseableKt.decodeSerializableElement(Unknown Source:10)
	at kotlinx.serialization.json.internal.StreamingJsonDecoder.decodeSerializableElement(Unknown Source:49)
	at kotlinx.serialization.internal.CollectionLikeSerializer.readElement(Unknown Source:7)
	at kotlinx.serialization.internal.AbstractCollectionSerializer.merge(Unknown Source:29)
	at kotlinx.serialization.internal.AbstractCollectionSerializer.deserialize(Unknown Source:0)
	at kotlinx.serialization.json.internal.StreamingJsonDecoder.decodeSerializableValue$1(Unknown Source:105)
	at kotlin.io.CloseableKt.decodeSerializableElement(Unknown Source:10)
	at kotlinx.serialization.json.internal.StreamingJsonDecoder.decodeSerializableElement(Unknown Source:49)
	at com.github.libretube.obj.YouTubeWatchHistoryFileItem$$serializer.deserialize(SourceFile:1)
	at com.github.libretube.obj.YouTubeWatchHistoryFileItem$$serializer.deserialize(SourceFile:2)
	at kotlinx.serialization.json.internal.StreamingJsonDecoder.decodeSerializableValue$1(Unknown Source:105)
	at kotlin.io.CloseableKt.decodeSerializableElement(Unknown Source:10)
[...]

PR #7976, opened 2025-12-01
(BUG): Made name and url optional for deserilization

SE-Bilal-Ahmad:
Url was required at the time of deserilization i have made it optional 

Bug #7972

### DIFF
--- a/app/src/main/java/com/github/libretube/obj/YouTubeWatchHistoryChannelInfo.kt
+++ b/app/src/main/java/com/github/libretube/obj/YouTubeWatchHistoryChannelInfo.kt
@@ -4,6 +4,6 @@ import kotlinx.serialization.Serializable
 
 @Serializable
 data class YouTubeWatchHistoryChannelInfo(
-    val name: String? = null,
+    val name: String,
     val url: String? = null
 )


app/src/main/java/com/github/libretube/obj/YouTubeWatchHistoryChannelInfo.kt
```diff
@@ -4,6 +4,6 @@ import kotlinx.serialization.Serializable
 
 @Serializable
 data class YouTubeWatchHistoryChannelInfo(
-    val name: String,
-    val url: String
+    val name: String? = null,
```
FineFindus:
```suggestion
    val name: String,
```

We should keep the name as non-null, as it still seems to be required.

SE-Bilal-Ahmad:
@FineFindus i will look into it

app/src/main/java/com/github/libretube/obj/YouTubeWatchHistoryChannelInfo.kt
SE-Bilal-Ahmad: Alright i will make the name as required i will just make the url optional

SE-Bilal-Ahmad: Name made required now

Issue #7968, opened 2025-11-29
Add additional sorting options to Downloads Tab


told-ishan:
### Describe your suggested feature

Description:

I would like to request the addition of sorting features in the Downloads and Playlists tabs.

Requested Features:

1. Sorting Options
Sort by date added (newest → oldest / oldest → newest)
Sort alphabetically (A–Z / Z–A)
Sort by duration
Sort by channel name
Sort by file size (for downloads, if available)

2. Custom Sorting
Ability to manually reorder items using drag-and-drop Slider.

Reasoning:
These features would make it easier to organize large playlists number and downloaded videos, improve navigation, and give users more control over how their content is arranged.

Btw really grateful for the work you guys have done.

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [x] I have written a short but informative title.
- [x] I will fill out all of the requested information in this form.

FineFindus:
Thanks for the suggestion, both of these tabs already support various sorting options.

told-ishan:
I was referring to custom sorting. When I download lecture videos, they sometimes get shuffled, so having an option for custom sorting would be really helpful. I understand it might be lot to implement right now, but I hope it can be considered for the future.

dhanuarf:
I'm not sure why this is closed... 
This is actually a good feature request though, because currently the we have only 'newest-oldest' sorting rule in download tab.

SE-Bilal-Ahmad:
I will look into it

Bnyro:
Custom sorting would add too much code that must be maintained in the future, so it's not worth being added.

The other sorting options could definitely make sense, so please feel free to look into that @Bilal-Ahmad123 .

PR #7986, opened 2025-12-08
Enhancement: Added Four new sorting options for downloads tab 

SE-Bilal-Ahmad:
1- Added Alphabeticall sort on the basis of title
2- Added sorting w.r.t channel name
3- Added video sorting w.r.t video size 
4- Added sorting w.r.t video length 

Bug #7968

### DIFF
--- a/app/src/main/java/com/github/libretube/enums/SortEnum.kt
+++ b/app/src/main/java/com/github/libretube/enums/SortEnum.kt
@@ -1,9 +1,0 @@
-package com.github.libretube.enums
-
-enum class SortEnum() {
-    OLDEST,
-    NEWEST,
-    ALPHABETICALLY,
-    DURATION,
-    CHANNEL
-}
--- a/app/src/main/res/values/array.xml
+++ b/app/src/main/res/values/array.xml
@@ -173,15 +173,6 @@
         <item>@string/alphabetic_reversed</item>
     </string-array>
 
-    <string-array name="downloadSortOptions">
-        <item>@string/least_recent</item>
-        <item>@string/most_recent</item>
-        <item>@string/sort_alphabetically</item>
-        <item>@string/sort_duration</item>
-        <item>@string/sort_channel</item>
-        <item>@string/sort_size</item>
-    </string-array>
-
     <string-array name="filterOptions">
         <item>@string/all</item>
         <item>@string/videos</item>
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -249,9 +249,7 @@
     <string name="most_recent">Newest</string>
     <string name="least_recent">Oldest</string>
     <string name="most_views">Most views</string>
-    <string name="sort_alphabetically">Alphabetically</string>
-    <string name="sort_duration">By duration</string>
-    <string name="sort_channel">Channel</string>
+    <string name="sort_channel">Channel name</string>
     <string name="sort_size">Size</string>
     <string name="least_views">Least views</string>
     <string name="required_network">Required connection</string>
--- a/app/src/main/java/com/github/libretube/ui/fragments/DownloadsFragment.kt
+++ b/app/src/main/java/com/github/libretube/ui/fragments/DownloadsFragment.kt
@@ -11,6 +11,7 @@ import android.os.IBinder
 import android.view.View
 import android.view.ViewGroup.MarginLayoutParams
 import androidx.activity.OnBackPressedCallback
+import androidx.annotation.StringRes
 import androidx.core.content.ContextCompat
 import androidx.core.os.bundleOf
 import androidx.core.view.isGone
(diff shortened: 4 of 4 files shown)


app/src/main/java/com/github/libretube/enums/SortEnum.kt
```diff
@@ -0,0 +1,9 @@
+package com.github.libretube.enums
+
+enum class SortEnum() {
```
Bnyro:
```suggestion
enum class SortEnum {
```
nit

app/src/main/res/values/strings.xml
```diff
@@ -249,6 +249,10 @@
     <string name="most_recent">Newest</string>
     <string name="least_recent">Oldest</string>
     <string name="most_views">Most views</string>
+    <string name="sort_alphabetically">Alphabetically</string>
+    <string name="sort_duration">By duration</string>
+    <string name="sort_channel">Channel</string>
+    <string name="sort_size">Size</string>
```
Bnyro: I'm pretty sure we already have strings for these sort options, e.g. https://github.com/libre-tube/LibreTube/blob/master/app/src/main/res/values/strings.xml#L415.

app/src/main/res/values/strings.xml
```diff
@@ -249,6 +249,10 @@
     <string name="most_recent">Newest</string>
     <string name="least_recent">Oldest</string>
     <string name="most_views">Most views</string>
+    <string name="sort_alphabetically">Alphabetically</string>
+    <string name="sort_duration">By duration</string>
+    <string name="sort_channel">Channel</string>
```
Bnyro: This should be called `Channel name`

app/src/main/res/values/strings.xml
```diff
@@ -249,6 +249,10 @@
     <string name="most_recent">Newest</string>
     <string name="least_recent">Oldest</string>
     <string name="most_views">Most views</string>
+    <string name="sort_alphabetically">Alphabetically</string>
+    <string name="sort_duration">By duration</string>
```
Bnyro: There shouldn't be any `By ` here, it should only be `Duration` for consistency

app/src/main/res/values/array.xml
```diff
@@ -176,6 +176,10 @@
     <string-array name="downloadSortOptions">
         <item>@string/least_recent</item>
         <item>@string/most_recent</item>
+        <item>@string/sort_alphabetically</item>
+        <item>@string/sort_duration</item>
+        <item>@string/sort_channel</item>
+        <item>@string/sort_size</item>
```
Bnyro: I think it would be better to delete this array here and move the string ids to `SortEnum`, i.e.


```kotlin
enum class SortEnum(@StringRes val stringId) {
    OLDEST(R.string.least_recent)
    ...
}
```

app/src/main/java/com/github/libretube/enums/SortEnum.kt
Bnyro, line 1: Please put this into `DownloadsFragment.kt` and rename it to `DownloadSortingOrder` (private class):
1. `SortEnum` is not a descriptive name
2. Only `DownloadsFragment` needs access to it, so it should be private to `DownloadsFragment` and not public

app/src/main/res/values/array.xml
SE-Bilal-Ahmad: Just saw the code right now the thing is that the array is being used at two places in the DownloadFragmentPage and PlayListDownloadFragment both of these fragments exist in DownloadFragment.kt if i remove the array than i will have to declare the private enum class in both of these fragments too

Bnyro: No, just declare it in that file, not within any of these classes :)

SE-Bilal-Ahmad: Okay my bad

app/src/main/java/com/github/libretube/enums/SortEnum.kt
SE-Bilal-Ahmad: SortEnum class now removed

app/src/main/res/values/strings.xml
SE-Bilal-Ahmad: There was one for alphabet and duration but none for size and Channel name

app/src/main/res/values/strings.xml
SE-Bilal-Ahmad: Now renamed to Channel name

app/src/main/res/values/strings.xml
SE-Bilal-Ahmad: By duration removed now using the existing duration string

app/src/main/res/values/array.xml
SE-Bilal-Ahmad: Now has been moved to Enum array removed

app/src/main/java/com/github/libretube/enums/SortEnum.kt
SE-Bilal-Ahmad: Has now been placed at DownloadsFragment.kt file with name DownloadSortingOrder

PR #7991, opened 2025-12-10
fix(video player): glitch when exiting from fullscreen on short videos

dhanuarf:
Fix a glitch when trying to exit from fullscreen mode by using the swipe down gesture on short (vertical) videos.

The scale animation is now applied to the video frame view instead to the `PlayerView` itself.

The glitch:

https://github.com/user-attachments/assets/46ac2a76-80bd-49e4-9b03-9ea928955aee

---
After some debugging I found that touch event position is jumping up and down, and I guess that the touch position is shifting along the scale that is being applied. Which is weird, because it works fine on horizontal videos. And touch event position should retain its original position even when a view is being scaled.

Dunno...

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/controllers/FullscreenGestureAnimationController.kt
+++ b/app/src/main/java/com/github/libretube/ui/controllers/FullscreenGestureAnimationController.kt
@@ -1,5 +1,6 @@
 package com.github.libretube.ui.controllers
 
+import android.view.View
 import android.view.animation.AccelerateDecelerateInterpolator
 import com.github.libretube.extensions.dpToPx
 import com.github.libretube.ui.views.CustomExoPlayerView
@@ -10,6 +11,7 @@ import com.github.libretube.ui.views.CustomExoPlayerView
  */
 class FullscreenGestureAnimationController(
     private val playerView: CustomExoPlayerView,
+    private val videoFrameView: View,
     private val onSwipeUpCompleted: () -> Unit,
     private val onSwipeDownCompleted: () -> Unit,
 ) {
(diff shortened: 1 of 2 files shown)


FineFindus (changes requested):
Thanks for the fix, however, this introduces another bug, where the user can get stuck in the player view by quickly swiping down and up.

dhanuarf:
> Thanks for the fix, however, this introduces another bug, where the user can get stuck in the player view by quickly swiping down and up.

Cant reproduce on my side. What type of 'stuck' does it produce? Does it freeze the app or the video frame stays on scaled size or other issue?

FineFindus:
The video stays in its original positions and does not respond to touch gestures. When exiting via the backgestures, it's already in the miniplayer position.

dhanuarf:
> The video stays in its original positions and does not respond to touch gestures. When exiting via the backgestures, it's already in the miniplayer position.

Actually, I do experience  something like that sometimes even before this scale animation patch, but quite rarely. And I think the issue is similar to this one: #7966.

If you look closely in the video attachment of that issue, at the end when exiting from fullscreen, the player position is slightly off (not at the top-dead position). And in `PlayerFragment` we have a code where we disable controls when `MotionLayout's progress > 0` , that's why the controls and gestures wont works in this state. 

So somehow the `MotionLayout` is receiving touch when the player in fullscreen. Maybe we can try to prevent it to receive touch event when we are in fullscreen and see what happens? I rarely experience this issue so I can't test this myself...

dhanuarf:
> So somehow the `MotionLayout` is receiving touch when the player in fullscreen. Maybe we can try to prevent it to receive touch event when we are in fullscreen and see what happens?

Or maybe just force the motion progress to 0f when we're entering fullscreen.
Or detect if we're in fullscreen and then just ignore this controls disabling logic regardless the progress.

FineFindus:
Hmm, must have never noticed that in the current implementation

> So somehow the MotionLayout is receiving touch when the player in fullscreen.

It might be that it isn't properly set to fullscreen. I remember there being a rare issue, where the player wouldn't enter fullscreen properly.

Issue #7987, opened 2025-12-08
Crash when switching video source


Novantric:
### Steps to reproduce

1. Start playing a local, downloaded video
2. Play another video that isn't downloaded. For reliable results, keep the local stream playing.

### Expected behavior

LibreTube should seamlessly switch between local and offline modes.

### Actual behavior

The app will crash very consistently if the steps above are performed, and only occasionally if done in reverse (online to offline).

### LibreTube version

0.29.1 (64)

### Android version

16

### Other details

Crash example 1:
```
java.lang.NullPointerException
	at com.github.libretube.services.OfflinePlayerService.startPlayback$suspendImpl(Unknown Source:100)
	at com.github.libretube.services.OfflinePlayerService$startPlayback$1.invokeSuspend(Unknown Source:11)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(Unknown Source:8)
	at kotlinx.coroutines.UndispatchedCoroutine.afterResume(Unknown Source:53)
	at kotlinx.coroutines.AbstractCoroutine.resumeWith(Unknown Source:22)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(Unknown Source:31)
	at kotlinx.coroutines.DispatchedTask.run(Unknown Source:106)
	at androidx.work.impl.utils.SerialExecutorImpl$Task.run(Unknown Source:13)
	at kotlinx.coroutines.scheduling.TaskImpl.run(Unknown Source:2)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(Unknown Source:92)
	Suppressed: kotlinx.coroutines.internal.DiagnosticCoroutineContextException: [StandaloneCoroutine{Cancelling}@9d79c34, Dispatchers.IO]
```

Crash example 2:
```
java.lang.NullPointerException
	at com.github.libretube.services.OfflinePlayerService.startPlayback$suspendImpl(Unknown Source:100)
	at com.github.libretube.services.OfflinePlayerService$startPlayback$1.invokeSuspend(Unknown Source:11)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(Unknown Source:8)
	at kotlinx.coroutines.UndispatchedCoroutine.afterResume(Unknown Source:53)
	at kotlinx.coroutines.AbstractCoroutine.resumeWith(Unknown Source:22)
[...]

PR #7996, opened 2025-12-12
fix: Crash when starting online stream while offline stream in progress

shantanu1k:
Fixes the crash that occurs when a new online video is played, while the offline video is playing.

Closes: https://github.com/libre-tube/LibreTube/issues/7987

Screen recording:

https://github.com/user-attachments/assets/cf387b52-4d14-4de1-b2a7-89cdba0a054d


Novantric:
Thanks! This is something I encountered almost daily!

Issue #6455, opened 2024-09-06
F-Droid can't build


licaon-kter:
### Describe your suggested feature

ref: https://gitlab.com/fdroid/fdroiddata/-/jobs/7760790503#L379

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

/LE: fyi https://gitlab.com/fdroid/fdroiddata/-/commit/f3e76bdabb388c9a7a197388af5222bb01d8a91c

Bnyro:
According to https://github.com/TeamNewPipe/NewPipeExtractor/issues/1066, the dependency might need to be lowercased to `...teamnewpipe:NewPipeExtractor...` instead of `...TeamNewpipe:NewPipeExtractor...`, although it worked fine for me the other way yesterday, probably due to dependency caching.

In CI (gh actions) of this repository, it sometimes passes and sometimes fails with the error you mentioned, I love deterministic build failures...

Bnyro:
Can you please try building from the latest commit (https://github.com/libre-tube/LibreTube/commit/34228f8493e3fec93f2ac24cf760c40ae2f3d862)?

This works for me now after clearing the IDEA caches.

licaon-kter:
@Bnyro 1d4304ce13a12be6ccc0200df6b281fd93ae8354 builds fine locally

licaon-kter:
hmmm https://gitlab.com/fdroid/checkupdates-bot-fdroiddata/-/jobs/8396839770#L229

licaon-kter:
New scanner code picked it up.

Just `exclude` does not make that dep FOSS, when it was build from non-FOSS code.

IzzySoft:
@Bnyro for what it's worth, I try to build the versions since 1.25.1 and fail with the same error. With v0.26.0 it is

```
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':app:mergeReleaseNativeLibs'.
> Could not resolve all files for configuration ':app:releaseRuntimeClasspath'.
   > Could not find com.github.teamnewpipe:NewPipeExtractor:6e3a4a6d9de61eafb73e7eb1b714847b5077856d.
     Searched in the following locations:
[...]

PR #8002, opened 2025-12-13
build: drop support for split-abi apks

FineFindus:
As we do not ship any native modules (and haven't done so since we removed cronet in https://github.com/libre-tube/LibreTube/commit/93275b9) the additional split [ABI apks are identical to the universal apk](
https://github.com/libre-tube/NightlyBuilds/blob/main/checksums). Dropping them simplifies our build process, avoids uploading  and makes it easier for users to download the 'right' apk.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -86,16 +86,6 @@ android {
         jvmTarget = "17"
     }
 
-    // Comment this block if issues occur while generating the baseline profile
-    splits {
-        abi {
-            isEnable = true
-            reset()
-            include("armeabi-v7a", "arm64-v8a", "x86", "x86_64")
-            isUniversalApk = true
-        }
-    }
-
     packaging {
         jniLibs.excludes.add("lib/armeabi-v7a/*_neon.so")
     }


Bnyro (commented):
I initially kept them because we otherwise need to
- tell F-Droid and IzzyOnDroid to change their build workflow.
- tell users that use Obtainium to change their configuration.

If you could do that, it'd be great :)

FineFindus:
> tell F-Droid and IzzyOnDroid to change their build workflow.

That's not necessary, neither one distributes ABI-specific binaries (i.e. they have no ABI tags on their respective websites, nor the downloads for them).

> tell users that use Obtainium to change their configuration.

I think it should be enough to write this into the next release announcement? I assume that if someone used Obtainium for an ABI-specific APK, it would notify them about failing to download that?

Bnyro:
> That's not necessary, neither one distributes ABI-specific binaries (i.e. they have no ABI tags on their respective websites, nor the downloads for them).
> 
No, IzzyOnDroid specifically pulls our GitHub releases armv8 apk as far as I know. F-Droid is likely fine without changes because our builds are not reproducible.
> 
> I think it should be enough to write this into the next release announcement? I assume that if someone used Obtainium for an ABI-specific APK, it would notify them about failing to download that?
>
Should be fine, yeah. If we don't forget doing so.

FineFindus:
> No, IzzyOnDroid specifically pulls our GitHub releases armv8 apk as far as I know. F-Droid is likely fine without changes because our builds are not reproducible.

Create an issue: https://gitlab.com/IzzyOnDroid/repo/-/issues/918

> Should be fine, yeah. If we don't forget doing so.

Luckily I'm not the one doing the release, so … :)

IzzySoft:
> No, IzzyOnDroid specifically pulls our GitHub releases armv8 apk as far as I know.

We pull the universal release – both, for our repo as well as for our RB builders. So if you keep the file name `app-universal-release.apk`, no changes would be needed at our end (except for the optional change stripping the other ones from our copy of your `build.gradle` after cloning the repo and before building the APK). I can put a note in the recipe file to do that with the next release (or wait if it fails, and do it then :see_no_evil:).

Bnyro:
> We pull the universal release – both, for our repo as well as for our RB builders. So if you keep the file name `app-universal-release.apk`, no changes would be needed at our end (except for the optional change stripping the other ones from our copy of your `build.gradle` after cloning the repo and before building the APK). I can put a note in the recipe file to do that with the next release (or wait if it fails, and do it then 🙈).
>
Thanks for the prompt response! <3

Starting with the next release, builds will be named `app-release.apk` because that's what Gradle names them by default if you don't build split ABI APKs. 

So it'd be great if you could just leave a reminder or something (or adapt it now) to pull `app-release.apk` in the future.

IzzySoft:
OK, then let me just leave all "as-is" for now. Our RB builder would yell about "no matching APKs at the release" – and then I can check to clean up the entire recipe (which can only be done when also running it, so we have to postpone that).

Just wanted to add a note to the recipe – and see there already is one, pointing to #6455 – I'll just add this one to it then for reference … done.

PR #8004, opened 2025-12-14
github/SECURITY.md

anindyakhansa-creator:
There is no history in Chrome and it is safe with strict security

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,21 @@
+# Security Policy
+
+## Supported Versions
+
+Use this section to tell people about which versions of your project are
+currently being supported with security updates.
+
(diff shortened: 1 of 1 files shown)


anindyakhansa-creator:
I hope it goes as I wish

FineFindus:
> There is no history in Chrome and it is safe with strict security

What? Why did you just add a security template?

ch2a:
> > There is no history in Chrome and it is safe with strict security
> 
> What? Why did you just add a security template?

I'm pretty sure this is a bot. Best to block them from the repository.

Bnyro:
We've already had a similar PR before that created the exact same file - certainly bots, yeah.

PR #8020, opened 2025-12-20
Enhancement: Min Video for feed in case older videos are not present

### LINKED ISSUE #8015: Make MAX_FEED_AGE_DAYS adjustable

It would be great if we can make this adjustable, which enables user to decide how many videos they want on their feed.

### Other details

app/src/main/java/com/github/libretube/repo/LocalFeedRepository.kt L53
app/src/main/java/com/github/libretube/repo/LocalFeedRepository.kt L196

### Acknowledgements

SE-Bilal-Ahmad:
I know its a very dumb solution but the problem is too tricky. I tried my best 

Closes: #8015

### DIFF
--- a/app/src/main/java/com/github/libretube/db/dao/SubscriptionsFeedDao.kt
+++ b/app/src/main/java/com/github/libretube/db/dao/SubscriptionsFeedDao.kt
@@ -21,8 +21,11 @@ interface SubscriptionsFeedDao {
     @Query("SELECT EXISTS (SELECT * FROM feedItem WHERE videoId = :videoId)")
     suspend fun contains(videoId: String): Boolean
 
-    @Query("DELETE FROM feedItem WHERE (uploaded < :olderThan AND uploaded != -1)")
-    suspend fun cleanUpOlderThan(olderThan: Long)
+    @Query("SELECT EXISTS (SELECT * FROM feedItem WHERE uploaderName = :uploaderName)")
+    suspend fun anyVideoExists(uploaderName: String): Boolean
+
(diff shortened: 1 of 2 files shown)


FineFindus (commented):
Thanks for the PR.

I don't quite understand why you check whether the channel doesn't have any videos when fetching the feed for it? It should be enough to just modify the cleanup SQL statement to keep a minimum amount of items and run it after the feed refreshing.

SE-Bilal-Ahmad:
Problem is at the very beginning of the Channel subscription there wont be any newer video nor any data in the db so we need to check if there arent any videos in the db and insert those videos. Now lets assume there is some video older than 3 months none of its entry will be in db and obviously hasnewerUploads would be false so we need to insert at least minimum videos in the db by checking if there arent in the db but upon further refresh we wont insert them again because now they would already be in the db so we dont delete them using cleanUpOlderThan and insert them again

FineFindus:
I see. The solution that you propose isn't great. For context, we currently fetch the channel twice, first using RSS (which is cheaper, but doesn't contain all the necessary information) and a second time using YouTube specific APIs (which is much more expensive, but has all the necessary information).
By fetching the feed every time we do not already have a video (which will often be the case, as the video will be remove before the next re-fetch if it's too old), we would need to fetch using the more expensive channel feed. This will increase power consumption and the time it takes to fetch the feed, as that API is more heavily rate-limited.
As this is only a problem if the feed has no items (i.e. the first time the feed is refreshed), we can either track the last refresh time externally (instead of using the last fetched video for that) or make having existing feed video an invalid state. This would require to fetch the feed every time a user subscribes to a channel, which may pose problems when importing many subscriptions at once.

PR #8034, opened 2025-12-27
feat(sleep-timer): show minutes left in player menu

PavelStefanov:
Hey!

I’d like to improve the SleepTimer UX by showing its current status in the player menu - whether it’s off, and if it’s running, how many minutes are left until it ends. This way users can see the current state at a glance without opening the sleep timer screen.
Screenshot_20251227-220926
Screenshot_20251227-220903

I’m new to open source, so I’d really appreciate any feedback or guidance on how to do this the best way 🙏

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/views/CustomExoPlayerView.kt
+++ b/app/src/main/java/com/github/libretube/ui/views/CustomExoPlayerView.kt
@@ -479,7 +479,7 @@ abstract class CustomExoPlayerView(
             R.drawable.ic_sleep,
             {
                 if (SleepTimer.timeLeftMillis > 0) {
-                    val minutesLeft = ceil(SleepTimer.timeLeftMillis.toDouble() / MILLIS_PER_MINUTE).toInt()
+                    val minutesLeft = ceil(SleepTimer.timeLeftMillis.toDouble() / DateUtils.MINUTE_IN_MILLIS).toInt()
                     context.resources.getQuantityString(R.plurals.minutes_left, minutesLeft, minutesLeft)
                 } else {
                     context.getString(R.string.disabled)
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/github/libretube/ui/views/CustomExoPlayerView.kt
```diff
@@ -474,7 +475,15 @@ abstract class CustomExoPlayerView(
         },
         BottomSheetItem(
             context.getString(R.string.sleep_timer),
-            R.drawable.ic_sleep
+            R.drawable.ic_sleep,
+            getCurrent = {
```
FineFindus:
```suggestion
            {
```

The explicit argument naming isn't necessary here.

app/src/main/java/com/github/libretube/ui/views/CustomExoPlayerView.kt
```diff
@@ -474,7 +475,15 @@ abstract class CustomExoPlayerView(
...
+                if (SleepTimer.timeLeftMillis > 0) {
+                    val minutesLeft = kotlin.math.ceil(SleepTimer.timeLeftMillis / 60000.0).toInt()
```
FineFindus: Please make `60000.0` a constant, to keep it in style with the other constants used in this file.

app/src/main/res/values/strings.xml
```diff
@@ -499,6 +499,7 @@
     <string name="no_segments_found">There are no segments for this video yet.</string>
     <string name="start_sleep_timer">Start sleep timer</string>
     <string name="disable_sleep_timer">Disable sleep timer</string>
+    <string name="minutes_left">%d min left</string>
```
FineFindus: I think this needs to support [pluralization](https://developer.android.com/guide/topics/resources/string-resource#Plurals), as not every language uses singular abbreviations.

app/src/main/java/com/github/libretube/ui/views/CustomExoPlayerView.kt
FineFindus: Please import the `ceil` function

FineFindus (changes requested):
Thanks, it does make sense to display the sleep timer state, as we already display the state for the other actions.

PavelStefanov:
@FineFindus  thank you for the feedback!

app/src/main/java/com/github/libretube/ui/views/CustomExoPlayerView.kt
dhanuarf: We can use android's built-in `DateUtils.MINUTE_IN_MILLIS`, so we dont need to create our own constant

FineFindus (approved):
Thanks! :tada:

Issue #4575, opened 2023-08-21
Display chapters in Timebar


FineFindus:
### Describe your suggested feature

The official YouTube application shows chapters on its timeline, by breaking it up using spaces. This allows users to quickly glance at the end of a chapter. It also allows easy navigation to subsequent chapters through scrubbing.

I'm planning to implement a similar feature in LibreTube. Sadly, [Exoplayer doesn't natively support chapters](https://github.com/google/ExoPlayer/issues/3903).
- There's a [library](https://github.com/vkay94/YouTubeTimeBar) available. It appears to have been unmaintained for some time, so a fork and subsequent maintenance might be necessary. Moreover, it contains functionalities that are redundant given our current time bar design.
- Draw out own timebar. However, this could lead to inconsistencies with [Exoplayer's time bar design](https://github.com/google/ExoPlayer/blob/03cadd2c809596e8922129f03ff2212618414734/library/ui/src/main/java/com/google/android/exoplayer2/ui/DefaultTimeBar.java#L883). The playhead's drawing mechanism would also need to be re-implemented, since it's private.
- Soft-fork Exoplayer. Considering the minimal changes needed, this might be a feasible approach to introduce chapter support.

I'm uncertain about the most appropriate approach and would appreciate feedback before diving in.

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

Bnyro:
We should neither add an additional library for that nor fork exoplayer which would mean there would be an additional amount of efforts and delayed updates from the media3 dependency.

The best way would be to implement our own time bar in my opinion - that could also allow us to improve the rendering of SponsorBlock segments on the time bar.

Otherwise we could also draw separators on the time bar similar to how we draw SponsorBlock segments, but that wouldn't allow us to makr the separator transparent.

FineFindus:
I already guessed that that would be the preferred approach.

> improve the rendering of SponsorBlock segments on the time bar.

Not sure what's meant by this? How should it be improved?

> Otherwise we could also draw separators on the time bar similar to how we draw SponsorBlock segments, but that wouldn't allow us to makr the separator transparent.

I already experimented with this approach and can't really recommend it. It may also conflict with the colors used by the Sponsorblock segments.

Bnyro:
> Not sure what's meant by this? How should it be improved?
>
[...]

PR #8038, opened 2025-12-30
feat: display chapters in timebar

FineFindus:
Displays the chapters of a video as breaks in the progress bar of a video. This allows users to quickly glance at the end of a chapter and makes scrubbing through chapters easier.
The breaks are implemented by erasing the existing paint, which includes both the SponsorBlock segments and the time bar itself. This has the drawback of also erasing the part of the scrubber if it's in the same position as a break, however it is barely noticeable in actual use.
To fix this we could completely write our own timebar implementation (which would help with code cleanup and future timebar features, e.g. heat maps and allow us to improve the accessibility), though it would require taking on additional complexity and maintenance.

| Before                                                                                                       	| After                                                                                                     	|
|--------------------------------------------------------------------------------------------------------------	|-----------------------------------------------------------------------------------------------------------	|
| Timebar without chapters 	| Timebar with chapters 	|

Closes: https://github.com/libre-tube/LibreTube/issues/4575

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/views/ChapterTimeBar.kt
+++ b/app/src/main/java/com/github/libretube/ui/views/ChapterTimeBar.kt
@@ -29,11 +29,11 @@ class ChapterTimeBar(
     private val progressBarHeight = 2f.dpToPx()
 
     override fun onDraw(canvas: Canvas) {
-        val checkpoint = canvas.saveLayer(null, null)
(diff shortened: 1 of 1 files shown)


FineFindus:
> This has the drawback of also erasing the part of the scrubber if it's in the same position as a break, however it is barely noticeable in actual use.

Actually, that's also the case for SponsorBlock segments already

PR #8048, opened 2026-01-04
Improve sleep timer with smooth fade out

PavelStefanov:
Added smooth exponential volume fade out to sleep timer for better UX. 
The volume gradually decreases using a quadratic curve during the last 10 seconds, 
which sounds more natural than linear fade out.

No other functionality was modified - snackbar and app closing logic remain unchanged.


app/src/main/java/com/github/libretube/services/AbstractPlayerService.kt
```diff
@@ -183,6 +183,11 @@ abstract class AbstractPlayerService : MediaLibraryService(), MediaLibrarySessio
                 }
                 updateNotification()
             }
+
+            args.containsKey(PlayerCommand.SET_VOLUME.name) -> {
+                val volume = args.getFloat(PlayerCommand.SET_VOLUME.name)
+                setPlayerVolume(volume)
+            }
```
Bnyro: This method declaration is redundant, you can already update the volume using
```kotlin
playerController.volume = 0.5f
```
,
so we don't need to implement a custom command here

app/src/main/java/com/github/libretube/ui/tools/SleepTimer.kt
```diff
@@ -1,74 +1,158 @@
...
-                snackBar.setText(context.getString(R.string.take_a_break) + remainingTime)
-            }
+
+    private fun setPlayerVolume(volume: Float) {
+        val act = activity ?: return
+
+        val serviceClass = if (isOffline) {
+            OfflinePlayerService::class.java
+        } else {
+            OnlinePlayerService::class.java
```
Bnyro: I don't like having this if/else here, perhaps it'd be better if `SleepTimer` doesn't know there are two different possible services.

So we should move that responsibility to `BackgroundHelper` by creating a method

```kotlin
getRunningPlayerServiceClass(context: Context) -> Class {
}
```
that checks if any of these services is running and returns the first one running that it finds.

app/src/main/java/com/github/libretube/ui/tools/SleepTimer.kt
```diff
@@ -1,74 +1,158 @@
...
+    private const val SNACKBAR_START_SECONDS = 5L
+    private const val MAX_VOLUME = 1.0f
+    private const val CLOSE_DELAY = 500L // Pause before closing
+
+    private var isOffline: Boolean = false
     var timeLeftMillis: Long = 0L
+        private set
+    private var timer: Timer? = null
+    private var snackBar: Snackbar? = null
+    private var activity: BaseActivity? = null
```
Bnyro: We should ideally not make `activity` an attribute in this class because this is prone to memory leaking it, i.e. when calling `activity.destroy()` it still cannot be freed out of memory because we still have a reference to it here, so this `activity` attribute might outlive the `BaseActivity` instance.

Hence, we should simply pass the `activity` attribute to the `onTimerTick` method to avoid having a global reference to it

app/src/main/java/com/github/libretube/ui/tools/SleepTimer.kt
```diff
@@ -1,74 +1,158 @@
...
+            }, CLOSE_DELAY)
+        }
+    }
+
+
+    private fun showOrUpdateSnackBar() {
+        val act = activity?.takeIf { !it.isFinishing && !it.isDestroyed } ?: run {
+            stop()
+            return
+        }
```
Bnyro:
```suggestion
    private fun showOrUpdateSnackBar(activity: BaseActivity) {
        if (activity.isFinishing || activity.isDestroyed) {
            stop()
            return
        }
```
nit: This code looks unnecessarily complicated.
1. Please don't name this variable `act`, we should avoid using such abbreviations that are not self-explaining.
2. Better would be to pass `activity` as parameter to `showOrUpdateSnackBar` so that its value is static and we don't have to do this magic here

app/src/main/java/com/github/libretube/ui/tools/SleepTimer.kt
PavelStefanov: Thanks for the feedback! 

I went a bit further and moved all the volume adjustment logic into BackgroundHelper. Now SleepTimer just calls BackgroundHelper.setVolume() and doesn't need to know anything about the services

app/src/main/java/com/github/libretube/ui/tools/SleepTimer.kt
PavelStefanov: You're right about the memory leak. I reverted to the original approach where context gets passed through the timer instead of storing it globally

PavelStefanov:
@Bnyro thank you for the detailed review! Really appreciate the feedback, it helped improve the implementation a lot

app/src/main/java/com/github/libretube/ui/tools/SleepTimer.kt
Bnyro: That's even better, nice! :+1:

PR #8049, opened 2026-01-04
Add quick-select chips to sleep timer

PavelStefanov:
Added preset duration chips (10, 20, 30, 45, 60 minutes) to the sleep timer bottom sheet. 

As a user, I want to quickly pick a duration without typing it in manually every time. Now you can just tap a chip to instantly set a common duration, making the sleep timer much faster to use for typical scenarios while still keeping manual input available for custom durations.

image

### DIFF
--- a/app/src/main/res/layout/sleep_timer_sheet.xml
+++ b/app/src/main/res/layout/sleep_timer_sheet.xml
@@ -59,8 +59,9 @@
                     android:id="@+id/quick_select_container"
                     android:layout_width="match_parent"
                     android:layout_height="wrap_content"
-                    android:layout_marginHorizontal="10dp"
                     android:layout_marginTop="10dp"
+                    android:paddingHorizontal="10dp"
+                    android:clipToPadding="false"
                     android:scrollbars="none">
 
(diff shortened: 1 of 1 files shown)


FineFindus (changes requested):
Thanks for the PR. Instead of defining each chip manually in the layout file, please create them programmatically, to avoid duplication and drift between the actual duration value and the displayed values.

See https://github.com/libre-tube/LibreTube/blob/81f9721ac67193c56521fbc3582ccd54da39e41a/app/src/main/java/com/github/libretube/ui/fragments/SubscriptionsFragment.kt#L246 or https://github.com/libre-tube/LibreTube/blob/81f9721ac67193c56521fbc3582ccd54da39e41a/app/src/main/java/com/github/libretube/ui/sheets/PlaybackOptionsSheet.kt#L38

For consistency reasons, the chips should also be displayed above the text field, similar to how the quick selection for speed values is displayed above the slider.

PavelStefanov:
@FineFindus thanks for taking the time to review! 
Fixed both comments - chips are now created programmatically and displayed above the text field like the speed selector.

PR #8059, opened 2026-01-09
fix: If the trend category is not supported in the country, fallback to category live

### LINKED ISSUE #7891: If the trend type is not supported in the country, the trend content cannot be changed after the page is reloading
### Steps to reproduce

1. Choose Germany as the trend region 
2. Then select "Podcasts" as the trend type. 
3. Refresh the home page or close and reopen the app. 
4. The "trend type" can be changed since the Trend page is no longer on the main page. 

### Expected behavior

If content is not supported at the time of selection, revert to supported content 

### Actual behavior

https://github.com/user-attachments/assets/7cc1dd73-a576-461d-9102-51bbd2c22ab5

### LibreTube version

0.29.0 (63)

### Android version

Android 13

### Other details

....

### Acknowledgements

VeryUsual:
fixes #7891

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/fragments/HomeFragment.kt
+++ b/app/src/main/java/com/github/libretube/ui/fragments/HomeFragment.kt
@@ -122,7 +122,11 @@ class HomeFragment : Fragment(R.layout.fragment_home) {
             val currentTrendingCategoryPref = PreferenceHelper.getString(
                 PreferenceKeys.TRENDING_CATEGORY,
                 TrendingCategory.LIVE.name
-            ).let { categoryName -> trendingCategories.first { it.name == categoryName } }
+            ).let { categoryName ->
+                trendingCategories.firstOrNull { it.name == categoryName }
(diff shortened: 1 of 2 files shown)


Bnyro (changes requested):
https://github.com/libre-tube/LibreTube/blob/master/app/src/main/java/com/github/libretube/api/NewPipeMediaServiceRepository.kt#L255

`MediaServiceRepository.instance.getTrendingCategories()` always returns the same value hardcoded above, so it's impossible that `trendingCategories.firstOrNull { it.name == categoryName }` returns null and hence your code can't work.

Have you actually tested this? The proper way would be to listen for errors when loading the trending videos and disable an error message or other way for the user to change the trends settings if the trends fail to load :)

VeryUsual:
I tested it although definitely not thoroughly and I just tested it after I did the changes and stuff didn't crash, I'll close this and do a better implementation, sorry.

Bnyro:
No worries, thanks for the PR anyways :)

PR #8061, opened 2026-01-10
Prevent video from moving to the start when re-adding to local playlist

PavelStefanov:
When you try to add a video that's already in a local playlist, it now gets silently skipped instead of being moved to the start of the playlist

Why?
* Users don't expect videos to jump around when re-adding them
* Keeps playlist order intact
* Matches expected behavior from YouTube, etc


PavelStefanov:
@Bnyro Thanks for your help! Fixed it

Bnyro (approved):
Thanks!

PR #8068, opened 2026-01-12
feat: include filtered search history in search suggestion

dhanuarf:
Changes:
- Combine both search history (filtered) and search suggestions in a single list. 
- The data collections are being redesigned using chained flows.
-  Suggestions `RecyclerView` list is now being displayed in normal order instead of reversed.
- `SearchHistoryAdapter` is deleted since we dont need that anymore.
- `SearchHistoryDao.getAll()` result is now ordered by its `rowid` to ensure we get the correct order
-  Included fix 1041fcda1b146d1ed4b27b39f69e244f032f80d1, this fix ensure recent submitted searches inserted to the database in the correct order

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/activities/MainActivity.kt
+++ b/app/src/main/java/com/github/libretube/ui/activities/MainActivity.kt
@@ -34,6 +34,8 @@ import com.github.libretube.compat.PictureInPictureCompat
 import com.github.libretube.constants.IntentData
 import com.github.libretube.constants.PreferenceKeys
 import com.github.libretube.databinding.ActivityMainBinding
+import com.github.libretube.db.DatabaseHelper
+import com.github.libretube.db.obj.SearchHistoryItem
 import com.github.libretube.enums.ImportFormat
 import com.github.libretube.enums.TopLevelDestination
 import com.github.libretube.extensions.anyChildFocused
@@ -300,6 +302,17 @@ class MainActivity : BaseActivity() {
         )
     }
 
+    private fun addSearchQueryToHistory(query: String) {
+        val searchHistoryEnabled =
+            PreferenceHelper.getBoolean(PreferenceKeys.SEARCH_HISTORY_TOGGLE, true)
+        if (searchHistoryEnabled && query.isNotEmpty()) {
+            lifecycleScope.launch(Dispatchers.IO) {
+                val newItem = SearchHistoryItem(query.trim())
+                DatabaseHelper.addToSearchHistory(newItem)
+            }
+        }
+    }
+
     override fun invalidateMenu() {
         // Don't invalidate menu when in search in progress
         // this is a workaround as there is bug in android code
@@ -336,6 +349,8 @@ class MainActivity : BaseActivity() {
                 }
 
                 navController.navigate(NavDirections.showSearchResults(query))
+
+                addSearchQueryToHistory(query)
 
                 return true
             }
--- a/app/src/main/java/com/github/libretube/ui/fragments/SearchResultFragment.kt
+++ b/app/src/main/java/com/github/libretube/ui/fragments/SearchResultFragment.kt
@@ -16,19 +16,14 @@ import androidx.paging.LoadState
 import androidx.recyclerview.widget.GridLayoutManager
 import androidx.recyclerview.widget.RecyclerView
 import com.github.libretube.R
-import com.github.libretube.constants.PreferenceKeys
 import com.github.libretube.databinding.FragmentSearchResultBinding
-import com.github.libretube.db.DatabaseHelper
-import com.github.libretube.db.obj.SearchHistoryItem
 import com.github.libretube.extensions.ceilHalf
-import com.github.libretube.helpers.PreferenceHelper
 import com.github.libretube.ui.activities.MainActivity
 import com.github.libretube.ui.adapters.SearchResultsAdapter
 import com.github.libretube.ui.base.DynamicLayoutManagerFragment
 import com.github.libretube.ui.extensions.setOnBackPressed
 import com.github.libretube.ui.models.SearchResultViewModel
 import com.github.libretube.util.TextUtils.toTimeInSeconds
-import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.flow.collectLatest
 import kotlinx.coroutines.launch
 import okhttp3.HttpUrl.Companion.toHttpUrlOrNull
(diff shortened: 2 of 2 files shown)


app/src/main/java/com/github/libretube/ui/adapters/SearchSuggestionsAdapter.kt
```diff
@@ -2,15 +2,81 @@ package com.github.libretube.ui.adapters
...
+                histories = oldList.filter { it.type == SearchDataType.HISTORY }
+
+                histories + (suggestionList!!.map { SearchItemData(it, SearchDataType.SUGGESTION) })
+            } else { // suggestionList == null
+                suggestions = oldList.filter { it.type == SearchDataType.SUGGESTION }
+
+                (historyList.map { SearchItemData(it, SearchDataType.HISTORY) }) + suggestions
+            }
+
+        submitNewList(newList)
```
Bnyro:
```suggestion
        val oldList = currentList.toImmutableList()
        val histories = historyList
            ?.map { SearchItemData(it, SearchDataType.HISTORY) }
            ?: oldList.filter { it.type == SearchDataType.HISTORY }
        val suggestions = suggestionList
            ?.map { SearchItemData(it, SearchDataType.SUGGESTION) }
            ?: oldList.filter { it.type == SearchDataType.SUGGESTION }

        val newList = histories + suggestions
        super.submitList(newList.distinctBy { it.query }, commitCallback)
```

I would prefer simplifying it this way to make it easier to read.

I know that this way, we make the `ImmutableList` copy (`val oldList = currentList.toImmutableList()`) even if we don't need it, but performance shouldn't be an issue here anyways because the amount of suggestions (<15 iirc) + history (<MAX_SEARCH_HISTORY_SIZE = 20 by default) is always quite small, so it won't matter.

app/src/main/java/com/github/libretube/ui/adapters/SearchSuggestionsAdapter.kt
```diff
@@ -19,8 +85,34 @@ class SearchSuggestionsAdapter(
...
         holder.binding.apply {
+            when (data.type) {
+                SearchDataType.HISTORY -> {
+                    deleteHistory.isVisible = true
+                    deleteHistory.setOnClickListener {
+                        val updatedList = currentList.toMutableList().also {
+                            it.remove(data)
+                        }
+                        runBlocking(Dispatchers.IO) {
+                            Database.searchHistoryDao().delete(SearchHistoryItem(suggestion))
```
Bnyro: You probably copied this from the old implementation, but in theory we should never use `runBlocking` on the UI thread...

We're also using that in some other places where we shouldn't because it could cause the UI to freeze, but yeah...

app/src/main/java/com/github/libretube/ui/models/SearchViewModel.kt
```diff
@@ -1,45 +1,116 @@
...
+     * Emits a [Pair] of ([SearchDataType.HISTORY] and the results ([List]))
+     */
+    @OptIn(ExperimentalCoroutinesApi::class)
+    private val filteredSearchHistory = combine(searchQuery, searchHistoryList) { query, list ->
+        query to list
+    }
+        .mapLatest { (query, list) ->
+            when {
+                query.isNullOrBlank() -> list
+                else -> list.take(MAX_FILTERED_SEARCH_HISTORY).filter { it.startsWith(query) }
```
Bnyro: nit: if there are only two cases, using a simple `if else` condition is cleaner

app/src/main/java/com/github/libretube/ui/models/SearchViewModel.kt
```diff
@@ -1,45 +1,116 @@
...
+            }
+        }
+        .mapLatest { SearchDataType.HISTORY to it }
```
Bnyro: nit: this can be collapsed with the `mapLatest` above, there's no reason to use `mapLatest` again here

app/src/main/java/com/github/libretube/ui/models/SearchViewModel.kt
```diff
@@ -1,45 +1,116 @@
...
                     Log.e("failed to fetch suggestions", e.stackTraceToString())
-
-                    return@withContext emptyList()
                 }
+
+                list
             }
-            else emptyList()
-        }
+            .mapLatest { SearchDataType.SUGGESTION to it }
```
Bnyro: nit: simplify to
```kt
 .mapLatest { (query, suggestionsEnabled) ->
                if (!suggestionsEnabled || query.isNullOrBlank()) {
                    return@mapLatest emptyList()
                }
                try {
                    MediaServiceRepository.instance.getSuggestions(query)
                } catch (e: Exception) {
                    Log.e("failed to fetch suggestions", e.stackTraceToString())
                    emptyList()
                }
}
```

Bnyro:
Really nice improvement, thanks for the PR :tada:

app/src/main/java/com/github/libretube/ui/adapters/SearchSuggestionsAdapter.kt
dhanuarf: > history (<MAX_SEARCH_HISTORY_SIZE = 20 by default) is always quite small, so it won't matter.

Yes if the users dont have unlimited search enabled in their settings. Regarding this history limit,  I wanted to say something about this but forgot to mention it in the PR. 

After this change merged , I think people will find their search history more useful than before because now you can see them as you type your query instead instead of just disappear. So, I'd like to propose a change so that we either:

1. Increase the default max history size; or
2. Option 1 but also remove the "Unlimited search history" checkbox and then we decide for good whats the limit; or
3. We make it customizable using text input (the same as watch history settings) replacing the "Unlimited search history" checkbox.

I think the most ideal option is we go with option 2, but option 2 and 3 are breaking changes so we might have to notify users on app update.

dhanuarf:
Force pushed with the suggestions and some changes:
-  Data class`SearchItemData` is renamed to `SearchDataItem` and moved to `.obj` package as a new file.
- The search suggestions result mapping is now done in the flow, so instead of submitting list of `String` we submit list of `SearchDataItem` to the adapter.
- The search history deletion is done outside the adapter now via lambda. And since we are using flow from the DAO, we dont need to submit the updated list ourselves.

app/src/main/java/com/github/libretube/ui/adapters/SearchSuggestionsAdapter.kt
Bnyro: I'm definitely in favor of option 2, as I think that the preference is quite useless anyways.

I don't think we have to notify the users via an in-app dialog though, it would be sufficient to label the commit as `BREAKING CHANGE: remove unlimited search history option` in the changelog and maybe add a note in the release description.

Issue #8099, opened 2026-01-23
"The page needs to be reloaded" error


borysl:
### Steps to reproduce

Reproducible in 90% of videos

1. Search any video
2. Click on it to play
3. Video loaded for a while

This is reproducable since yesterday.
I have tried to switch off piped

### Expected behavior

Video starts

### Actual behavior

![Image](https://github.com/user-attachments/assets/be6a1d71-1bfa-4650-81f5-46b3d0fb5a60)

Receive message "The page needs to be reloaded"
Video is empty (continue loading)

### Workaround

Just retry 4-5 times in a row.

### LibreTube version

0.29.1

### Android version

Android 18

### Other details

**One UI version** 8.0
Samsung A54

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [x] I have written a short but informative title.
- [x] I will fill out all of the requested information in this form.

borysl:
Surprising enough - even for already loaded video it might happen (if I refuse to replay)

RAHUL12348697695848579:
I have also had this issue it's a playback issue and not a ip issue as when you play any video most of the time it loads but majority of the times now you gotta go and reload the page I am  on Android 15

RAHUL12348697695848579:
Again does it two times in a row and sometimes it freezes without giving the this page needs to be reloaded  the other time it had happened to me 4 times in a row so not sure what that problem was

RAHUL12348697695848579:
![Image](https://github.com/user-attachments/assets/4110dda7-98d4-4214-b8d0-a1dc1e5fe8aa) I get this same but 4 times in a row

RAHUL12348697695848579:
I am also in ver 0.29.1 and here is a video so you can better see it https://vimeo.com/1157457370?fl=pl&fe=sh
Also newpipe is also having the same error I guess youtube really did somthing 
https://github.com/TeamNewPipe/NewPipe/issues/13093

Figim:
I reported this issue on the matrix. Maybe this is a problem with the Extractor. Yt has not changed anything. Ytdlp is not affected

It happens often now.

RAHUL12348697695848579:
I see keep us updated if possible

Figim:
https://github.com/TeamNewPipe/NewPipe/issues/13082

Seq0Fool:
[...]

PR #8113, opened 2026-01-28
Fix page reload error

Pittvandewitt:
Closes https://github.com/libre-tube/LibreTube/issues/8099

This PR switches back to the official NewPipeExtractor. The fork implements a fix that has been merged in the official extractor, so I assume it is safe to switch back to the original extractor: https://github.com/TeamNewPipe/NewPipe/issues/12907


FineFindus:
Thanks for the PR. However, we would prefer to continue to use the extractor fork, as it allows us to merge features and fixes early, without having to wait for upstream merging it.

I'll rebase the extractor fork later today.

PR #8121, opened 2026-01-28
feat: fast forward action now increases playback speed instead of skipping frames

### LINKED ISSUE #8060: Hold to 2x speed skips frames
### Steps to reproduce

lnstall the latest debug versIon
open app and play a video
hold the screen to initiate 2x speed

lt skips frames while playing in 2x.

### Expected behavior

It should play the video in 2x while playing the video not skipping frames

### Actual behavior

It dosa increase the video speed to 2x but it doesnt play it as it skaps the frames

### LibreTube version

Latest debug

### Android version

Android 14

### Other details

_No response_

### Acknowledgements

Bnyro:
closes #8060

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/views/CustomExoPlayerView.kt
+++ b/app/src/main/java/com/github/libretube/ui/views/CustomExoPlayerView.kt
@@ -29,6 +29,7 @@ import androidx.core.view.updateLayoutParams
 import androidx.fragment.app.findFragment
 import androidx.lifecycle.findViewTreeLifecycleOwner
 import androidx.media3.common.C
+import androidx.media3.common.PlaybackParameters
 import androidx.media3.common.Player
 import androidx.media3.common.text.Cue
 import androidx.media3.session.MediaController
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/github/libretube/ui/views/CustomExoPlayerView.kt
```diff
@@ -861,25 +882,38 @@ abstract class CustomExoPlayerView(
...
-                player?.seekBy(PlayerHelper.FAST_FORWARD_INCREMENT)
-                delay(PlayerHelper.FORWARD_INCREMENT_DELAY)
-            }
+        val player = player ?: return
+
+        // manually set playback speed is already larger than fast forward speed,
+        // so using the fast forward action would actually slow down seeking
+        if (player.playbackParameters.speed >= PlayerHelper.FAST_FORWARD_SPEED) {
+            return
         }
```
FineFindus, line 895: Why not simply multiply by `2`?

Bnyro: That's indeed a good catch, didn't think about that.

We should keep in mind that `*4` is the maximum, but we can simply clamp there.

PR #8122, opened 2026-01-28
refactor: change playback speed relatively to current speed in fast forward action

Bnyro:


app/src/main/java/com/github/libretube/helpers/PlayerHelper.kt
```diff
@@ -68,7 +68,12 @@ object PlayerHelper {
      *
      * Should be kept in sync with `fast_forward_view.xml`
      */
-    const val FAST_FORWARD_SPEED = 3f
+    const val FAST_FORWARD_SPEED_FACTOR = 2f
+
+    /**
+     * Maximum playback speed supported by ExoPlayer
+     */
+    const val MAXIMUM_PLAYBACK_SPEED = 4f
```
FineFindus, line 76: I think it's actually `8`?

https://github.com/androidx/media/blob/01c37572827dc77543fe6d0deda3efe06ec95f61/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/DefaultAudioSink.java#L520

Bnyro: Strange, I thought we were using the maximum value in the playback options sheet (4 is the maximum used there).

But yeah, the linked code seems to be the better source of truth.

Bnyro: I guess it probably won't have an impact on the average user anyways

FineFindus: I usually use playback speed near `4`, but I also don't use the fast forward feature, so :shrug:

Bnyro: Yeah, same...

PR #8149, opened 2026-01-31
fix: subscription list is displaying outdated data

dhanuarf:
Fetch subscription list on every lifecycle started in `SubscriptionsBottomSheet` and `EditChannelGroupSheet` to ensure the latest data is displayed.

- This fixes an issue where recently unsubbed channels are being re-shown when these sheets opened. This change also make sure newly subbed channels are showing without the need to refresh the feed first.

- `fetchSubscriptions()` function in `SubscriptionsFragment` is removed because I think we dont need that anymore?
---
While making this patch, I came across this line in `MainActivity.setupSubscriptionsBadge()`:
 https://github.com/libre-tube/LibreTube/blob/141b5fb70da1793ac8ba19c4692900649cc07192/app/src/main/java/com/github/libretube/ui/activities/MainActivity.kt#L270  
Is this a typo and supposed to be `getFeed()`?

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/sheets/SubscriptionsBottomSheet.kt
+++ b/app/src/main/java/com/github/libretube/ui/sheets/SubscriptionsBottomSheet.kt
@@ -3,10 +3,13 @@ package com.github.libretube.ui.sheets
 import android.annotation.SuppressLint
 import android.os.Bundle
 import android.view.View
-import androidx.core.content.ContextCompat
 import androidx.core.view.isVisible
 import androidx.core.widget.addTextChangedListener
 import androidx.fragment.app.activityViewModels
+import androidx.lifecycle.Lifecycle
+import androidx.lifecycle.asFlow
+import androidx.lifecycle.lifecycleScope
+import androidx.lifecycle.repeatOnLifecycle
 import com.github.libretube.R
 import com.github.libretube.api.obj.Subscription
 import com.github.libretube.constants.PreferenceKeys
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/github/libretube/ui/sheets/SubscriptionsBottomSheet.kt
```diff
@@ -62,39 +85,40 @@ class SubscriptionsBottomSheet : ExpandedBottomSheet(R.layout.sheet_subscription
...
+                    .show(parentFragmentManager, null)
+            }
+
+            // refresh displayed list of channels when channel groups have been edited
+            if (binding.groupSubsBtn.isChecked) {
+                showFilteredSubscriptions()
+            }
+        }
+        channelGroupsModel.groups.observe(viewLifecycleOwner) { groups ->
+
```
Bnyro: That observer doesn't do anything, perhaps you forgot to remove this after testing something?

Bnyro (approved):
Looks good to me, thanks for the PR (and sorry for the delay with the review)!

> fetchSubscriptions() function in SubscriptionsFragment is removed because I think we dont need that anymore?
>
Indeed, that's not needed anymore, good catch!

app/src/main/java/com/github/libretube/ui/sheets/SubscriptionsBottomSheet.kt
dhanuarf: Yeah, that was really silly of me, good catch!

PR #8150, opened 2026-01-31
feat: use RYD Proxy API

FineFindus:
Switches from the default Return YouTube Dislike API, as it theoretically allows for tracking the user's viewing habits, to the Piped RYD Proxy. This has been discussed before, but was conclude to not be necessary. However, recently the developer added a premium version that allows [access to some analytics](https://www.patreon.com/posts/advanced-update-141941708), which again raises privacy concerns.

Ref: https://github.com/libre-tube/LibreTube/pull/6483

### DIFF
--- a/app/src/main/java/com/github/libretube/api/ExternalApi.kt
+++ b/app/src/main/java/com/github/libretube/api/ExternalApi.kt
@@ -21,7 +21,7 @@ import retrofit2.http.Url
 
 private const val GITHUB_API_URL = "https://api.github.com/repos/libre-tube/LibreTube/releases/latest"
 private const val SB_API_URL = "https://sponsor.ajay.app"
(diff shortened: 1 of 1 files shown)


Bnyro (approved):
Given the fact that Kavin's services have often been unreliable recently, I'd be more happy with that if we had control of the RYD proxy ourselves.

But it's probably not worth setting up a LibreTube VPS solely for RYD, and I don't think we have anything else that we would want to host for LibreTube... (website is served via GitHub pages, which is probably much more reliable and better distributed than a single VPS)

FineFindus:
The last few years I've been using it, it has been relatively stable.

> But it's probably not worth setting up a LibreTube VPS solely for RYD, and I don't think we have anything else that we would want to host for LibreTube...

I can think a few things, but yeah.

PR #8187, opened 2026-02-13
feat(fast-foward): enable with double-tap-to-seek and allow higher playback speeds

FineFindus:
Updates the maximum playback speed to the same max. playback speed as used by exoplayer, allowing it to be used on all currently available playback speeds.
Also enabled fast-forward when double tap to seek is enabled, instead of swipe gestures, as it doesn't involve any swiping and can be used similar to double tap to seek, to navigate through a video.

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/views/CustomExoPlayerView.kt
+++ b/app/src/main/java/com/github/libretube/ui/views/CustomExoPlayerView.kt
@@ -883,7 +883,7 @@ abstract class CustomExoPlayerView(
     }
 
     override fun onLongPress() {
-        if (!PlayerHelper.swipeGestureEnabled) return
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/github/libretube/ui/views/CustomExoPlayerView.kt
FineFindus, line 904: I'm honestly not sure, whether we should remove these checks entirely (considering that they will now always succeed) and simply add a comment to check that to speed factor works/accept the risk that it breaks with we change it in the future.

Bnyro: I'd prefer to keep them to be on the safe side, we don't know what the future brings for this code 🤷‍♂️

Issue #8185, opened 2026-02-12
Minor UX improvements


p0wl90:
Hey,
I came here to suggest some UX improvements that are probably pretty easy to implement but make quite a big difference from the user perspective:

1. Skipping forward by double tapping (e.g. 15s) works for jumping once but is tedious for more than one step. Other players implement: 2 taps: 15s, 3 taps: 30s,... 
2. The hitbox for the time slider is pretty small and often missed. maybe increase it a little bit (the hitbox of the dot as well as the hitbox of the line of the slider)
3. When users disable "Play automatically" and open a video, the play button (or all player controls) shold be visible. Maybe, it also makes sense to load the thumbnail instead of the first frame.

Thanks for your work!

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [x] I have written a short but informative title.
- [x] I will fill out all of the requested information in this form.

Bnyro:
> The hitbox for the time slider is pretty small and often missed. maybe increase it a little bit (the hitbox of the dot as well as the hitbox of the line of the slider)
>
We already increased the touch area for the timebar recently, I don't think that's an issue anymore personally.

> Skipping forward by double tapping (e.g. 15s) works for jumping once but is tedious for more than one step. Other players implement: 2 taps: 15s, 3 taps: 30s,...
>
https://github.com/libre-tube/LibreTube/pull/8198

> When users disable "Play automatically" and open a video, the play button (or all player controls) shold be visible. Maybe, it also makes sense to load the thumbnail instead of the first frame.
>
https://github.com/libre-tube/LibreTube/pull/8199 now automatically shows the player controls - I don't think it's worth loading the thumbnail instead of the first frame here, considering the required implementation efforts.

PS: In the future, please open one issue per feature suggestion, and don't put multiple suggestions into a single issue.

p0wl90:
thanks for the quick implementation and I wont put multiple requests in one issue in the future.

sheltpat:
I know this is closed, but I would like to add something to point 2:

The bar does not accept simple taps, does it?
For me it never triggers upon a tap. I need to either drag the circle, or tap and move my finger on the bar to be accepted.

FineFindus:
> The bar does not accept simple taps, does it?

No, this was removed in https://github.com/libre-tube/LibreTube/pull/7757

sheltpat:
[...]

PR #8192, opened 2026-02-15
fix: accumulate double-tap seek and show correct increment (#8185)

foxoffire33:
Pause playback while user is tapping, accumulate the total seek amount, and display the running total. Icon spins continuously during taps. Text slide-out animation is deferred until tapping stops, then the batched seek is applied and playback resumes. Text size scales down for 3+ digit numbers to fit inside the icon.

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/views/CustomExoPlayerView.kt
+++ b/app/src/main/java/com/github/libretube/ui/views/CustomExoPlayerView.kt
@@ -1,5 +1,6 @@
 package com.github.libretube.ui.views
 
+import android.animation.ObjectAnimator
 import android.annotation.SuppressLint
 import android.content.Context
 import android.content.pm.ActivityInfo
@@ -13,6 +14,7 @@ import android.util.Log
 import android.view.KeyEvent
 import android.view.MotionEvent
 import android.view.Window
+import android.view.animation.LinearInterpolator
 import android.widget.FrameLayout
 import android.widget.ImageView
 import android.widget.TextView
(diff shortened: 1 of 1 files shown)


Bnyro:
Hey, thanks for your interest in contributing. However, since this pull request seems to be written by using LLMs, we unfortunately can't accept these changes (reference https://github.com/foxoffire33/LibreTube/commit/dc9852d3664aff114c5e25a7d2c544e6cb9717fa with Claude as co-author.)

See https://github.com/libre-tube/LibreTube/pull/8193#issuecomment-3908134764 for our AI policy.

app/src/main/java/com/github/libretube/ui/views/CustomExoPlayerView.kt
```diff
@@ -528,93 +534,111 @@ abstract class CustomExoPlayerView(
...
-        // clear previous animation
-        imageView.animate()
-            .rotation(0F)
-            .setDuration(0)
-            .start()
+        if (seekIconAnimator?.target == imageView && seekIconAnimator?.isRunning == true) return
+
+        seekIconAnimator?.cancel()
+        val direction = if (isRewind) -1f else 1f
+        seekIconAnimator = ObjectAnimator.ofFloat(imageView, "rotation", 0f, direction * 360f).apply {
```
Bnyro, line 608: It doesn't make any sense here to use `ObjectAnimator`, `imageView.animate()
            .translationX(0f)
            .setDuration(0)` is a much more modern API.

Also, why does this remove the text animation? ...

PR #8193, opened 2026-02-16
fix: hide chapter button in audio mode when video has no chapters

### LINKED ISSUE #8027: Audio mode shows chapter button even if there are no chapters when switching modes
### Steps to reproduce

Open a video that does not have any chapters in video mode. Swap to audio mode. 

### Expected behavior

The chapter button is only shown if the video has chapters

### Actual behavior

The chapter button will be shown, no matter if the video has chapters

### LibreTube version

0.29.1

### Android version

Android 15

### Other details

_No response_

### Acknowledgements

pierreeurope:
## Summary

Fixes #8027 — The chapter button in the audio player was shown even when the video has no chapters, specifically when switching from video mode to audio mode.

## Root Cause

Two issues:
1. The `open_chapters` button had no default visibility set in the XML layout, so it defaults to `visible`
2. When connecting to the player service in `handleServiceConnection()`, the initial metadata was used to update stream info but the chapter button visibility was not checked — it was only updated in `onMediaMetadataChanged`, which doesn't fire if the metadata is already set

## Changes

- **`fragment_audio_player.xml`**: Set default visibility of the chapter button to `gone`
- **`AudioPlayerFragment.kt`**: Added initial chapter visibility check when the player service connects, using the same logic as `onMediaMetadataChanged`

## Testing

- Open a video without chapters → switch to audio mode → chapter button is hidden ✅
- Open a video with chapters → switch to audio mode → chapter button is shown ✅
- Open audio player directly for a video without chapters → chapter button is hidden ✅

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/fragments/AudioPlayerFragment.kt
+++ b/app/src/main/java/com/github/libretube/ui/fragments/AudioPlayerFragment.kt
@@ -465,7 +465,15 @@ class AudioPlayerFragment : Fragment(R.layout.fragment_audio_player), AudioPlaye
                 _binding?.openChapters?.isVisible = !chapters.isNullOrEmpty()
             }
         })
-        playerController?.mediaMetadata?.let { updateStreamInfo(it) }
+        playerController?.mediaMetadata?.let { mediaMetadata ->
+            updateStreamInfo(mediaMetadata)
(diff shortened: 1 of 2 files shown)


FineFindus:

Thanks for understanding.

PR #8198, opened 2026-02-18
feat: allow seeking with single tap after first double click

Bnyro:
see #8185

cc @dhanuarf, you probably know this code better than me since you wrote it :)

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/listeners/PlayerGestureController.kt
+++ b/app/src/main/java/com/github/libretube/ui/listeners/PlayerGestureController.kt
@@ -13,6 +13,7 @@ import androidx.activity.viewModels
 import com.github.libretube.ui.base.BaseActivity
 import com.github.libretube.ui.interfaces.PlayerGestureOptions
 import com.github.libretube.ui.models.CommonPlayerViewModel
+import java.time.Instant
 import kotlin.math.abs
 
 class PlayerGestureController(activity: BaseActivity, private val listener: PlayerGestureOptions) {
@@ -30,6 +31,7 @@ class PlayerGestureController(activity: BaseActivity, private val listener: Play
     private var scaleGestureWasInProgress = false
     private var isMoving = false
     var longPressInProgress = false
+    var lastDoublePressTime: Instant? = null
 
     var areControlsLocked = false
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/github/libretube/ui/listeners/PlayerGestureController.kt
```diff
@@ -165,10 +167,25 @@ class PlayerGestureController(activity: BaseActivity, private val listener: Play
...
         override fun onSingleTapConfirmed(e: MotionEvent): Boolean {
+            // if there has been a recent double click within the threshold, single clicks
+            // are being treated as a repetition of the double click
+            // e.g. this allows to continue seeking forward using single clicks after double-clicking once
+            lastDoublePressTime?.takeIf {
+                val millisElapsedSinceDoubleClick = Instant.now().toEpochMilli() - it.toEpochMilli()
+                millisElapsedSinceDoubleClick < DOUBLE_TAP_REPEAT_TIME_THRESHOLD_MILLIS
+            }?.let {
+                e.action = MotionEvent.ACTION_UP
+                onDoubleTapEvent(e)
```
dhanuarf: I read somewhere (cant remember where) that it's a bad idea to change the event action directly. If we want to inject a fake event action it's recommended that we copy the existing MotionEvent using `obtain()` and then call `recycle()` after we're done using it, like how we did it in other places:
```kotlin
val upEvent = MotionEvent.obtain(e)
upEvent.action = UP
onDoubleTapEvent(upEvent)
upEvent.recycle()
```

Bnyro: Thanks for the hint.

I wasn't sure what the best way to clone a `MotionEvent` is, so your explanation definitely makes sense.

Bnyro:
> Though IMO injecting a double tap event from a single tap event is kind of unintuitive, since it might cause "a little" maintainability issue because the function `onDoubleTap()` doesnt behave like a double tap.
> 
> But I think this is the simplest way to implement this feature, so maybe it's fine for now?
>
Yeah, it definitely is the easiest way to add it. In theory we could subclass `GestureDetector.SimpleOnGestureListener` to also handle repeated double tap events, but I guess that would be an overkill for such a small feature.

Thanks for the review.

Issue #8067, opened 2026-01-12
apply SponsorBlock rules to downloads


atloo1:
### Describe your suggested feature

When using the online player, I have segments auto skipped. This functionality is absent in downloaded videos. 

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [x] I have written a short but informative title.
- [x] I will fill out all of the requested information in this form.

slimshady045:
simultaneous coment reading similar to online mode

PR #8204, opened 2026-02-20
feat: store sponsorblock segments in database

Bnyro:
this is a pre-requisite for #8067

The UI part (i.e. actual skipping of downloaded segments) will likely be done by using the online player for all downloads, see https://github.com/libre-tube/LibreTube/pull/7928.

### DIFF
--- a/app/src/main/java/com/github/libretube/api/obj/Segment.kt
+++ b/app/src/main/java/com/github/libretube/api/obj/Segment.kt
@@ -2,12 +2,14 @@ package com.github.libretube.api.obj
 
 import android.os.Parcelable
 import androidx.collection.FloatFloatPair
+import com.github.libretube.db.obj.DownloadSponsorBlockSegment
 import kotlinx.parcelize.IgnoredOnParcel
 import kotlinx.parcelize.Parcelize
 import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/github/libretube/api/obj/Segment.kt
```diff
@@ -26,6 +28,21 @@ data class Segment(
     @IgnoredOnParcel
     val segmentStartAndEnd = FloatFloatPair(segment[0], segment[1])
 
+    // reminder: all the attributed that are asserted as non-null here are declared non-null by the
+    // SponsorBlock API, so it's safe to assert they're not null (if we trust the SponsorBlock API docs)
```
FineFindus, line 32: Any reason not to change them to non-null in the class?

Bnyro: I assumed that this would require changes in many places (i.e. removing null-checks) and I didn't want to touch any unrelated code in this PR.

If you're motivated, I think there's nothing stopping us from making them `non-null`.

FineFindus: Done in https://github.com/libre-tube/LibreTube/pull/8205. Luckily, all the other code is accessing them from lists, so we need the null checks anyway.

PR #8205, opened 2026-02-20
refactor: mark required SponsorBlock properties as non-null

FineFindus:
These values are guaranteed by the API to be always set, so the should be marked as non-null.

Ref: https://github.com/libre-tube/LibreTube/pull/8204#discussion_r2835163436


Bnyro:
Thanks 👍

PR #8229, opened 2026-03-01
feat: use background to show selected items in bottomsheet

FineFindus:
Instead of adding unicode checkmark symbol after the icon name, use a differently colored background, which is automatically selected when the textview is marked as selected. This also helps with accessibility, as assistive tools are informed which item is selected.
| Before                                                                                                           	| After                                                                                                         	|
|------------------------------------------------------------------------------------------------------------------	|---------------------------------------------------------------------------------------------------------------	|
| Selection without background 	| Selection with background 	|


Bnyro:
Have you found some issues with this? I.e. why did you mark it as a draft?

FineFindus:
Yes, there are contrast issues between the text and the background color for some background colors in dark mode. I have a fix ready, but would like to test it a bit more first.

PR #8235, opened 2026-03-03
fix: prevent playlist download from stalling on unavailable videos

bermount:
When there is a private/unavailable video in a playlist, "videoInfo = null" previously skipped amountOfVideosDone count causing the service to never complete. This logic now ensures the progress counter is always incremented even if a video cannot be fetched.


Bnyro (approved):
Thanks!

We should probably do proper error handling here in the future, but that's better than the old behavior for now.

PR #8238, opened 2026-03-04
feat: rename `Mark as unwatched` → `Reset playback position`

FineFindus:
Renames the option to mark a video as unwatched to `Reset playback position`, this better indicates what the option does and avoids having two very similarly named options directly next to each other.


Bnyro:
But doesn't this also delete it from the watch history?

FineFindus:
It does.

cotienbang:
I don't think it should be renamed to something sound this nerdy.

Bnyro:
I hope you're not too opinionated about this change @FineFindus, but I think it's better to keep it as is because "Reset playback position" doesn't sound like it would also be removed from the watch history.

PR #8239, opened 2026-03-04
fix(video player): video appears in small corner

### LINKED ISSUE #5816: Opening a video from the mini-player moves the video to a corner
### Steps to reproduce

1. Open the video
2. Swipe it down to the mini-player
3. Open the video from the mini-player

### Expected behavior

The video should open normally.

### Actual behavior

The video shifts to the corner after opening.

### LibreTube version

0.22.1

### Android version

Android 14

### Other details

If you open the video full screen / rewind the video, it will fall into place. Here's a screen recording:
<details>

https://github.com/libre-tube/LibreTube/assets/67514343/561698db-6cc5-47b4-9823-d048dd179270

</details>

### Acknowledgements

flexxxxer:
after video's first frame rendered app slightly changes and shortly after restores container size where video being played from. it is done unnoticeably to the app user; very hacky fix

Personally, I hope better way without hacks exists. Still, this is what I can provide for a target issue

Closes #5816

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
+++ b/app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
@@ -230,6 +230,18 @@ class PlayerFragment : Fragment(R.layout.fragment_player), CustomPlayerCallback 
     private var bufferingTimeoutTask: Runnable? = null
 
     private val playerListener = object : Player.Listener {
+        override fun onRenderedFirstFrame() {
+            super.onRenderedFirstFrame()
+            val oldProgress = binding.playerMotionLayout.progress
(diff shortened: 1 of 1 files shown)


FineFindus:
Thanks, I appreciate the idea, though I would prefer to find the root cause and implement a proper fix, rather than having a hack, that may lead to other issues down the line (or a bunch of hacks which make it difficult to maintain the application). If I remember correctly, this bug didn't always exists, so if you have a reliable way of reproducing it, it might be worthwhile to try to bisect it :)

flexxxxer:
I will mark that PR as draft, feel free to close it any time soon. I will add notes about consistent way to reproduce that bug in issue comments

flexxxxer:
This fix causes issues in fullscreen. Next time will test better

PR #8242, opened 2026-03-04
feat: already set auto-fullscreen on shorts before rendering starts

Bnyro:
This is a rather hacky implementation because `PlayerFragment` doesn't have access to `videoStreams` (i.e. only the `AbstractPlayerService` has access to it). So when the `Streams` are loaded for the first time, the `isShort` variable has to be updated using `videoStreams`, and may never be changed afterwards.

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
+++ b/app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
@@ -250,7 +250,7 @@ class PlayerFragment : Fragment(R.layout.fragment_player), CustomPlayerCallback 
 
         override fun onPlaybackStateChanged(playbackState: Int) {
             // set the playback speed to one if having reached the end of a livestream
-            if (playbackState == Player.STATE_BUFFERING && binding.player.isLive &&
+            if (playbackState == Player.STATE_BUFFERING && streams.isLive &&
                 playerController.duration - playerController.currentPosition < 700
             ) {
                 playerController.setPlaybackSpeed(1f)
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/github/libretube/api/obj/Streams.kt
```diff
@@ -48,11 +48,21 @@ data class Streams(
...
     @IgnoredOnParcel
     val isLive = livestream || duration <= 0
 
+    init {
+        // this can't be lazily computed because not all streams instances have access to videoStreams
+        videoStreams.firstOrNull()?.let {
+            isShort = (it.height ?: 0) > (it.width ?: 0)
+        }
+    }
+
```
FineFindus: Why not set this when retrieving the streams? Also NPE has a `isShortForm` method which returns whether a video is a short.

Bnyro: True, that makes sense 🫣

Looks like I've been overthinking this

FineFindus: Actually, I was wrong, the NPE method always returns false (as it's using the same player as for non-shorts videos and there unfortunately doesn't seem to be a way to differentiate between the two), sorry.

Bnyro: Oh :/

Anyways your solution is cleaner, I'll just copy the code from the Piped media repo to the NPE media repo later

FineFindus:
I have been thinking about that a bit more. In the majority of cases we actually know a priori whether a video is a short (e.g. it was started from the shorts tab, is a related short video, etc.), maybe it would be worthwhile to pass that to the player? It would speed up the auto-fullscreen and avoid relying on the aspect-ratio hack.

Bnyro:
> maybe it would be worthwhile to pass that to the player
>
It would, but it would also require a lot of refactoring (i.e. you have to pass `isShort` everywhere where you can start a video from or navigate to an other video). Currently our logic for that is really awkward (we pass everything as an individual bundle argument), I planned to create a data class that contains all relevant information so that it's easier to handle, but too lazy. It would look smth like
```kt
data class StartVideoArguments(
  val videoId: String,
  val isShort: Boolean,
  val shuffle: Boolean,
  val timeStamp: Long,
  ...
)
```
and we would use that in `NavigationHelper` and `BackgroundHelper`.

PR #8313, opened 2026-04-14
add recommendation system

Developer246:
add a recommendation system based on history and add some improvements


Bnyro:

Thanks for understanding.

PR #8377, opened 2026-04-28
feat: show SponsorBlock icon on video label

FineFindus:
Adds a badge for the selfpromotion category, as well as the same icon used for the thumbnail label, to help users associate the thumbnail icons to the labels. 
Sponsorblock badge with icon

### DIFF
--- a/app/src/main/java/com/github/libretube/api/SponsorBlockLabelHelper.kt
+++ b/app/src/main/java/com/github/libretube/api/SponsorBlockLabelHelper.kt
@@ -2,6 +2,7 @@ package com.github.libretube.api
 
 import android.util.LruCache
 import androidx.annotation.DrawableRes
+import androidx.annotation.StringRes
 import com.github.libretube.R
 import com.github.libretube.api.obj.VideoLabelData
 import com.github.libretube.extensions.sha256Sum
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/github/libretube/api/SponsorBlockLabelHelper.kt
```diff
@@ -30,4 +32,17 @@ object SponsorBlockLabelHelper {
...
         }.getOrNull()
     }
+
+    /**
+     * Returns a suitable drawable to display the category.
+     *
+     * If there is no matching icon, `null` is returned.
+     */
+    @DrawableRes
+    fun icon(category: String?): Int? = when (category) {
```
Bnyro:
```suggestion
    fun getIconForLabelCategory(category: String?): Int? = when (category) {
```
Intellij otherwise starts suggesting this method to you if you just type `icon`, so it's confusing.

app/src/main/java/com/github/libretube/ui/adapters/VideoCardsAdapter.kt
```diff
@@ -135,14 +133,11 @@ class VideoCardsAdapter(private val columnWidthDp: Float? = null) :
...
-                                "selfpromo" -> context.getDrawable(R.drawable.ic_selfpromo_content)
-                                "sponsor" -> context.getDrawable(R.drawable.ic_paid_content)
-                                else -> null
-                            }
-                        )
+                        SponsorBlockLabelHelper.icon(sponsor?.segments?.firstOrNull()?.category)?.let {
+                            sponsorBadgeIcon.setImageDrawable(
+                                context.getDrawable(it)
+                            )
+                        }
```
Bnyro, line 141: I think we should also set `sponsorBadgeCard.tooltipText` here (or whatever the attribute is called in Kotlin, at least it's `tooltipText` in XML). To the same text as https://github.com/libre-tube/LibreTube/pull/8377/changes#diff-867bcbc2370e5a8906b3c70e7236a018a28cd91451a75c1bb1db96fc50dcce77R59 I guess.

app/src/main/java/com/github/libretube/api/SponsorBlockLabelHelper.kt
FineFindus: Hmm, I'm not sure what you're referring to, since I've never had it suggest class method without using the class.

app/src/main/java/com/github/libretube/ui/adapters/VideoCardsAdapter.kt
FineFindus: I set it on the icon, since that should also be helpful for screen readers.

Bnyro (approved):

PR #8378, opened 2026-04-28
fix search, feed, channel page in full local mode

ItsMeSamey:
Currently, the search does not work, nor does refreshing the feed (when in full local mode).
Trying to load a channel page gives you 503.

This pr fixes all of these

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -335,6 +335,7 @@
                 <data android:pathPrefix="/shorts/" />
                 <data android:pathPrefix="/live/" />
                 <!-- channel prefix -->
+                <data android:pathPrefix="/@" />
                 <data android:pathPrefix="/channel/" />
                 <data android:pathPrefix="/user/" />
                 <data android:pathPrefix="/c/" />
--- a/app/src/main/java/com/github/libretube/api/NewPipeMediaServiceRepository.kt
+++ b/app/src/main/java/com/github/libretube/api/NewPipeMediaServiceRepository.kt
@@ -428,7 +428,7 @@ class NewPipeMediaServiceRepository : MediaServiceRepository {
     }
 
     override suspend fun getChannel(channelId: String): Channel {
-        val channelUrl = "$YOUTUBE_FRONTEND_URL/${channelId}"
+        val channelUrl = if (channelId.startsWith("UC") && channelId.length == 24) "$YOUTUBE_FRONTEND_URL/channel/$channelId" else "$YOUTUBE_FRONTEND_URL/$channelId"
         val channelInfo = ChannelInfo.getInfo(NewPipeExtractorInstance.extractor, channelUrl)
 
         val channel = channelInfo.toChannel()
@@ -475,7 +475,7 @@ class NewPipeMediaServiceRepository : MediaServiceRepository {
     }
 
     override suspend fun getChannelNextPage(channelId: String, nextPage: String): Channel {
-        val url = "$YOUTUBE_FRONTEND_URL/${channelId}/videos"
+        val url = if (channelId.startsWith("UC") && channelId.length == 24) "${YOUTUBE_FRONTEND_URL}/channel/${channelId}/videos" else "${YOUTUBE_FRONTEND_URL}/${channelId}/videos"
         val listLinkHandler = ListLinkHandler(url, url, channelId, listOf("videos"), "")
         val tab = getChannelTab(listLinkHandler.toTabDataString(), nextPage)
         return Channel(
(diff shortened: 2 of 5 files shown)


app/src/main/AndroidManifest.xml
```diff
@@ -335,6 +335,7 @@
                 <data android:pathPrefix="/shorts/" />
                 <data android:pathPrefix="/live/" />
                 <!-- channel prefix -->
+                <data android:pathPrefix="/@" />
```
Bnyro, line 338: That's a good idea, I agree on this at least.

Bnyro:
> Currently, the search does not work, nor does refreshing the feed (when in full local mode). Trying to load a channel page gives you 503.
>
Most of your changes are only targeting if the user uses Piped, not full local mode. So your PR actually changes the wrong code?

Your PR does a lot of unrelated stuff, so please clarify the intention first.

app/src/main/java/com/github/libretube/api/NewPipeMediaServiceRepository.kt
```diff
@@ -427,8 +428,30 @@ class NewPipeMediaServiceRepository : MediaServiceRepository {
         return relatedStreamsResp.content.map { it.toStreamItem() } to relatedStreamsResp.nextpage
     }
 
+    private fun buildChannelUrl(channelReference: String): String {
```
Bnyro: Why? What's wrong with the previous way how we built channel URLs?

app/src/main/java/com/github/libretube/extensions/ToID.kt
```diff
@@ -14,10 +21,48 @@ fun String.toID(): String {
         .removePrefix(YOUTUBE_MUSIC_URL)
         .removePrefix(YOUTUBE_SHORT_URL)
         .replace("/watch?v=", "") // videos
-        .replace("/channel/", "") // channels
         .replace("/playlist?list=", "") // playlists
+        .replace("watch?v=", "") // videos without leading slash
+        .replace("playlist?list=", "") // playlists without leading slash
```
Bnyro: That pattern doesn't exist? What's an example for this?

app/src/main/java/com/github/libretube/extensions/ToID.kt
```diff
@@ -14,10 +21,48 @@ fun String.toID(): String {
...
+        .removePrefix("/")
+        .removePrefix("channel/") // channels
+        .removePrefix("user/") // legacy usernames
+        .removePrefix("c/") // custom channel urls
```
Bnyro:
```suggestion
        .removePrefix("/channel/") // channels
        .removePrefix("/user/") // legacy usernames
        .removePrefix("/c/") // custom channel urls
```

Bnyro: We should probably also remove `/@` prefixes here?

app/src/main/java/com/github/libretube/extensions/ToID.kt
```diff
@@ -14,10 +21,48 @@ fun String.toID(): String {
...
         // channel urls for different categories than the main one
         .removeSuffix("/shorts")
         .removeSuffix("/streams")
         .removeSuffix("/videos")
 }
+
+fun String.isCanonicalChannelId() = canonicalChannelIdRegex.matches(this)
+
+fun String?.toChannelReference(): ChannelReference? {
```
Bnyro: Again, I don't think we need this method. `toID()` already does that.

app/src/main/java/com/github/libretube/ui/fragments/ChannelFragment.kt
```diff
@@ -210,19 +218,6 @@ class ChannelFragment : Fragment(R.layout.fragment_channel) {
         TabLayoutMediator(binding.tabParent, binding.pager) { tab, position ->
             tab.text = tabList[position].name
         }.attach()
-
-        channelAdapter = VideosAdapter(showChannelInfo = false).also {
-            it.submitList(response.relatedStreams)
-        }
-        tabList.clear()
```
Bnyro, line 217: why change this?

app/src/main/java/com/github/libretube/repo/PipedNoAccountFeedRepository.kt
Bnyro, line 1: undo, we don't want to have implicit fallbacks like this.

Just don't use Piped if you don't have a working instance.

app/src/main/java/com/github/libretube/repo/PipedAccountFeedRepository.kt
Bnyro, line 1: undo, we don't want to have implicit fallbacks like this.

Just don't use Piped if you don't have a working instance.

app/src/main/java/com/github/libretube/api/PipedMediaServiceRepository.kt
Bnyro, line 1: undo, we don't want to have implicit fallbacks like this.

Just don't use Piped if you don't have a working instance.

ItsMeSamey:
IDK how but this somehow automatically fixed itself. It's a Heisenbug i suppose!

PR #8406, opened 2026-05-10
Follow-up for #8305

j-lakeman:
Sorry it's actually `www.youtube-nocookie.com` and the `www` seems to be important.
Follow-up for #8305


Bnyro (approved):
You're right, thanks!

PR #8412, opened 2026-05-14
Revert "feat(playback-options): set speed steps to 0.25"

### LINKED ISSUE #8352: Bring back adjustable playback speed

In version 3.0, users could adjust the playback speed increment, but versions 31.0 and 31.1 removed this option.

Please bring it back; it is very useful.

In the current version, users can only set the playback speed increment at a fixed 0.25. In the previous version, they could adjust it in the settings.

A 0.25 increment does not work for everyone.

Or set a fixed 0.05 increment so users get more granular control.

### Other details

_No response_

### Acknowledgements

deckstose:
This reverts commit d4d6d7170a5dab02daad1dc849d18861a6df8b84.

Closes: https://github.com/libre-tube/LibreTube/issues/8352

### DIFF
--- a/app/src/main/res/layout/playback_bottom_sheet.xml
+++ b/app/src/main/res/layout/playback_bottom_sheet.xml
@@ -59,9 +59,9 @@
                             android:id="@+id/speed"
                             android:layout_width="match_parent"
                             android:layout_height="wrap_content"
-                            android:stepSize="0.25"
+                            android:stepSize="0.05"
                             android:value="1.0"
-                            android:valueFrom="0.25"
+                            android:valueFrom="0.2"
                             android:valueTo="4.0"
                             android:layout_marginTop="5dp"
                             android:layout_marginHorizontal="10dp" />


FineFindus:
Thanks for the patch. We currently have no plans to revert to the old design, however we're open to improvements to the current one and will be keeping on eye on feedback about it.
You're of course welcome to maintain the patch yourself in your own fork :)

deckstose:
Can you give some background from a maintainers point of view why this cannot be reverted or you are not interested in this?

FineFindus:
Sure, happy to. The current design is preferable because it follows the established interaction pattern used by most video players. That familiarity reduces the cognitive overhead for users, especially new ones, when selecting a different playback speed.Using much smaller selection steps makes it much, much harder to select a specific speed, especially on thinner devices/with different controls. Larger, more well-spaced options are simpler to navigate and select. Additionally, a step size of 0.25 offers a more meaningful distinction between different playback speeds, than the much smaller steps.
Ideally we would have some data to see what users actually prefer, but unfortunately telemetry is greatly disliked in open source projects. As already said, we will monitor the feedback and adjust the design, if it continues to be an issue.

PR #8422, opened 2026-05-14
Preserve subscription feed thumbnail data

mzl2233:
This keeps existing subscription feed metadata when a refreshed local feed item omits fields that were already stored, including thumbnails. During refresh, the local feed repository now looks up saved feed rows by video ID and the feed item conversion falls back to the stored thumbnail and related metadata when the latest extracted item has missing values.

Verification: targeted content assertions for the fallback paths passed, and `git diff --check` passed. I also attempted `./gradlew :app:compileDebugKotlin --no-daemon -Dorg.gradle.java.home=/usr/lib/jvm/java-21-openjdk-amd64`, but this environment does not have an Android SDK configured (`SDK location not found`).

### DIFF
--- a/app/src/main/java/com/github/libretube/api/obj/StreamItem.kt
+++ b/app/src/main/java/com/github/libretube/api/obj/StreamItem.kt
@@ -45,19 +45,19 @@ data class StreamItem(
         )
     }
 
-    fun toFeedItem() = SubscriptionsFeedItem(
+    fun toFeedItem(existingFeedItem: SubscriptionsFeedItem? = null) = SubscriptionsFeedItem(
         videoId = url!!.toID(),
-        title = title,
-        thumbnail = thumbnail,
-        uploaderName = uploaderName,
-        uploaded = uploaded,
-        uploaderAvatar = uploaderAvatar,
-        uploaderUrl = uploaderUrl,
-        duration = duration,
-        uploaderVerified = uploaderVerified ?: false,
-        shortDescription = shortDescription,
-        views = views,
-        isShort = isShort
+        title = title ?: existingFeedItem?.title,
+        thumbnail = thumbnail ?: existingFeedItem?.thumbnail,
+        uploaderName = uploaderName ?: existingFeedItem?.uploaderName,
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/github/libretube/api/obj/StreamItem.kt
```diff
@@ -45,19 +45,19 @@ data class StreamItem(
...
+        thumbnail = thumbnail ?: existingFeedItem?.thumbnail,
+        uploaderName = uploaderName ?: existingFeedItem?.uploaderName,
+        uploaded = uploaded.takeIf { it > 0 } ?: existingFeedItem?.uploaded ?: uploaded,
+        uploaderAvatar = uploaderAvatar ?: existingFeedItem?.uploaderAvatar,
+        uploaderUrl = uploaderUrl ?: existingFeedItem?.uploaderUrl,
+        duration = duration ?: existingFeedItem?.duration,
+        uploaderVerified = uploaderVerified ?: existingFeedItem?.uploaderVerified ?: false,
+        shortDescription = shortDescription ?: existingFeedItem?.shortDescription,
+        views = views ?: existingFeedItem?.views,
+        isShort = isShort || existingFeedItem?.isShort == true
```
Bnyro, line 60: This is very ugly: As we only care about thumbnails, this `toFeedItem` method shouldn't be changed at all, and instead we should do

```kotlin
                   collectedFeedItems.flatten().map { streamItem ->
                        val existingFeedItem = existingFeedItems.getOrNull(streamItem.url?.toID())
                        if (streamItem.thumbnail.isNullOrEmpty()) streamItem.thumbnail = existingFeedItem.thumbnail
                        streamItem.toFeedItem()
                    }
```

Bnyro:
> Verification: targeted content assertions for the fallback paths passed, and git diff --check passed. I also attempted ./gradlew :app:compileDebugKotlin --no-daemon -Dorg.gradle.java.home=/usr/lib/jvm/java-21-openjdk-amd64, but this environment does not have an Android SDK configured (SDK location not found).
>
1. This looks very AI-generated: we don't accept AI PRs.
2. If you couldn't test it, you shouldn't submit a PR...

PR #8439, opened 2026-05-19
fix: default resolution not properly set

### LINKED ISSUE #8351: Default resolution not working on non fullscreen
### Steps to reproduce

1. Watch any video on non fullscreen. The quality will be in auto
2. When change to fullscreen, video will change to what user defined in video resolution settings.

### Expected behavior

Non fullscreen and fullscreen should have same resolution.

### Actual behavior

Non fullscreen and fullscreen show different resolution.

### LibreTube version

31.1

### Android version

Android 16

### Other details

Image

Image

Image

### Acknowledgements

dhanuarf:
Due to recent changes to the `CustomExoPlayerView`, the patch b77a95fc586bfe5b288a55ae7bee7cec9766cfd5 no longer work because we tried to update the resolution when the `Player` instance is not yet initialized.

The default resolution set call is now moved to `PlayerFragment` inside `setPlayerDefaults()`, so the function `updateResolution()` is now made public.

The player initialization logic also slightly modified, so that the playerView is initialized after we get the `MediaController/Player` ready, rather than calling `connectToPlayerView()` in `onViewCreated()`.

And then I realized the code can be slightly simplified more, so I made the change d057951fccd7ea212b37056a28454fb0f13f442f by merging `connectViewToPlayer()` with `initialize()` inside `CustomExoPlayerView`, so now we dont need to override the `setPlayer()` function since the initialization call only happen once already.

Closes: #8351

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/views/CustomExoPlayerView.kt
+++ b/app/src/main/java/com/github/libretube/ui/views/CustomExoPlayerView.kt
@@ -197,13 +197,15 @@ class CustomExoPlayerView(
         commonPlayerViewModel: CommonPlayerViewModel,
         playerViewModel: PlayerViewModel,
         viewLifecycleOwner: LifecycleOwner,
-        playerCallback: CustomPlayerCallback
+        playerCallback: CustomPlayerCallback,
+        player: Player,
     ) {
         this.chaptersViewModel = chaptersViewModel
         this.playerViewModel = playerViewModel
         this.commonPlayerViewModel = commonPlayerViewModel
         this.viewLifecycleOwner = viewLifecycleOwner
         this.playerCallback = playerCallback
+        super.player = player
 
         initializeGestureProgress()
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/github/libretube/ui/views/CustomExoPlayerView.kt
```diff
@@ -352,23 +354,6 @@ class CustomExoPlayerView(
             submitDialog.show((context as BaseActivity).supportFragmentManager, null)
         }
 
-        fullscreenResolution = PlayerHelper.getDefaultResolution(context, true)
-        noFullscreenResolution = PlayerHelper.getDefaultResolution(context, false)
-        updateResolution(commonPlayerViewModel.isFullscreen.value == true)
-    }
-
-    override fun setPlayer(player: Player?) {
```
Bnyro, line 360: To avoid future mistakes when fiddling with the code again, I think it would be good to still override `setPlayer` and just deprecate it. I.e. it shouldn't be used from outside the `CustomExoPlayerView`.

E.g. smth like
```kotlin
@Deprecated("Do not call this directly. Call `CustomExoPlayerView::initialize` instead!")
 override fun setPlayer(player: Player?) {
    super.setPlayer(player)
 }
```

Bnyro:
Very nice changes, thank you! This is much cleaner than the previous solution.

Overriding `setPlayer` has just been a very hackish work-around that I haven't really overthought - I just wanted a quick fix and that was the first thing that came into my mind.

Sorry for the from my side - I have been focusing on other FOSS projects recently, so I kinda lost LibreTube out of my eyes :)

app/src/main/java/com/github/libretube/ui/views/CustomExoPlayerView.kt
dhanuarf: Also since we still need to detach the player (setting it to null), so function `detachPlayer()` is added in [6d5f661](https://github.com/libre-tube/LibreTube/pull/8439/commits/6d5f661595b6dc376323e1b23f53496a74591fe9) so that we dont call a deprecated function

Bnyro (approved):
Again, thanks!

Issue #2186, opened 2022-12-02
Volume Booster in the player


GET100PERCENT:
### Describe your suggested feature

Some of the YouTube videos have a very low intensity sound. So it will be very useful if libretube supports volume Booster just like VLC, in which volume can be boosted upto 200%.

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

Bnyro:
Why I don't think it should be added to the app:
1. I doubt it would be used a lot as the uses cases are only very few
2. It decreases the quality of the audio, hence it wouldn't be too useful anyways
3. There's currently really no space for new controls anywhere at the player.
I'd recommend using a 3rd party app instead which should work fine too.

GET100PERCENT:
1. It decreases the quality of the audio, but not enough to be not understandable.
2. VLC has not any dedicated icon or indicator for volume Booster in the player interface.
![Screenshot_20221206-141927](https://user-images.githubusercontent.com/116144024/205864149-9630e1a9-53c7-4f20-bc6d-62c379d525a8.jpg)
It uses the same volume indicator instead of a dedicated one.

GET100PERCENT:
Okay 👍

ildar:
+1 to the topic starter
> 1. I doubt it would be used a lot as the uses cases are only very few

Unfortunatelly no. YouTube videos diverge a lot. The most troublesome in the wild are videos of old live conferences recorded with awful crappy sound. :-(

> 2. It decreases the quality of the audio, hence it wouldn't be too useful anyways

Nope if you turn it on for volumes >100%
> 3. There's currently really no space for new controls anywhere at the player.

No need for that. Please take a look how it's done in VLC. It's a masterpiece.
> I'd recommend using a 3rd party app instead which should work fine too.

Mostly very uncomfortable. Using 2 apps in the situation is a bad user experience.
Please re-triage.
And yes, thanks a lot for the app. It is really close to perfect and best in its class.

PR #8442, opened 2026-05-20
feat: Add an option to modify in-app video volume instead of system volume with fullscreen gestures

ymcx:
I'd like to start by expanding on why I think this feature is needed: Currently I'm using this app for subscribing to a couple of content creators and listening to music uploaded to YouTube. My IEMs seem to be too easy for my device to drive, which makes the vast majority of the volume steps available on the device useless, as anything past like 25% is way too loud for all kind of content. I'm sure this is an issue many other people are facing as well, since for all I know it should affect pretty much all mobile devices and a large amount of headphones regardless of the form factor. Currently when there is a song I really feel like listening to, I open it via the web browser in desktop mode and watch the ads, just so I can limit the volume further down via the desktop video player.

Features similar to this one have been discussed in past, yet they seem to get closed as wontfix due to a [couple of main issues:](https://github.com/libre-tube/LibreTube/issues/2186#issuecomment-1338955653)
1. Few people would benefit from the feature
2. It decreases the audio quality
3. There's no room in the video player

I already addressed the first issue and as for the second issue, allowing users to decrease the volume shouldn't decrease the audio quality at all and this PR doesn't allow users to go past 100% volume. As for the last point, this is something I've thought about a lot lately and in the end I settled on implementing it as an alternative for the swiping gesture that's already implemented.

Adding another button/slider to the player would be bad UX on mobile devices, hence why it's not done in pretty much any popular mobile video player. An additional setting in the menu would be confusing as well, especially considering it would add another place from which the volume could be modified on top of the player from which it can be altered via the gesture. Lastly, I figured it would make the most sense to implement this feature using the already implemented volume gesture, since it requires no additional code and is already used by a lot of users.

This feature is also quite trivial to implement with this approach, since we can just utilize 1) the already defined gestures and 2) the existing function for controlling the video volume, which was added for use with the sleep timer in https://github.com/libre-tube/LibreTube/commit/993888aa5cf92cc28ae9098cdccdd36cf0f04f5d.

With that said, this PR pretty much just adds a new setting to the player settings menu for controlling whether the fullscreen volume gesture controls the system audio (default/old behavior) or the video audio, which is 100% by default and can be set as anything from 0% to 100%. The setting is saved for the current runtime only, meaning you don't have to set it for each video but for each session instead. The aforementioned setting is greyed out if the gestures are disabled.


FineFindus (commented):
Thanks for the suggestion, I have some concerns, as I fail to see how this addresses, that it is a nice use-case. Additionally, a user may (accidentally, swipe gestures are not really discoverable) decrease the volume/completely mute the sound. With in-app volume, it's relatively hard for a user to restore the previous volume, opposed to simply increasing the system volume again.

> With that said, this PR pretty much just adds a new setting to the player settings menu for controlling whether the fullscreen volume gesture controls the system audio (default/old behavior) or the video audio, which is 100% by default and can be set as anything from 0% to 100%.

We generally prefer to add preference, especially when the only serve a niche use-case. If you believe this is the better default, it should be set for everyone.

ymcx:
> If you believe this is the better default, it should be set for everyone.

I figured it would make merging this PR much easier if I didn't change the default bahavior at all, hence why I instead opted to implement the feature as an optional setting the user can toggle.

Regarding whether I think the gesture should control the video volume by default, the answer is yes. I'm not sure if anyone else can relate, but I personally find it confusing that an in-app gesture can modify global, system level options such as the brightness level or the media volume.

On the other hand, I wouldn't want to have two gestures used at the same time, one of which controls an in-app setting and the other of which would control a system-level property. If it would be made the default, we should at the very least do something similar with the brightness, which in turn wouldn't make much sense since I can't see anyone realistically wanting to modify the video brightness, especially considering you can't go above the default of 100%, since it would only introduce overexposure and completely ruin the video quality.

Summing up, the new behavior should be configurable, disabled by default as is currently done in this PR. This is unless the brightness gesture is changed as well to something that makes sense rather than just applying a darkening filter on top of the video similarly to how the video volume is modified.

> a user may decrease the volume/completely mute the sound

That's a valid concern and something I hadn't really thought about yet. However, one could say the user wouldn't accidentally mute the volume since to do so they would first need to manually toggle the setting introced in this PR, in which case you would think the user would have at least some kind of understanding how the gesture should work in practice.

If you think it's necessary, there could be some kind of notification/message displayed for a (limited?) amount of time after the user has set the video volume as 0% if we'd assume that there's no rational reason for muting the video that way unless it has been done erroneously. 

You also shouldn't forget we already have an indicator displaying the modified value in the player, which could be enough for signaling the user they've muted the volume, accidentally or not.

> With in-app volume, it's relatively hard for a user to restore the previous volume

Agreed. The state persists for the length of the session and the volume can only be modified if there's a video playing. UX-wise this might not be ideal, but what's the alternative? Resetting the in-app volume on each video player instance wouldn't make much sense in cases such as mine, where the user wants to listen to a playlist consisting of music videos, all of which are too loud. We shouldn't make the user configure the audio for each video separately.

> I fail to see how this addresses, that it is a nice use-case

Do you mean you fail to see how this feature would prove itself useful? I somewhat agree it's unnecessary to have a gesture controlling a separate volume level just so one could have a more quiet baseline volume for too sensive speakers. But for the reasons mentioned in the initial comment, I figured this approach would make the most sense as compared to the alternative methods for achieving the same goal.

FineFindus:
> I figured it would make merging this PR much easier if I didn't change the default bahavior at all, hence why I instead opted to implement the feature as an optional setting the user can toggle.

I appreciate the thought, however it makes maintenance more challenging for us, as we now have to maintain and test an additional code path. It also increases the complexity for users when trying to use the settings. Thus, we've been moving away from preferences for every niche use case. Feel free to read https://ometer.com/preferences.html for more details or see https://github.com/libre-tube/LibreTube/issues/8465.

> Do you mean you fail to see how this feature would prove itself useful?

No. I fail to see how this would benefit more than a few people, i.e. I think it is still a niche use case.

ymcx:
> It also increases the complexity for users when trying to use the settings

I personally prefer opinionated software and agree that we shouldn't even present the user the possibility of modifying something in case we can provide a good default value. Because of that, I'm inclined to agree with you purely from this perspective. It's just that I really couldn't think of any alternative approaches to this other than if we'd change the default behavior, which we already discussed previously.

> I fail to see how this would benefit more than a few people

Fair enough. Do you want me to close this PR?

FineFindus:
> Fair enough. Do you want me to close this PR?

I would be ever so slightly leaning to not merging this, considering the current UX issues, though I think they might be solvable. Happy to let others discuss/decide (cc @Bnyro)

flexxxxer:
Feel free to open issue regardless adding that feature to [FlexTube](https://github.com/flexxxxer/FlexTube) and/or submitting PR on adding patch for that. I would like to add description to inform user about possible pitfalls like totally muting in-app sound, make that feature opt-in (disabled by default). If this feature will be part of LibreTube this is totally acceptable and I would be happy to maintain one less feature

FineFindus:
> Feel free to open issue regardless adding that feature to [FlexTube](https://github.com/flexxxxer/FlexTube) and/or submitting PR on adding patch for that.

While we have no problem with the fork (in fact we appreciate that the license makes it possible), we would kindly ask you to reduce your comments relating to it. The focus of this repository should be on the development of LibreTube, rather than about features of forks of it.

app/src/main/res/drawable/ic_volume.xml
Bnyro, line 1: This icon is just a volume icon and not related to gestures, so it should just be named `ic_volume.xml` or something similar

app/src/main/java/com/github/libretube/helpers/AudioHelper.kt
```diff
@@ -6,24 +6,41 @@ import androidx.core.content.getSystemService
...
 
-    var volume: Int
+    private var videoVolumeCached = 1f
+
+    var videoVolume: Float
+        get() = videoVolumeCached
+        set(value) {
+            videoVolumeCached = value.coerceIn(0f, 1f)
+            BackgroundHelper.setVolume(context, videoVolumeCached)
+        }
```
Bnyro:
```suggestion
   var videoVolume: Float = 1f
        set(value) {
            field = value.coerceIn(0f, 1f)
            BackgroundHelper.setVolume(context, field)
        }
```
I don't really understand why we need this caching logic here? You can just assign `field`, which changes the value of the variable in the setter.

app/src/main/java/com/github/libretube/helpers/AudioHelper.kt
```diff
@@ -6,24 +6,41 @@ import androidx.core.content.getSystemService
...
+        if (PlayerHelper.swipeGestureSystemVolume) {
+            systemVolume = value.normalize(minValue, maxValue, minimumVolumeIndex, maximumVolumeIndex)
+        } else {
+            videoVolume = value.toFloat().normalize(minValue.toFloat(), maxValue.toFloat(), 0f, 1f)
+        }
     }
 
     fun getVolumeWithScale(maxValue: Int, minValue: Int = 0): Int {
-        return volume.normalize(minimumVolumeIndex, maximumVolumeIndex, minValue, maxValue)
+        return if (PlayerHelper.swipeGestureSystemVolume) {
```
Bnyro, line 37: I already hated these `setVolumeWithScale` and `getVolumeWithScale` method when they were initially implemented - In my opinion, the setters and getters of `systemVolume` should also just be of type `Float` and work with values between `0` and `1` so that they work the same way as `videoVolume`.

You don't need to do this refactoring yourself, I can also do it after merging your PR.

Bnyro:
Sorry for the delay from my side, I got a friendly reminder from @FineFindus to look into it yesterday, so I'm doing that now :)

I personally wouldn't use this option either, but I understand the argumentation that you don't want the video player to mess with your system volume - which is reasonable. And I always appreciate contributions from new people :)

Hence, I'm okay with merging this. I've left a review with some suggestions to make things a bit cleaner.

Again, sorry for the delay :)

EDIT: The force push was only to rebase it on `master` to fix video playback issues - you can ignore it and just continue from where you stopped.

ymcx:
> Sorry for the delay from my side, I got a friendly reminder from @FineFindus to look into it yesterday, so I'm doing that now :)
> 
> I personally wouldn't use this option either, but I understand the argumentation that you don't want the video player to mess with your system volume - which is reasonable. And I always appreciate contributions from new people :)
> 
> Hence, I'm okay with merging this. I've left a review with some suggestions to make things a bit cleaner.
> 
> Again, sorry for the delay :)
> 
> EDIT: The force push was only to rebase it on `master` to fix video playback issues - you can ignore it and just continue from where you stopped.

Alright, I'll come to back to this once I have time. Been quite busy lately. Thanks for the feedback!

ymcx:
Thanks for the review! I don't use LibreTube anymore for unrelated reasons, so this PR isn't super high on my lists of priorities. Sorry about that.

>This icon is just a volume icon and not related to gestures, so it should just be named ic_volume.xml or something similar

I just realized we already have ic_volume_up.xml which is identical to the now-renamed ic_volume.xml with the exception of the color. Because of this, I kept both ic_volume_up.xml and ic_volume.xml.

>I don't really understand why we need this caching logic here? You can just assign field, which changes the value of the variable in the setter

As for the second suggested change, I simply just applied it (the suggested change) as is without really paying that much attention to it, mostly because GitHub allowed me to do so via the web interface and because I remember not being that satisfied with the approach I chose. I can't for the life of me remember why I even went with the cached approach, maybe it has something to do with me not really being that comfortable with Java/Kotlin and getting confused.

>I already hated these setVolumeWithScale and getVolumeWithScale method when they were initially implemented - In my opinion, the setters and getters of systemVolume should also just be of type Float and work with values between 0 and 1 so that they work the same way as videoVolume.

I haven't really worked on LibreTube before this PR, so I'm not totally in on why we have a scaled volume stored as an integer in the first place instead of a float [0,1]. To be honest, I found it quite comfusing while initially working on this PR. For reasons stated above, I'm not sure if I'm willing to look into this change in specific, at least for the foreseeable future.

>...And I always appreciate contributions from new people.
>Hence, I'm okay with merging this.

I'm not sure if you're okay with merging this purely because of the fact it's from a new contributor or because you agree that having an option for configuring the gesture would be useful, even if you wouldn't use it yourself. If it's the former, then I'd like to disagree. I personally find opinionated software satisfying in the sense that they have a strong direction they want to head towards rather than being a kitchen sink containing all kinds of features just because someone once submitted a PR. Especially UX-wise I'd rather not loosely accept all PRs.

Bnyro:
We kept the PR open for some time now to see if there's interest from other users in this feature, but it doesn't seem like it.

Thank you for your pull request, but ultimately I don't think it's worth it if even you don't need it anymore.

PR #8502, opened 2026-06-01
feat: use more user-friendly terms for audio tracks

FineFindus:
Switches to using for user-friendly terms for the audio tracks, by avoid technical terms and multiple types, such as 'original or main' instead of simply 'Original' or 'default or unknown' instead of simply 'Default'.

Ideally, we would also mark tracks as auto-translated as appropriate, but unfortunately, that's not possible with the current setup.

Audio tracks with user friendly terms

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/views/CustomExoPlayerView.kt
+++ b/app/src/main/java/com/github/libretube/ui/views/CustomExoPlayerView.kt
@@ -1121,9 +1121,7 @@ class CustomExoPlayerView(
                 .map {
                     PlayerHelper.getAudioTrackNameFromFormat(context, it)
                 },
-                preselectedItem = selectedAudioLanguageAndRoleFlags?.let {
-                    PlayerHelper.getAudioTrackNameFromFormat(context, it)
-                },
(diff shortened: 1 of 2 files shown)


FineFindus:
I also experimented with moving the track type to the subtitle, but not sure about the looks yet.

Track type as subtitle

Bnyro (approved):

Bnyro:
> I also experimented with moving the track type to the subtitle, but not sure about the looks yet.
> 
> Track type as subtitle
>
I think it's better to keep as is and instead maybe move the text into brackets like `English (Default)`.

If we show the language larger than the type, this implies that there can only be one "entry" per language - which is probably wrong, because it's e.g. possible to have a default and an audio description stream for the same language?

FineFindus:
> If we show the language larger than the type, this implies that there can only be one "entry" per language - which is probably wrong, because it's e.g. possible to have a default and an audio description stream for the same language?

Hmm, yes, that's a good point. Using brackets might look at bit weird for multiple audio tracks with the same name (e.g. 'English (US)' and 'English (GB)', though we currently display both as 'English').

Issue #8465, opened 2026-05-24
Please don't open issues about adding settings for very small behavior changes


Bnyro:
There currently are many options for very small behavior changes, e.g. default playback speed, automatic playback start, .... Currently, it likely takes about 20 minutes or more to go through LibreTube's settings page and understand everything.

On one hand, that makes it very difficult for new users to get into the app. On the other hand, that increases the maintenance burden, because we can't test all of these very specific options and hence they break frequently without us noticing.

In most cases, having good default settings fits 99% of the users very well, so we can keep things much simpler for the average user by focusing on good defaults instead of having too much customisability.

We're currently trying to remove a lot of options that don't add much value to the app, e.g. https://github.com/libre-tube/LibreTube/issues/7659. Of course we won't remove infinitely many options, fundamental settings like the theme, notifications, ... and privacy-related settings like automatic update checks, SponsorBlock, ... will stay forever.

Examples for such small behavior changes that we don't want to implement:
- https://github.com/libre-tube/LibreTube/issues/7906
- https://github.com/libre-tube/LibreTube/issues/8461
- https://github.com/libre-tube/LibreTube/issues/7956

Gitoffthelawn:
> having good default settings fits 99% of the users very well

What is the evidence to support this claim?  99% of the time this claim is made, it is false. ;)

FineFindus:
> What is the evidence to support this claim? 99% of the time this claim is made, it is false. ;)

This isn't about a hard rule, but having defaults that fit the majority well. You can also look at all the defaults the app already has, where none/few of our users ever complained about it :)

flexxxxer:
2/3 features listed implemented in [FlexTube](https://github.com/flexxxxer/FlexTube). There are other features too, most notable:
- dismissed channels
- ability to hide "shorts"
- ability to hide counters (likes, dislikes, view count)

Feel free to request behavior changes and settings for them in FlexTube. Keep in mind FlexTube's goal: provide distraction-free experience.

Bnyro:
> Keep in mind FlexTube's goal: provide distraction-free experience.
>

That's also LibreTube's goal - hence we are removing distracting options instead of adding new ones :)

flexxxxer:
> That's also LibreTube's goal - hence we are removing distracting options instead of adding new ones :)

[...]

PR #8525, opened 2026-06-10
Drop useless settings (image cache size and parallel downloads option)

Bnyro:
Please see the individual commit descriptions on why each individual feature gets removed.

related: https://github.com/libre-tube/LibreTube/issues/8465

### DIFF
--- a/app/src/main/java/com/github/libretube/helpers/ImageHelper.kt
+++ b/app/src/main/java/com/github/libretube/helpers/ImageHelper.kt
@@ -4,7 +4,9 @@ import android.content.Context
 import android.graphics.Bitmap
 import android.graphics.Color
 import android.net.Uri
+import android.os.storage.StorageManager
 import android.widget.ImageView
+import androidx.core.content.getSystemService
 import androidx.core.net.toUri
 import coil3.ImageLoader
 import coil3.disk.DiskCache
@@ -16,14 +18,12 @@ import coil3.request.ImageRequest
 import coil3.request.crossfade
 import coil3.toBitmap
 import com.github.libretube.BuildConfig
-import com.github.libretube.constants.PreferenceKeys
 import com.github.libretube.extensions.toAndroidUri
 import com.github.libretube.util.DataSaverMode
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.withContext
 import okhttp3.OkHttpClient
 import okhttp3.logging.HttpLoggingInterceptor
-import java.io.File
 import java.nio.file.Path
 
 object ImageHelper {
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/github/libretube/helpers/ImageHelper.kt
```diff
@@ -56,29 +52,18 @@ object ImageHelper {
...
                     memoryCachePolicy(CachePolicy.ENABLED)
 
-                    val diskCache = generateDiskCache(
-                        directory = context.coilFile,
-                        size = maxCacheSize.toInt()
-                    )
+                    val diskCache = DiskCache.Builder()
+                        .directory(context.coilFile)
+                        .maxSizeBytes(128 * 1024 * 1024L /* 128MB */)
+                        .build()
```
FineFindus: Considering the storage sizes of current devices and that the OS will delete files from the cache if it needs space, I think it's fine to use a larger cache size (e.g. 256MB or even more). Alternatively, we could also use [`getCacheQuotaBytes`](https://developer.android.com/reference/android/os/storage/StorageManager#getCacheQuotaBytes(java.util.UUID)) to check how much cache the OS has available for our app.

FineFindus (dismissed):
Removing the orientation setting breaks auto rotation and landscape, i.e. videos are still displayed portrait mode even when entering full screen.

app/src/main/java/com/github/libretube/helpers/ImageHelper.kt
Bnyro: Sounds like a very good idea, I didn't know about `getCacheQuotaBytes`. Implemented.

For me, it returns something like 120MB, which doesn't sound like that much? As per the docs, we can use more storage - the system will just wipe our cache first if it needs more space. So maybe we should multiply `getCacheQuotaBytes` with some factor?

I'm also not sure if we should use the whole cache size for images or not, but as I don't remember any other place in the app where we are caching something on the disk, I think it should be fine?

Bnyro:
> Removing the orientation setting breaks auto rotation and landscape, i.e. videos are still displayed portrait mode even when entering full screen.
>
Wdym by that? If I enter fullscreen, videos get played in fullscreen properly. The only bug I found is the visual glitch when rotating the screen - not sure what to do about that.

FineFindus:
> Wdym by that? If I enter fullscreen, videos get played in fullscreen properly. The only bug I found is the visual glitch when rotating the screen - not sure what to do about that.

What is the expected behavior after this change?

Bnyro:
> What is the expected behavior after this change?
>
The screen automatically rotates to the orientation the user wants? If the user has auto-rotation enabled, that's the orientation of the screen - otherwise if auto-rotation is disabled on system level, that's probably portrait.

Anyways, I've dropped the commit and will move it to a new draft PR.

Issue #7393, opened 2025-05-08
Swipe gestures


Joni8829:
### Describe your suggested feature

It would be cool if we could swipe left or right on songs in the playlist to perform several actions such as queue, play next, download etc.

Personally I would only be interested in queue next, but an option to change what the swiping does would be great!

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [x] I have written a short but informative title.
- [x] I will fill out all of the requested information in this form.

sanskarlohani:
is any one working on this particular enhancement.

Bnyro:
> is any one working on this particular enhancement.
>
no

Bnyro:
> Personally I would only be interested in queue next, but an option to change what the swiping does would be great!
>
I decided against this feature for the following reasons:
- swiping is usually used for deleting stuff everywhere in the app and for nothing else, so this would be counter-intuitive
- we already have a lot of settings which we're currently trying to get rid off and I think that this one would be yet another rather unnecessary option that'd be used by only few users.

sanskarlohani:
ok thats a great

Issue #4363, opened 2023-08-04
Quick actions (swipe left/right) inside feed


MFlisar:
### Describe your suggested feature

Inside the feed (abos, the view that shows the new videos of subscribed channels) I can long press a video and select from options like:

* play in background
* add to playlist
* download
* share
* mark as seen/unseen

My desired feature would be to be able to select any of the above options as "quick actions" that are executed when swiping left or right on a video.

### Other details

My use case would be to be able to mark a video as seen (and hide it from the feed) by simple swiping left and to download it by swiping right it - instead of long pressing it and click the corresponding button inside the bottom sheet menu... By adding the ability to select any action as quick action the usage possibilities would be a lot higher than my personal usage of this feature...

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

Bnyro:
Hey, thank you for sharing your idea, I appreciate your input.

However, I decided to reject the swiping gestures for the following reasons:
* Adding too many options for small behavior changes (in this case swiping actions) can clutter the interface, more specifically the preferences. The current design prioritizes simplicity and ease of use, and I believe that adding more options may detract from that.
* Since the options are already available when long pressing the video, some actions would be duplicated which is generally matter to avoid.
* It'd be some effort to implement, and I just don't think it's worth it: all the actions (download, ...) already exist, and they don't take you much more time than swipe gestures.

That being said, I appreciate your interest in improving the app.

PR #8528, opened 2026-06-12
feat: added configurable video actions on video card swipes

Gateway3B:
First time contributor. Please let me know if anything here is not correct or not allowed.

This pr allows the user to configure video actions to happen on swiping the video card to the left or right in subscriptions and in search. For your own playlists swipe left is delete and swipes on channel videos change the tab so this functionality wasn't applied to either.

My youtube viewing pattern is I go to my subscriptions, add any videos that look interesting to my watch list playlist up to the "you are up to date card". Then I watch my playlist. This feature makes my pattern much simpler by reducing this action from a long tap to open the bottom card, then tap to select add to playlist, then tap add to playlist in the selector modal, to a single swipe.

Right now the preferences are in general because I didn't know where to put them. I saw some discussion that there are too many settings that make the app feel clunkier to the user. If there is a suggestion of a better location, as well as a better name for the settings, I will do that.

There is a swipe right setting and a swipe left setting. You can select the same option for both if you want. For add to playlist, another setting appears that allows you to select manual, which will make the playlist selector modal appear, or select one of your existing playlists. A cool use case is if you want longer videos in a "podcasts" playlist and set that to swipe right, and a normal "watch list" playlist and set that to swipe left.

For me, selecting one of your existing playlists is the key feature. I can single swipe videos to add to my playlist (or add to queue). I also made it so the setting reverts to manual if you delete the selected playlist.

I added the other video options that appear in the bottom sheet when you long tap a video. I took the disable logic from there and used it in getMovementFlags to disable the ability to swipe a card when it wouldn't show in the bottom sheet.

I am very excited for this feature. Please let me know any changes I need to make so it is up to any standards or quality I may have broken.
[swipe_actions_demo_compressed.webm](https://github.com/user-attachments/assets/5d3cb9bf-6de7-4b20-9aed-c753eb089e38)

### DIFF
--- a/app/src/main/java/com/github/libretube/extensions/RecyclerViewAdapter.kt
+++ b/app/src/main/java/com/github/libretube/extensions/RecyclerViewAdapter.kt
@@ -5,50 +5,121 @@ import android.graphics.Paint
 import android.graphics.PorterDuff
 import android.graphics.PorterDuffXfermode
 import android.graphics.drawable.ColorDrawable
+import androidx.annotation.ColorInt
+import androidx.annotation.DrawableRes
 import androidx.core.content.ContextCompat
 import androidx.recyclerview.widget.ItemTouchHelper
 import androidx.recyclerview.widget.RecyclerView
 import com.github.libretube.R
 import com.github.libretube.helpers.ThemeHelper
+import kotlin.math.absoluteValue
+
+data class SwipeOptions(
+    val clamp: Boolean = false,
+    @field:DrawableRes val icon: Int? = null,
+    val iconScale: Double = 1.0,
+    @field:ColorInt val iconTintColor: Int? = com.google.android.material.R.attr.colorPrimaryFixed,
+    @field:ColorInt val backgroundColor: Int? = androidx.appcompat.R.attr.colorPrimary,
+    val getDisableSwipe: ((position: Int) -> Boolean)? = null,
+    val onSwipeListener: (position: Int) -> Unit = {}
+) {}
 
 fun RecyclerView.setOnDismissListener(onDismissedListener: (position: Int) -> Unit) {
     setActionListener(
-        allowSwipe = true,
-        onDismissedListener = onDismissedListener
+        swipeLeft = SwipeOptions(
+            icon = R.drawable.ic_delete,
+            iconTintColor = com.google.android.material.R.attr.colorOnError,
+            backgroundColor = androidx.appcompat.R.attr.colorError,
+            onSwipeListener = onDismissedListener
+        )
     )
 }
 
 fun RecyclerView.setOnDraggedListener(onDragListener: (from: Int, to: Int) -> Unit) {
     setActionListener(
-        allowDrag = true,
         onDragListener = onDragListener
     )
 }
 
 fun RecyclerView.setActionListener(
-    allowSwipe: Boolean = false,
-    allowDrag: Boolean = false,
-    onDragListener: (from: Int, to: Int) -> Unit = { _, _ -> },
-    onDismissedListener: (position: Int) -> Unit = {}
+    swipeLeft: SwipeOptions? = null,
+    swipeRight: SwipeOptions? = null,
+    onDragListener: ((from: Int, to: Int) -> Unit)? = null,
+    getDisableActions: ((position: Int) -> Boolean)? = null
 ) {
+    var hasSwiped = false
     val itemTouchCallback =
         object : ItemTouchHelper.SimpleCallback(
-            if (allowDrag) ItemTouchHelper.UP or ItemTouchHelper.DOWN else 0,
-            if (allowSwipe) ItemTouchHelper.LEFT else 0
+            if (onDragListener != null) ItemTouchHelper.UP or ItemTouchHelper.DOWN else 0,
+            when {
+                swipeLeft != null && swipeRight != null -> ItemTouchHelper.LEFT or ItemTouchHelper.RIGHT
+                swipeLeft != null -> ItemTouchHelper.LEFT
+                swipeRight != null -> ItemTouchHelper.RIGHT
+                else -> 0
+            }
         ) {
+            override fun getMovementFlags(
+                recyclerView: RecyclerView,
+                viewHolder: RecyclerView.ViewHolder
+            ): Int {
+                val position = viewHolder.absoluteAdapterPosition
+
+                if (getDisableActions?.invoke(position) == true) {
+                    return 0
+                }
+
+                var flags = super.getMovementFlags(recyclerView, viewHolder)
+                if (swipeLeft?.getDisableSwipe?.let { it(position) } == true) {
+                    flags = flags and ((ItemTouchHelper.LEFT shl 8).inv())
+                }
+                if (swipeRight?.getDisableSwipe?.let { it(position) } == true) {
+                    flags = flags and ((ItemTouchHelper.RIGHT shl 8).inv())
+                }
+
+                return flags
+            }
+
             override fun onMove(
                 recyclerView: RecyclerView,
                 viewHolder: RecyclerView.ViewHolder,
                 target: RecyclerView.ViewHolder
             ): Boolean {
-                if (!allowDrag) return false
+                if (onDragListener == null) return false
 
                 onDragListener.invoke(viewHolder.absoluteAdapterPosition, target.absoluteAdapterPosition)
                 return true
             }
 
+            override fun getSwipeThreshold(viewHolder: RecyclerView.ViewHolder): Float {
+                return when {
+                    swipeLeft != null && swipeLeft.clamp -> Float.MAX_VALUE
+                    swipeRight != null && swipeRight.clamp -> Float.MAX_VALUE
+                    else -> super.getSwipeThreshold(viewHolder)
+                }
+            }
+
+            override fun getSwipeEscapeVelocity(defaultValue: Float): Float {
+                return when {
+                    swipeLeft != null && swipeLeft.clamp -> Float.MAX_VALUE
+                    swipeRight != null && swipeRight.clamp -> Float.MAX_VALUE
+                    else -> super.getSwipeEscapeVelocity(defaultValue)
+                }
+            }
+
+            override fun getSwipeVelocityThreshold(defaultValue: Float): Float {
+                return when {
+                    swipeLeft != null && swipeLeft.clamp -> Float.MAX_VALUE
+                    swipeRight != null && swipeRight.clamp -> Float.MAX_VALUE
+                    else -> super.getSwipeVelocityThreshold(defaultValue)
+                }
+            }
+
             override fun onSwiped(viewHolder: RecyclerView.ViewHolder, direction: Int) {
-                onDismissedListener.invoke(viewHolder.absoluteAdapterPosition)
+                when (direction) {
+                    ItemTouchHelper.LEFT -> swipeLeft?.onSwipeListener?.invoke(viewHolder.absoluteAdapterPosition)
+                    ItemTouchHelper.RIGHT -> swipeRight?.onSwipeListener?.invoke(viewHolder.absoluteAdapterPosition)
+                    else -> return
+                }
             }
 
             override fun onChildDraw(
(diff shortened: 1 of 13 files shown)


Bnyro:
Hey, thanks for your PR. As a general heads-up, please first open an issue to discuss it or ask in the Matrix room whether your idea is a feature that is in scope for LibreTube.

As mentioned in #8465, we're currently trying to simplify the app's settings to make it easier to use. The problem is that additional options are not only making it more difficult for users to find what they're looking for - it also adds a lot of additional burden to the maintainers as we have to make sure that all these options stay functional, even when we are doing larger code changes.

We've already rejected https://github.com/libre-tube/LibreTube/issues/7393 and https://github.com/libre-tube/LibreTube/issues/4363 that propose similar swipe actions. I understand why this feature might be very helpful for you specifically - but I think 99% of users just don't add their videos to a playlist before viewing them :)

Currently, it's already possible to do exactly what you want by just long-pressing the video and choosing "Add to playlist" - which is admittedly two extra steps, but in my opinion that's still quick enough. 

I'm still not 100% sure about this - I definitely also don't like the current approach that playlists can be selected from within the settings. In my opinion, the "content" (i.e. playlists, videos, ...) should be clearly separated from the settings (which only configure how to interact with the "content", but don't access the content).

I think these changes might be interesting for @flexxxxer's [FlexTube](https://github.com/flexxxxer/FlexTube), which is a fork of LibreTube that focuses more on adding more such niche customizability.

After discussing with @FineFindus, we both agreed that we don't think that's a feature we are willing to maintain in LibreTube. Nevertheless thank you a lot contributing your changes here, this is very much appreciated!

app/src/main/java/com/github/libretube/helpers/SwipeActionHelper.kt
```diff
@@ -0,0 +1,247 @@
...
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.runBlocking
+
+class SwipeActionHelper(
+    private val context: Context,
+    private val childFragmentManager: FragmentManager,
+    private val iconScale: Double
+) {
+    fun getSwipeOptions(
```
Bnyro, line 38: As this is almost exactly the same code as in `VideoOptionsBottomSheet`, this should have been split into a new helper class instead of just duplicating all the code.

app/src/main/java/com/github/libretube/ui/adapters/SearchResultsAdapter.kt
```diff
@@ -91,6 +91,11 @@ class SearchResultsAdapter(
         }
     }
 
+    fun getStreamItemAt(position: Int): StreamItem? {
+        val item = getItem(position)
+        return if (item?.type == StreamItem.TYPE_STREAM) item.toStreamItem() else null
```
Bnyro, line 96:
```suggestion
        return getItem(position)?.takeIf { it.type == StreamItem.TYPE_STREAM }?.toStreamItem()
```

app/src/main/java/com/github/libretube/ui/preferences/GeneralSettings.kt
```diff
@@ -84,6 +87,79 @@ class GeneralSettings : BasePreferenceFragment() {
...
+            }
+        }
+
+        togglePlaylist(swipeLeft?.value, swipeLeftPlaylist)
+        togglePlaylist(swipeRight?.value, swipeRightPlaylist)
+
+        swipeLeft?.setOnPreferenceChangeListener { _, newValue ->
+            if (newValue is String) {
+                togglePlaylist(newValue, swipeLeftPlaylist)
+            }
```
Bnyro, line 133:
```suggestion
            togglePlaylist(newValue as String, swipeLeftPlaylist)
```
The value is guaranteed to be a string.

app/src/main/java/com/github/libretube/ui/preferences/GeneralSettings.kt
```diff
@@ -84,6 +87,79 @@ class GeneralSettings : BasePreferenceFragment() {
...
+            val values = preference.entryValues
+            val index = values?.indexOf(newValue)
+            if (index != null && index >= 0 && entries != null && index < entries.count()) {
+                preference.summary = entries[index]
+            }
+        }
+        swipeLeftPlaylist?.setOnPreferenceChangeListener { preference, newValue ->
+            if (preference is ListPreference && newValue is String) {
+                setSummary(preference, newValue)
+            }
```
Bnyro, line 154:
```suggestion
            setSummary(preference as ListPreference, newValue as String)
```

app/src/main/java/com/github/libretube/ui/preferences/GeneralSettings.kt
```diff
@@ -84,6 +87,79 @@ class GeneralSettings : BasePreferenceFragment() {
...
+        val swipeLeftPlaylist = findPreference<ListPreference>(PreferenceKeys.SWIPE_LEFT_PLAYLIST)
+
+        val swipeRight = findPreference<ListPreference>(PreferenceKeys.SWIPE_RIGHT)
+        val swipeRightPlaylist = findPreference<ListPreference>(PreferenceKeys.SWIPE_RIGHT_PLAYLIST)
+
+        val togglePlaylist = { newValue: String?, playlistPreference: ListPreference? ->
+            if (newValue == "addToPlaylist") {
+                lifecycleScope.launch {
+                    kotlin.runCatching {
+                        PlaylistsHelper.getPlaylists()
```
Bnyro, line 101: These should probably be cached within `GeneralSettings` or a `ViewModel` and not reloaded each time the user selects the `addToPlaylist`option

app/src/main/java/com/github/libretube/ui/preferences/GeneralSettings.kt
```diff
@@ -84,6 +87,79 @@ class GeneralSettings : BasePreferenceFragment() {
...
+                        if (!filteredPlaylists.map { it.id }.contains(playlistPreference?.value)) {
+                            playlistPreference?.value = "manual"
+                            playlistPreference?.summary = resources.getString((R.string.manual))
+                        } else {
+                            playlistPreference?.value = currentValue
+                            playlistPreference?.summary = playlistPreference.entry
+                        }
+                    }.onFailure {
+                        context?.toastFromMainThread(R.string.error)
+                    }
```
Bnyro, line 120: `runCatching` doesn't really make sense here because we could achieve the same in simpler by just using `try catch`

app/src/main/java/com/github/libretube/ui/fragments/SubscriptionsFragment.kt
```diff
@@ -186,6 +188,19 @@ class SubscriptionsFragment : DynamicLayoutManagerFragment(R.layout.fragment_sub
...
+        val swipeRight = PreferenceHelper.getString(PreferenceKeys.SWIPE_RIGHT, "none")
+        val swipeLeftPlaylist = PreferenceHelper.getString(PreferenceKeys.SWIPE_LEFT_PLAYLIST, "manual")
+        val swipeRightPlaylist = PreferenceHelper.getString(PreferenceKeys.SWIPE_RIGHT_PLAYLIST, "manual")
+        val swipeActionHelper = SwipeActionHelper(requireContext(), childFragmentManager, 2.0)
+        viewModel.videoFeed.value?.also {
+            val getVideoAt = { position: Int -> it[position] }
+            binding.subFeed.setActionListener(
+                swipeLeft = swipeActionHelper.getSwipeOptions(swipeLeft, swipeLeftPlaylist,  getVideoAt),
+                swipeRight = swipeActionHelper.getSwipeOptions(swipeRight, swipeRightPlaylist, getVideoAt)
+            )
```
Bnyro, line 201: Why don't you just pass the feed down here? As Java/Kotlin only passes references, this is actually less overhead than passing a callback.
```suggestion
        viewModel.videoFeed.value?.also { feed ->
            binding.subFeed.setActionListener(
                swipeLeft = swipeActionHelper.getSwipeOptions(swipeLeft, swipeLeftPlaylist,  feed),
                swipeRight = swipeActionHelper.getSwipeOptions(swipeRight, swipeRightPlaylist, feed)
            )
```

app/src/main/java/com/github/libretube/ui/fragments/SearchResultFragment.kt
```diff
@@ -115,6 +119,28 @@ class SearchResultFragment : DynamicLayoutManagerFragment(R.layout.fragment_sear
...
+            viewModel.searchResultsFlow.collect { pagingData ->
+                searchResultsAdapter.submitData(pagingData)
+
+                val getCurrentItem = { position: Int -> searchResultsAdapter.getStreamItemAt(position) }
+                val getDisableActions = { position: Int -> searchResultsAdapter.getItemViewType(position) != 0 }
+
+                val swipeLeft = PreferenceHelper.getString(PreferenceKeys.SWIPE_LEFT, "none")
+                val swipeRight = PreferenceHelper.getString(PreferenceKeys.SWIPE_RIGHT, "none")
+                val swipeLeftPlaylist = PreferenceHelper.getString(PreferenceKeys.SWIPE_LEFT_PLAYLIST, "manual")
+                val swipeRightPlaylist = PreferenceHelper.getString(PreferenceKeys.SWIPE_RIGHT_PLAYLIST, "manual")
```
Bnyro, line 132: Might have been worth putting `"none", "manual", ...` into constants to make it more readable

app/src/main/java/com/github/libretube/ui/fragments/SearchResultFragment.kt
```diff
@@ -115,6 +119,28 @@ class SearchResultFragment : DynamicLayoutManagerFragment(R.layout.fragment_sear
...
+                val swipeActionHelper = SwipeActionHelper(requireContext(), childFragmentManager, 1.0)
+                viewModel.searchResultsFlow.also {
```
Bnyro, line 134: The use of `also` here doesn't make any sense?

app/src/main/java/com/github/libretube/helpers/SwipeActionHelper.kt
```diff
@@ -0,0 +1,247 @@
...
+    private fun isCurrentVideoPlaying(video: StreamItem): Boolean =
+        PlayingQueue.getCurrent()?.url?.toID() == video.url?.toID()
+
+    private fun isQueueActive(): Boolean =
+        PlayingQueue.isNotEmpty() && PlayingQueue.queueMode == PlayingQueueMode.ONLINE
+
+    private fun hasWatchPositionEntry(video: StreamItem): Boolean {
+        val videoId = video.url?.toID() ?: return false
+
+        val watchHistoryEntry = runBlocking(Dispatchers.IO) {
```
Bnyro, line 241: `runBlocking` is usually a very bad idea because it can block the UI thread (i.e. freeze the screen). We're working on getting rid of all `runBlocking` uses, but it admittedly requires some refactoring to the code

app/src/main/java/com/github/libretube/extensions/RecyclerViewAdapter.kt
```diff
@@ -5,50 +5,121 @@ import android.graphics.Paint
...
+                if (getDisableActions?.invoke(position) == true) {
+                    return 0
+                }
+
+                var flags = super.getMovementFlags(recyclerView, viewHolder)
+                if (swipeLeft?.getDisableSwipe?.let { it(position) } == true) {
+                    flags = flags and ((ItemTouchHelper.LEFT shl 8).inv())
+                }
+                if (swipeRight?.getDisableSwipe?.let { it(position) } == true) {
+                    flags = flags and ((ItemTouchHelper.RIGHT shl 8).inv())
```
Bnyro, line 76: Please generally document magic operations (like these shifts) with comments.

app/src/main/java/com/github/libretube/extensions/RecyclerViewAdapter.kt
```diff
@@ -68,34 +139,103 @@ fun RecyclerView.setActionListener(
...
+                            ItemTouchHelper.RIGHT -> swipeRight?.onSwipeListener?.invoke(viewHolder.absoluteAdapterPosition)
+                        }
+                    }
+
+                    val swipeClamp = itemView.width * 0.4f
+                    val maxOvershoot = itemView.width * 0.1f
+                    val clamped = dX.coerceIn(-swipeClamp, swipeClamp)
+                    val overshoot = dX - clamped
+                    val overshootFactor = (overshoot.absoluteValue / maxOvershoot).coerceIn(0f, 1f) * 0.2f
+                    clamped + (overshoot * overshootFactor).coerceIn(-maxOvershoot, maxOvershoot)
```
Bnyro, line 179: Such math operations should also be documented, it's not obvious how this works.

app/src/main/java/com/github/libretube/extensions/RecyclerViewAdapter.kt
```diff
@@ -68,34 +139,103 @@ fun RecyclerView.setActionListener(
...
+                    else -> return
                 }
-                background.draw(c)
 
-                val deleteIcon = ContextCompat.getDrawable(context, R.drawable.ic_delete)!!.apply {
-                    setTint(onBackgroundColor)
+                val itemViewEndWithOffset = itemViewEnd + dX.toInt()
+
+                if (swipe.backgroundColor != null) {
+                    // Draw the background
```
Bnyro, line 191: That comment e.g. is useless. `background.draw` already makes it pretty clear that we draw the background.

app/src/main/java/com/github/libretube/extensions/RecyclerViewAdapter.kt
```diff
@@ -68,34 +139,103 @@ fun RecyclerView.setActionListener(
...
+                    val icon = ContextCompat.getDrawable(context, swipe.icon)!!.apply {
+                        if (swipe.iconTintColor != null) {
+                            setTint(ThemeHelper.getThemeColor(context, swipe.iconTintColor))
+                        }
+                    }
+
+                    val intrinsicHeight = (icon.intrinsicHeight * swipe.iconScale).toInt()
+                    val intrinsicWidth = (icon.intrinsicWidth * swipe.iconScale).toInt()
+
+                    // Calculate position of the icon
```
Bnyro, line 213: Yes, but how? A comment would have been nice.

Bnyro:
I've left a superficial review of the changes that might help you to improve your code, e.g. if you want to contribute something else to LibreTube in the future :)

As explained in my previous message, we're closing this PR. Anyways, thank you!

Gateway3B:
Thank you for the review. I understand this isn't the right feature for LibreTube and I apologise for not looking close enough at the issues or making my own. As this is the most important feature for a YouTube client for me (personally lol) besides being able to play videos, I'll either maintain this myself on my own fork or see if Flextube is interested. Thank you again and I hope to contribute to the project in the future.

PR #8535, opened 2026-06-13
feat(downloads): add support for downloading SABR streams

### LINKED ISSUE #8527: Downloads broken in debug builds
### Steps to reproduce

Download a video.

### Expected behavior

It shouldn't crash the app.

### Actual behavior

Crash, whose log is irrelevant here because it's not related to the actual problem that we don't support SABR downloads yet.

### LibreTube version

debug

### Android version

*

### Other details

This is due to #7689, where we forgot that we also need to add SABR support to the downloads :)

### Acknowledgements

Bnyro:
(the diff only looks so big due to the database migrations)

This also fixes some other small download bugs and cleans the code up a bit.

Still a draft because not tested much yet (only short video streams tested so far, without pause and resume).

Closes: https://github.com/libre-tube/LibreTube/issues/8527


app/src/main/java/com/github/libretube/player/parser/SabrClient.kt
```diff
@@ -346,7 +358,7 @@ class SabrClient private constructor(
      */
     private suspend fun fetchStreamData(
         playbackRequest: PlaybackRequest,
-        audioFormat: Representation,
+        audioFormat: Representation?,
```
FineFindus, line 349: We should be fine using `!!` here, since requesting media data without any audio stream is invalid anyway. So I would rather have an explicit null error earlier, than an opaque/non-descript SABR error later.

app/src/main/java/com/github/libretube/repo/RawByteStreamDownloadProvider.kt
```diff
@@ -0,0 +1,110 @@
...
+
+    private suspend fun startConnection(
+        url: HttpUrl,
+        alreadyRead: Long,
+        readLimit: Long?
+    ): ResponseBody? {
+        val limit = readLimit?.let {
+            // generate a random byte distance to make it more difficult to fingerprint
+            val nextBytesToReadSize = (BYTES_PER_REQUEST_MIN..BYTES_PER_REQUEST_MAX).random()
+            min(readLimit, alreadyRead + nextBytesToReadSize)
```
FineFindus: Why not imitate the download size of an official client? Not that I think there is any fingerprinting going on for streaming urls.

app/src/main/java/com/github/libretube/repo/SabrDownloadProvider.kt
```diff
@@ -0,0 +1,92 @@
...
+        downloadItem: DownloadItem,
+        streams: Streams,
+        stream: PipedStream
+    ): SabrDownloaderHandle {
+        val manifest = SabrManifest(downloadItem.videoId, streams)
+        val sabrClient = SabrClient(manifest)
+        val streamRepresentation = manifest.adaptationSets.flatMap { it.representations }
+            .filterNotNull()
+            .first { it.stream == stream }
+        sabrClient.selectFormat(streamRepresentation)
```
FineFindus: Hmm, maybe it's better to move `Representation` construction to a constructor that also builds the format. I didn't want to do this originally, since it's a bit messy with audio/video, but would skip creating a manifest …

app/src/main/java/com/github/libretube/player/parser/SabrClient.kt
```diff
@@ -281,6 +281,18 @@ class SabrClient private constructor(
         }
     }
 
+    /**
+     * Get the number of the last segment number.
+     *
+     * Calling [getNextSegment] with [PlaybackRequest.segment] set to a value larger than the last
+     * segment number will crash the client.
+     */
+    fun getEndSegmentNumber(formatId: FormatId): Long? = initializedFormats[formatId.itag]?.endSegmentNumber
```
FineFindus, line 278:
```suggestion
    /**
     * Get the number of the last segment.
     *
     * Calling [getNextSegment] with [PlaybackRequest.segment] set to a value larger than the last
     * segment number will crash the client.
     */
    fun getEndSegmentNumber(formatId: FormatId): Long? = initializedFormats[formatId.itag]?.endSegmentNumber
```

app/src/main/java/com/github/libretube/player/parser/SabrClient.kt
```diff
@@ -175,7 +175,7 @@ class SabrClient private constructor(
     private val dispatcher = Dispatchers.IO.limitedParallelism(1)
 
     /** Audio format video format */
-    private lateinit var audioFormat: Representation
+    private var audioFormat: Representation? = null
```
FineFindus, line 166: I don't quite get why this is nullable now?

FineFindus (changes requested):

app/src/main/java/com/github/libretube/player/parser/SabrClient.kt
Bnyro: Why is this invalid? I managed to download a video stream whilst having `audioStream` set to `null`.

app/src/main/java/com/github/libretube/player/parser/SabrClient.kt
Bnyro: Otherwise the `SabrClient` crashed because it wasn't initialized.

app/src/main/java/com/github/libretube/repo/RawByteStreamDownloadProvider.kt
Bnyro: Well, now we only need to know what the download size of the official client is? :)

I don't have YouTube premium to test it.

FineFindus: Neither do I. [NewPipe](https://github.com/TeamNewPipe/NewPipe/blob/ebcc72ac8dc839db32be1c8f25a9503f2df7b9a6/app/src/main/java/us/shandian/giga/get/DownloadMission.java#L39) uses `512 * 1024`, so we could also go with that.

Bnyro: I think there was some reason why we switched away from that - but I really don't know anymore.

app/src/main/java/com/github/libretube/player/parser/SabrClient.kt
FineFindus: If no audio is selected the server responds with a `sabr.no_audio_selected` error. Video-only download where supported, but have been deprecated for a while now. There are several possibilities why it still work (e.g. server bug, server still accepting it, reversed the deprecation, etc.), but I'm not sure if we should count on those. AFAIK what yt-dlp and googlevideos/kira does is just ignoring the video stream when downloading.

Bnyro:
Fyi, I refactored some stuff to get rid of the generics - turns out that this was just a lot of unnecessary complexity that can be simplified by refactoring the download service code.

app/src/main/java/com/github/libretube/player/parser/SabrClient.kt
Bnyro: So we basically have to set a dummy audio stream (e.g. just choose a random one) when downloading a video and can leave everything else unchanged?

FineFindus: Preferably we use either the smallest stream or one the user selected for download anyway. If it works without, it could be worthwhile to investigate why, it's possible that they reversed their decision or it's a workaround.

I'm also wondering what's the use case for downloading a video without audio? If a user wanted to watch a video without audio, they can still mute the audio, no?

Bnyro: The downloader currently downloads audio and video separately from each other, so it doesn't really matter if the user decided video and audio or only video.

But yes, technically it doesn't make much sense to download no audio. But I don't either see why we should make the download dialog more complicated and force people to select an audio track.

Bnyro: I still haven't done anything to always set an audio track as we discussed above - In my opinion, we can postpone that to some time to the future when it's actually needed, but if you disagree please feel free to push a commit on top of mine to change that.

app/src/main/java/com/github/libretube/api/obj/Streams.kt
```diff
@@ -75,6 +75,7 @@ data class Streams(
         }
 
         if (!subCode.isNullOrEmpty()) {
+            // TODO: doesn't support track types, i.e. if there are two caption tracks for the same language
             val subtitle = subtitles.find { it.code == subCode }
```
FineFindus, line 79: Can't we simply use more metadata to identify the subtitle? E.g. full language tag + track type?

app/src/main/java/com/github/libretube/player/parser/SabrClient.kt
```diff
@@ -269,6 +286,18 @@ class SabrClient private constructor(
...
+     * Calling [getNextSegment] with [PlaybackRequest.segment] set to a value larger than the last
+     * segment number will crash the client.
+     */
+    fun getEndSegmentNumber(formatId: FormatId): Long? = initializedFormats[formatId.itag]?.endSegmentNumber
+
+    /**
+     * Download the stream chunk specified in [playbackRequest].
+     * Segments are usually of a length between 2 and 10 seconds.
+     */
     fun getNextSegment(playbackRequest: PlaybackRequest): Segment? {
```
FineFindus, line 284:
```suggestion
    /**
     * Returns the segment specified in [playbackRequest].
     * Segments are usually of a length between 2 and 10 seconds.
     */
fun getNextSegment(playbackRequest: PlaybackRequest): Segment? {
```

This may not always download segments, e.g. a segment may already be buffered.

app/src/main/java/com/github/libretube/player/parser/SabrClient.kt
```diff
@@ -173,6 +186,10 @@ class SabrClient private constructor(
         ByteString.copyFrom(manifest.videoPlaybackUstreamerConfig)
     )
 
+    constructor(videoId: String, url: String, ustreamerConfig: String) : this(
+        videoId, url, ByteString.copyFrom(Base64.decode(ustreamerConfig, Base64.URL_SAFE))
+    )
```
FineFindus: I don't particularly like that we know have two constructors, with one being nearly identical to the private constructors (and sharing the base64 decoding responsibility with the manifest). I would prefer it if we constructed a new manifest instead, to also avoid having two diverging code paths.

app/src/main/java/com/github/libretube/repo/RawByteStreamDownloadProvider.kt
```diff
@@ -0,0 +1,107 @@
...
+
+        return withContext(Dispatchers.IO) {
+            // Retry connecting to server for n times.
+            try {
+                val call = httpClient.newCall(request)
+                val response = call.execute()
+
+                if (response.code == 403) {
+                    response.close()
+                    return@withContext null
```
FineFindus, line 89: We should log this, to help us when debugger issues in the future (e.g. if subtitle URLs require PoTokens)

FineFindus (dismissed):

app/src/main/java/com/github/libretube/api/obj/Streams.kt
Bnyro: Yes, but that's unrelated to this PR. We currently don't have database columns for that

PR #8536, opened 2026-06-13
perf(sabr): avoid copying segment data

FineFindus:
Uses a `CompositeBuffer`, which wraps multiple smaller contiguous buffers and allows reading from them, as if they were a single buffer. This avoids copying the segment data into a single array, before reading from it, improving performance and reducing memory pressure.

This was originally part of https://github.com/libre-tube/LibreTube/pull/7689, however it must have been some in a rebase at some point.

### DIFF
--- a/app/src/main/java/com/github/libretube/player/parser/CompositeBuffer.kt
+++ b/app/src/main/java/com/github/libretube/player/parser/CompositeBuffer.kt
@@ -1,0 +1,64 @@
+package com.github.libretube.player.parser
+
+import kotlin.math.min
+
+/**
+ * A buffer that wraps multiple smaller contiguous buffers,
+ * allowing to treat them, as if they were a single buffer.
+ */
+class CompositeBuffer(
+    private val data: List<ByteArray>,
+) {
+    private var position: Int = 0
+    private val length = data.sumOf { it.size }
+
+    /**
+     * Returns the number of [Byte]s remaining in the buffer.
+     */
+    fun remaining() = length - position
+
+    /**
+     * Returns whether the buffer still has [Byte]s to read.
+     */
+    fun hasRemaining() = position != length
+
+    /**
+     * Transfers [length] bytes into [destination] starting at [offset].
+     */
+    @Throws(IndexOutOfBoundsException::class)
+    fun read(destination: ByteArray, offset: Int, length: Int) {
+        if (destination.size < offset + length) {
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/github/libretube/player/parser/SabrClient.kt
```diff
@@ -81,19 +81,6 @@ data class Segment(
...
-     */
-    fun data(): ByteArray {
-        val result = ByteArray(length())
-        var offset = 0
-        for (chunk in data) {
-            System.arraycopy(chunk, 0, result, offset, chunk.size)
-            offset += chunk.size
-        }
-        return result
-    }
```
Bnyro, line 96: For #8535 , does this mean I would need to access the `data` attribute and join the byte arrays in it together by hand?

app/src/main/java/com/github/libretube/player/parser/CompositeBuffer.kt
```diff
@@ -0,0 +1,61 @@
...
+    /**
+     * Returns whether the buffer still has [Byte]s to read.
+     */
+    fun hasRemaining() = position != length
+
+    /**
+     * Transfers [length] bytes into [destination] starting at [offset].
+     */
+    fun read(destination: ByteArray, offset: Int, length: Int) {
+        assert(destination.size >= offset + length) { "Destination buffer does not have enough space to hold ${offset + length} bytes" }
```
Bnyro: Wouldn't it be better to throw an exception here so that it can gracefully be handled by the app? Otherwise we should probably add a warning that it will crash when you don't make sure that `length < remaining()` :)

app/src/test/java/com/github/libretube/player/parser/CompositeBufferTest.kt
Bnyro, line 1: If you want full test coverage, a test for the overflow assertion in `read` would also be nice :)

app/src/main/java/com/github/libretube/player/parser/SabrClient.kt
FineFindus: No, you can simply iterate over the chunks and directly write them. The joining/`CompositeBuffer` is only necessary if you need to treat the data as a single buffer.
```kt
for chunk in segment.data {
    sink.write(bytes)
}
```

Bnyro (approved):

PR #8537, opened 2026-06-14
refactor: merge `EditChannelGroupsModel` into `SubscriptionsViewModel`

dhanuarf:
Merging these view models since they are both subscriptions-related and both lifecycle are tied to the activity (shared view models). And theres not much data stored in each, so merging these two will make the code look cleaner, hopefully :).


Bnyro (approved):
True, thanks!

PR #8543, opened 2026-06-19
Fix : shows 1 week ago instead of 0 weeks ago for 7 days old items (#8295)

### LINKED ISSUE #8295: Inaccurate relative time calculation
### Steps to reproduce

I don't exactly know how to reproduce this and currently can't reproduce it myself, but here is my best guess: 
1. View list of uploaded videos of a channel
2. Find a video that was uploaded **exactly** 7 days ago (not an hour or so more)

### Expected behavior

The relative upload time display prints `1 week ago`.

### Actual behavior

It prints `0 weeks` ago.

Image

### LibreTube version

0.30.0 (65)

### Android version

16

### Other details

This bug occurred to me only once. It was displayed right in between videos that were uploaded `6 days ago` and `1 week ago`. That's why I guess this bug only occurs for videos that were uploaded **exactly** 7 days ago.
After I refreshed the site a few minutes later, it was displayed correctly - probably because some time had passed.

### Acknowledgements

Phoenix1808:
Fixes #8295 

## Reason : In else block of formatRelativeDate(), weeks were being calculated in bit of contradictory method :
1. `date.until(now, ChronoUnit.WEEKS)` => this counted the weeks exactly according to the calendar algorithm i.e returns 1 week on 7 days.
2. Previously we used to call `DateUtils.getRelativeTimeSpanString(....)` along with `minResolution = WEEK_IN_MILLIS` , Here minResolution was a flag which asks DateUtils to shows show "weeks in unit" but DateUtils counts milliseconds inside itself seperately due to which if time was bit less than 7 then it returns count to be 0 i.e "0 weeks ago".

In short : our count(1) and inner count(0) of DateUtils clashes hence contradiction occurs.

## FIX: 
Before :
``val weeks = date.until(now, ChronoUnit.WEEKS)``
``val minResolution = if (weeks > 0) DateUtils.WEEK_IN_MILLIS else 0L``
``DateUtils.getRelativeTimeSpanString(unixTime, System.currentTimeMillis(), minResolution)``

After :
``val weeks = date.until(now, ChronoUnit.WEEKS)``
``if (weeks > 0) {``
    ``RelativeDateTimeFormatter.getInstance()``
      ``  .format(weeks.toDouble(), RelativeDateTimeFormatter.Direction.LAST, RelativeDateTimeFormatter.RelativeUnit.WEEKS)``
``} else {``
    ``DateUtils.getRelativeTimeSpanString(unixTime, System.currentTimeMillis(), 0L)``
``}``

## Changes :

1. We Removed the concept of ``minResolution``` where ``DateUtils``  used to handle  week operations.
2. Now when ``weeks>=1``, the formatting is done with the ``RelativeDateTimeFormatter`` , through this that "0 weeks" thing becomes impossible.
3. When ``weeks<1`` then ``DateUtils`` now shows ``DateUtils( 0L )`` like this shows the proper format in days.

Verified by compiling the project successfully.

### DIFF
--- a/app/src/main/java/com/github/libretube/util/TextUtils.kt
+++ b/app/src/main/java/com/github/libretube/util/TextUtils.kt
@@ -123,8 +123,13 @@ object TextUtils {
                 .format(time.toDouble(), RelativeDateTimeFormatter.Direction.LAST, timeFormat)
         } else {
             val weeks = date.until(now, ChronoUnit.WEEKS)
-            val minResolution = if (weeks > 0) DateUtils.WEEK_IN_MILLIS else 0L
-            DateUtils.getRelativeTimeSpanString(unixTime, System.currentTimeMillis(), minResolution)
+            
+            if(weeks > 0){
+                RelativeDateTimeFormatter.getInstance()
+                  .format(weeks.toDouble(), RelativeDateTimeFormatter.Direction.LAST, RelativeDateTimeFormatter.RelativeUnit.WEEKS)
+            } else {
+                DateUtils.getRelativeTimeSpanString(unixTime, System.currentTimeMillis(), 0L)
+            }
         }
     }
 
--- a/app/src/main/java/com/github/libretube/util/TextUtils.kt
+++ b/app/src/main/java/com/github/libretube/util/TextUtils.kt
@@ -107,30 +107,42 @@ object TextUtils {
     }
 
     fun formatRelativeDate(unixTime: Long): CharSequence {
-        val date = LocalDateTime.ofInstant(Instant.ofEpochMilli(unixTime), ZoneId.systemDefault())
+        
         val now = LocalDateTime.now()
-        val months = date.until(now, ChronoUnit.MONTHS)
+        val relativeTime = getRelativeTimeUnit(unixTime, now)
 
-        return if (months > 0) {
-            val years = months / 12
-
-            val (timeFormat, time) = if (years > 0) {
-                RelativeDateTimeFormatter.RelativeUnit.YEARS to years
-            } else {
-                RelativeDateTimeFormatter.RelativeUnit.MONTHS to months
+        return if (relativeTime != null) {
+            val (unit,amount) = relativeTime
+            val formatUnit = when (unit){
+                ChronoUnit.YEARS -> RelativeDateTimeFormatter.RelativeUnit.YEARS
+                ChronoUnit.MONTHS -> RelativeDateTimeFormatter.RelativeUnit.MONTHS
+                else -> RelativeDateTimeFormatter.RelativeUnit.WEEKS
             }
+           
             RelativeDateTimeFormatter.getInstance()
-                .format(time.toDouble(), RelativeDateTimeFormatter.Direction.LAST, timeFormat)
+                .format(amount.toDouble(), RelativeDateTimeFormatter.Direction.LAST, formatUnit)
         } else {
-            val weeks = date.until(now, ChronoUnit.WEEKS)
-            
-            if(weeks > 0){
-                RelativeDateTimeFormatter.getInstance()
-                  .format(weeks.toDouble(), RelativeDateTimeFormatter.Direction.LAST, RelativeDateTimeFormatter.RelativeUnit.WEEKS)
-            } else {
                 DateUtils.getRelativeTimeSpanString(unixTime, System.currentTimeMillis(), 0L)
             }
-        }
+    }
+    
+
+    internal fun getRelativeTimeUnit(
+        unixTime:Long, now: LocalDateTime, zone: ZoneId = ZoneId.systemDefault()
+    ): Pair<ChronoUnit, Long>? {
+
+     val date= LocalDateTime.ofInstant(Instant.ofEpochMilli(unixTime), zone)
+     val months = date.until(now, ChronoUnit.MONTHS)
+     if(months>0){
+        val years = months/12
+        return if (years>0) ChronoUnit.YEARS to  years 
+        else ChronoUnit.MONTHS to months
+     }
+
+     val weeks = date.until(now, ChronoUnit.WEEKS)
+     if(weeks>0) return ChronoUnit.WEEKS to weeks
+     return null
+
     }
 
     fun formatBitrate(bitrate: Int?): String {
--- a/app/src/test/java/com/github/libretube/RelativeDateTest.kt
+++ b/app/src/test/java/com/github/libretube/RelativeDateTest.kt
@@ -1,0 +1,20 @@
+package com.github.libretube
+
+import com.github.libretube.util.TextUtils
+import java.time.LocalDateTime
+import java.time.ZoneId
+import java.time.ZonedDateTime
+import java.time.temporal.ChronoUnit
+import org.junit.Assert.assertEquals
+import org.junit.Test
+
+class RelativeDateTest {
+    @Test
+    fun reportOneWeek() {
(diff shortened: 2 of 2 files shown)


Bnyro:
> Verified by compiling the project successfully.
>
So you didn't actually test if dates get formatted correctly?

Phoenix1808:
> > Verified by compiling the project successfully.
> 
> So you didn't actually test if dates get formatted correctly?

I actually tested it on my physical device. For a timestamp exactly 7 days old it correctly shows '1 week ago'
Here's the Proof :

image

---

image

Bnyro:
But do you also have a test case, where without this change, it would show 0 weeks?

Bnyro:
Also, the screenshot doesn't "proof" anything - even before this change, there were videos that have been shown as "1 week ago"

Phoenix1808:
> But do you also have a test case, where without this change, it would show 0 weeks?



> Also, the screenshot doesn't "proof" anything - even before this change, there were videos that have been shown as "1 week ago"

You're right — the screenshot doesn't prove it, since without DST the old code also shows 1 week ago.
I went into it  this is a DST edge case. When an item is exactly 7 days old and a  DST transition falls in that window. ChronoUnit.WEEKS returns 1 (so the old code forced the weeks unit via minResolution), but DateUtils counts milliseconds and gets 0 weeks ago .
 My fix avoids this by formatting weeks directly from the calendar count.
My timezone has IST, so I can't reproduce it locally with real content — that's why my screenshot didn't show a difference. I can set my device to a DST timezone around a transition to demonstrate, or add a unit test for the date logic — whichever you prefer.

Bnyro:
> I can set my device to a DST timezone around a transition to demonstrate, or add a unit test for the date logic 
>
In this case, I think it actually makes sense to create one single unit test for verifying that the date gets formatted properly when coming across a DST timezone change. But it doesn't make sense to create countless unit tests for testing all possible date formatting cases, like in your previous PR, because that would only test functionality of the `DateUtils` library, which doesn't make sense to do in LibreTube.

Phoenix1808:
> > I can set my device to a DST timezone around a transition to demonstrate, or add a unit test for the date logic
> 
> In this case, I think it actually makes sense to create one single unit test for verifying that the date gets formatted properly when coming across a DST timezone change. But it doesn't make sense to create countless unit tests for testing all possible date formatting cases, like in your previous PR, because that would only test functionality of the `DateUtils` library, which doesn't make sense to do in LibreTube.

Okk..So I'll add single test for DST case. There will be two small things it will need - `1. formatRelativeDate` uses LocalDateTime.now() so will add optional now parameter. So that test can pin fixed moment DST edge. 2. I plan to add RoboElectric for this as it uses RelativeDateTimeFormatter unless you'd prefer instrumented test. 

Does this approach works for you ??

Bnyro:
Please just add this as a normal JUnit test.

Phoenix1808:
> Please just add this as a normal JUnit test.

I added a plain JUnit test for the DST case. To make it testable I extracted the date unit decision into a small internal function getRelativeTimeUnit — formatRelativeDate itself can't be unit tested directly since it uses RelativeDateTimeFormatter/DateUtils, which don't run in a plain unit test.

The test verifies the DST case: across a spring-forward, an item that is 7 calendar days old actually has less than 7×24h of real time elapsed. So the old code's DateUtils millisecond count disagreed with the calendar week count and showed "0 weeks ago". The new code uses the calendar week count, so it correctly reports 1 week.

FineFindus:
Thanks, I decided to fix this in https://github.com/libre-tube/LibreTube/pull/8549 and also simplify the code a bit.

Phoenix1808:
> Thanks, I decided to fix this in #8549 and also simplify the code a bit.

Can you explain why not this ? I geniunely wanted to contribute in it !

FineFindus:
> Can you explain why not this ? I geniunely wanted to contribute in it !

Sure, while your solution isn't incorrect, it's overly complicated, compared to what is required to fix the bug. It also allowed me to refactor and simplify the code, which is why I opened a PR, rather than reviewing this one.

> I geniunely wanted to contribute in it !

Don't worry, there are still plenty of issues waiting to be solved :)

Phoenix1808:
> Sure, while your solution isn't correct, it's overly complicated, compared to what is required to fix the bug. It also allowed me to refactor and simplify the code, which is why I opened a PR, rather than reviewing this one.
> 

Can you please review how was this over complicated I wrote the JUnit test as asked and how could I have improve it ?

app/src/main/java/com/github/libretube/util/TextUtils.kt
```diff
@@ -107,25 +107,42 @@ object TextUtils {
...
-        val months = date.until(now, ChronoUnit.MONTHS)
-
-        return if (months > 0) {
-            val years = months / 12
-
-            val (timeFormat, time) = if (years > 0) {
-                RelativeDateTimeFormatter.RelativeUnit.YEARS to years
-            } else {
-                RelativeDateTimeFormatter.RelativeUnit.MONTHS to months
+        val relativeTime = getRelativeTimeUnit(unixTime, now)
```
FineFindus, line 112:
```suggestion
        val (unit, amount) = getRelativeTimeUnit(unixTime, now) ?: DateUtils.getRelativeTimeSpanString(unixTime, System.currentTimeMillis(), 0L)
```

app/src/main/java/com/github/libretube/util/TextUtils.kt
```diff
@@ -107,25 +107,42 @@ object TextUtils {
...
-            val minResolution = if (weeks > 0) DateUtils.WEEK_IN_MILLIS else 0L
-            DateUtils.getRelativeTimeSpanString(unixTime, System.currentTimeMillis(), minResolution)
-        }
+                DateUtils.getRelativeTimeSpanString(unixTime, System.currentTimeMillis(), 0L)
+            }
+    }
+    
+
+    internal fun getRelativeTimeUnit(
+        unixTime:Long, now: LocalDateTime, zone: ZoneId = ZoneId.systemDefault()
```
FineFindus, line 131:
```suggestion
        unixTime:Long, now: LocalDateTime
```
Having this `zone` parameter only for testing isn't great, since it might be accidentally be set and lead to bugs.

FineFindus, line 131: There isn't really a reason for this function to exist, it should be done inline in `formatRelativeDate`.

app/src/test/java/com/github/libretube/RelativeDateTest.kt
```diff
@@ -0,0 +1,20 @@
...
+class RelativeDateTest {
+    @Test
+    fun reportOneWeek() {
+        val zone = ZoneId.of("America/New_York")
+        val now = LocalDateTime.of(2025, 3, 12, 12, 0)
+        val sevenago = ZonedDateTime.of(2025, 3, 5, 12, 0, 0, 0, zone).toInstant().toEpochMilli()
+
+        assertEquals(ChronoUnit.WEEKS to 1L, TextUtils.getRelativeTimeUnit(sevenago, now, zone))
+    }
+}
```
FineFindus, line 20: This tests that the `date.until` function returns the correct time (which we already knew), rather than testing that `formatRelativeDate` works for weeks with DST timestamps.

FineFindus (commented):
I left some basic review comments, you can take a look at https://github.com/libre-tube/LibreTube/pull/8549, to see how it could be done simpler.

PR #8544, opened 2026-06-20
feat: Added coroutine worker for importing watch history

SE-Bilal-Ahmad:
Currently working on resolving the previous comments

1- Previously a coroutine was executed which imported watch history from youtube but there was no way to know the status of the import history i have a added a coroutine worker whose job is to import the watch history with a foreground notification.

2- I have also added dagger hilt in order to inject dependencies. I could have used other approaches but that wouldnt have been a clean solution

3- For now only watch history is being imported through coroutine worker. The reason is that we are inserting watch item one by one but for subscriptions one complete list is inserted which makes it nearly impossible to know the status of those imports however i will still look for a way to achieve that and for playlists too

Bug https://github.com/libre-tube/LibreTube/issues/7907


SE-Bilal-Ahmad:
All comments from the previous PR resolved apologies for the delay 

https://github.com/libre-tube/LibreTube/pull/7973

app/src/main/java/com/github/libretube/db/obj/StoredItemState.kt
Bnyro, line 1: unused class.

Bnyro:
Rebased only - gonna clean everything up and merge then.

app/src/main/java/com/github/libretube/notification/NotificationProvider.kt
Bnyro, line 1: It doesn't make sense to put this behind an interface abstraction - we only have a single implementation class and we're not an enterprise.

app/src/main/java/com/github/libretube/notification/NotificationHandler.kt
```diff
@@ -0,0 +1,76 @@
...
+import android.content.Context
+import androidx.core.app.NotificationCompat
+import androidx.work.WorkManager
+import com.github.libretube.LibreTubeApp.Companion.IMPORT_CHANNEL_NAME
+import com.github.libretube.R
+import com.github.libretube.enums.ImportState
+import com.github.libretube.receivers.ImportReceiver
+import java.util.UUID
+
+class NotificationHandler (private  val uuid: UUID,private val context: Context): NotificationProvider {
```
Bnyro:
```suggestion
class ImportNotificationHandler (private  val uuid: UUID,private val context: Context): NotificationProvider {
```

Bnyro:
Okay, this still needs more work than I thought - I don't have time to continue this now.

I also don't think that this is really helpful for the watch history, import workers for subscriptions and playlists would probably be more useful because they require API calls to YouTube for each imported video / channel.

SE-Bilal-Ahmad:
Okay so should i look in to importing subscriptions and playlists using this?

Bnyro:
> Okay so should i look in to importing subscriptions and playlists using this?
>
If you want to, sure! Otherwise I'll do it sometime next week.

Please make sure to pull my changes on your branch before continuing work :)

SE-Bilal-Ahmad:
> > Okay so should i look in to importing subscriptions and playlists using this?
> 
> If you want to, sure! Otherwise I'll do it sometime next week.
> 
> Please make sure to pull my changes on your branch before continuing work :)

Alright i will start working on it thanks

PR #8549, opened 2026-06-23
fix(TextUtils): correctly format weeks with DST transitions

### LINKED ISSUE #8295: Inaccurate relative time calculation

FineFindus:
Fixes an issues, where the relative week would be displayed as `0 weeks ago` if the week included a DST transitions. DST transitions are still taken into account for time spans less than 1 week.

Closes: https://github.com/libre-tube/LibreTube/issues/8295

Supersedes https://github.com/libre-tube/LibreTube/pull/8543/changes

### DIFF
--- a/app/src/main/java/com/github/libretube/util/TextUtils.kt
+++ b/app/src/main/java/com/github/libretube/util/TextUtils.kt
@@ -4,7 +4,6 @@ import android.content.Context
 import android.icu.text.RelativeDateTimeFormatter
 import android.net.Uri
 import android.text.format.DateUtils
-import androidx.core.text.isDigitsOnly
 import com.github.libretube.BuildConfig
 import com.github.libretube.R
 import com.github.libretube.extensions.formatShort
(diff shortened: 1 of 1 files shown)


Bnyro (approved):
Just for my understanding: If a DST change happened within the last 7 days (so that the week would effectively have one hour more) and we have a video that was uploaded `7 days + 30 minutes ago`, we would now display `1 week ago` although the week actually would only be complete if it was `7 days + 1 hour ago`?

Anyways the behavior doesn't matter as long as we don't display `0 weeks`, timezones always confuse me...

Btw, nice refactoring 👍

FineFindus:
In this case the issues was than it was less than a week (i.e. 7 days - 1 hour) and `until` is timezone-agnostic, but DateUtils isn't as it uses the milliseconds that have passed. I don't think an hour more would change anything (in the worst case we would display `7 days ago` which is correct.).

> timezones always confuse me...

Yeah, me too

PR #8551, opened 2026-06-24
fix: updated ungrouped channels immediately after adding to group

### LINKED ISSUE #8428: Ungroup Channel Group Not Updating
### Steps to reproduce

1. Add channels to a group
2. Look at the ungroup channels group
3.  See that the now grouped channel is still showing in it

### Expected behavior

Grouped channel is removed from the ungrouped category

### Actual behavior

Ungrouped category is the same as my all channels category

Image

### LibreTube version

v31.2 and 31.1

### Android version

16

### Other details

_No response_

### Acknowledgements

Phoenix1808:
Fixes #8428

## Issue

When a channel was added in the group, then it wasn't getting updated instantly and also was showing wrong count of the channels and the channel were still in the ungrouped section even after adding them in group.

## Fix

1. Ungrouped Count was using size of all the subsciptions rather than of ungrouped. Now it uses getUngroupedChannels().size.

2. On the long press -> Add to Group was updating the DB but not viewModel.groups hence list was lagging. Now after save viewModel.groups also gets updated hence instant refresh.

## Tested

On Physical Device 

WhatsApp Image 2026-06-24 at 3 52 05 PM

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/sheets/AddChannelToGroupSheet.kt
+++ b/app/src/main/java/com/github/libretube/ui/sheets/AddChannelToGroupSheet.kt
@@ -11,9 +11,13 @@ import com.github.libretube.ui.adapters.AddChannelToGroupAdapter
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
+import androidx.fragment.app.activityViewModels
+import com.github.libretube.ui.models.SubscriptionsViewModel
 
 class AddChannelToGroupSheet : ExpandedBottomSheet(R.layout.dialog_add_channel_to_group) {
     private lateinit var channelId: String
+
+    private val viewModel: SubscriptionsViewModel by activityViewModels()
 
     private val addToGroupAdapter by lazy(LazyThreadSafetyMode.NONE) {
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/github/libretube/ui/sheets/AddChannelToGroupSheet.kt
Bnyro, line 44: I thought about if it's possible to use the viewmodel groups here instead of loading them from the database (i.e. `val subscriptionGroups = viewModel.groups.value.orEmpty().toMutableList()`) - but that would only work if the user opened the subscriptions tab before-hand, so I think it's better the current way.

Bnyro (approved):
Good catch, thank you!

app/src/main/java/com/github/libretube/ui/sheets/AddChannelToGroupSheet.kt
Phoenix1808: > I thought about if it's possible to use the viewmodel groups here instead of loading them from the database (i.e. `val subscriptionGroups = viewModel.groups.value.orEmpty().toMutableList()`) - but that would only work if the user opened the subscriptions tab before-hand, so I think it's better the current way.

Exactly !

PR #8557, opened 2026-06-26
chore: lower minimum SDK requirement to API 25

fish47:
This patch adds LibreTube compatibility for Nougat devices.

The change is not that big, so I think it is worth bringing old devices back to life.

As the StorageManager API used for image cache size calculation is unavailable on Nougat, I referenced commit 2200a4c to select the cache size (16/32/64/128/256/512 MB) based on internal storage instead.

### DIFF
--- a/app/src/main/java/com/github/libretube/helpers/ImageHelper.kt
+++ b/app/src/main/java/com/github/libretube/helpers/ImageHelper.kt
@@ -4,6 +4,7 @@ import android.content.Context
 import android.graphics.Bitmap
 import android.graphics.Color
 import android.net.Uri
+import android.os.Build
 import android.os.storage.StorageManager
 import android.widget.ImageView
 import androidx.core.content.getSystemService
@@ -24,6 +25,7 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.withContext
 import okhttp3.OkHttpClient
 import okhttp3.logging.HttpLoggingInterceptor
+import java.io.File
 import java.nio.file.Path
 
 object ImageHelper {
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/github/libretube/helpers/ImageHelper.kt
```diff
@@ -57,14 +59,26 @@ object ImageHelper {
...
-                    storageManager.getUuidForPath(context.coilFile)
-                )
+                fun Context.maxCacheSizeBytes(path: File): Long {
+                    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+                        getSystemService<StorageManager>()?.let {
+                            return it.getCacheQuotaBytes(it.getUuidForPath(path))
+                        }
+                    }
+
+                    val totalGb = path.totalSpace / (1024 * 1024 * 1024)
```
Bnyro, line 69: If dropping such code, you should add an explanation where this is from or what it is supposed to do

Bnyro:
Why? Please add a PR description.

FineFindus:
The minimum SDK was increased due to breaking issues on lower Android versions (e.g. https://github.com/libre-tube/LibreTube/issues/7039) and a lack of testing. How does this PR address these issues?

app/src/main/java/com/github/libretube/helpers/ImageHelper.kt
fish47: The legacy code (2200a4c) had an option to set the image cache size (16/32/64/128/256/512MB), but that code was deleted. I tried to select the image cache size based on the internal storage size. If these lines of code like to be confusing, I can write comments for them.

Bnyro: Yes, please. It's not obvious why you chose the numbers like that.

You don't need to explain the code - only why the numbers are chosen as they are.

fish47:
I updated the description. It looks like supporting legacy devices has been discussed before. If the conclusion remains the same, I think we can close this PR.

Bnyro (commented):
> The change is not that big, so I think it is worth bringing old devices back to life.
>
So what's the motivation behind the PR? I assume you have some old device that doesn't get software updates anymore? Given that SDK 25 (Android 7) is already 9 years old, I'm not sure if it's worth maintaining this, unless there are good reasons why SDK25 is still required.

I think the very minimum we could support based on our dependencies is SDK23, but agai. the question is at what maintenance cost.

Also, please make your PR title consistent with the changes: the title talks about SDK24, the changes decrease minsdk to 25 though.

fish47:
The git commit message should have been modified to match the PR title, but I left it unchanged.

PR #8561, opened 2026-06-27
 feat(Comment): display whether comment is edited

FineFindus:
Adds an icon to indicate that a comment has been edited.

image

Honestly not to sure about this one, since it doesn't really add any useful information.


Bnyro (commented):
The edit button kind of looks like it's clickable and allows the user to modify the comment.

I think it's better to just put `(edited)` behind the commentor date, e.g. `9 hours ago (edited)`.

PR #8571, opened 2026-07-06
chore: update SABR client version, add TODOs

FineFindus:


app/src/main/java/com/github/libretube/player/parser/SabrClient.kt
```diff
@@ -375,7 +376,7 @@ class SabrClient private constructor(
                     .setClientInfo(
                         StreamerContext.ClientInfo.newBuilder()
                             .setClientName(1)
-                            .setClientVersion("2.20250122.04.00")
+                            .setClientVersion("2.20260626.01.00")
```
Bnyro: I'm not sure if you can answer that, but do you know if that changes anything in the response layout or is that just some "useless" meta information that could be used for blocking outdated versions?

Bnyro:
Please squash when merging - I don't think these changes are worth 5 separate entries in the changelog :)

app/src/main/java/com/github/libretube/player/parser/SabrClient.kt
FineFindus: > if that changes anything in the response layout

Not to my knowledge (also I'm not sure protocol updates are bound to client versions, since the protocol is backwards-compatible).

> could be used for blocking outdated versions?

They have been know to block outdated clients (or rather deprecate them and force an update).

FineFindus:
> Please squash when merging - I don't think these changes are worth 5 separate entries in the changelog :)

Not sure what I should put into the squashed commit, since these changes are too different?

Bnyro:
> Not sure what I should put into the squashed commit, since these changes are too different?
>
- chore(SABR): add remaining todos
- refactor(SABR): fix linter warnings
- chore(SABR): update client version

FineFindus:
Ah, I thought squashing into a single commit.

Also changed it to using the client version retrieved by NewPipeExtractor, which is retrieved from the API, so it doesn't require regular updating.

Bnyro (approved):

PR #8577, opened 2026-07-09
Add relative seeking when dragging

### LINKED ISSUE #8569: Add relative seeking when dragging

Currently, seeking works by pressing on the playback position and dragging it directly. It would be nice to support relative dragging instead, similar to YouTube, where the seek position adjusts relative to the initial touch. This makes fine seeking easier and more precise.

### Other details

_No response_

### Acknowledgements

YasinMoridi:
Closes #8569 
Currently, DismissableTimeBar seeks in absolute mode: touching the bar jumps the thumb (and playback position) directly to the X-coordinate of the touch. This makes fine seeking difficult, especially on longer videos where a small touch offset translates to a large time jump.
This PR changes the seeking behavior to relative dragging, similar to YouTube: the seek position is now calculated as an offset from the player's position at the start of the drag, rather than being mapped directly from the touch's absolute X position.
Changes

Relative delta calculation: On ACTION_DOWN, no seek happens immediately. Once a drag is detected (past touch slop), the current playback position is captured as initialPlaybackPosition. Subsequent ACTION_MOVE events compute a time delta from the horizontal drag distance (deltaX / barWidth * duration) and apply it relative to that anchor.
Removed synthetic ACTION_DOWN forwarding: The previous approach faked a DOWN event at the touch location to trick DefaultTimeBar into absolute-seeking from that point. This is no longer needed since the scrub position is computed and set manually via setPosition.
setPosition override: Prevents background player progress updates from overwriting the bar's visual position while the user is actively dragging.
Manual listener notification: onScrubStart / onScrubMove / onScrubStop are now invoked directly with the calculated relative position instead of relying on DefaultTimeBar's internal scrub listener.
ACTION_CANCEL handling: Seeking is now correctly skipped when the touch is canceled (isCanceled check), in addition to the existing vertical dismiss-to-cancel threshold.
Live stream / invalid duration fallback: If player.duration is unavailable or <= 0, the view falls back to the previous (super) touch handling to avoid division-by-zero and undefined relative-seek behavior.

Behavior

Tap without drag: no-op (as before), just triggers performClick().
Drag horizontally: seek position moves relative to the player's position at drag start, not the absolute touch X.
Drag vertically past the dismiss threshold: cancels the seek on release (existing DismissableTimeBar behavior, preserved).

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/views/DismissableTimeBar.kt
+++ b/app/src/main/java/com/github/libretube/ui/views/DismissableTimeBar.kt
@@ -8,7 +8,6 @@ import android.view.ViewConfiguration
 import androidx.media3.common.Player
 import androidx.media3.common.util.UnstableApi
 import androidx.media3.ui.DefaultTimeBar
-import androidx.media3.ui.PlayerControlView
 import androidx.media3.ui.TimeBar
 import androidx.media3.ui.TimeBar.OnScrubListener
 import com.github.libretube.extensions.dpToPx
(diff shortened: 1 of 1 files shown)


FineFindus:
Thanks for the PR. Considering that the current behavior was deliberately introduced in https://github.com/libre-tube/LibreTube/pull/7757 it should be discussed first whether we want to return to the previous behavior.

This should also not be done by adding new code (especially since that is the default behavior in media3 Timebars), but rather by reverting the previous PR.

YasinMoridi:
> Thanks for the PR. Considering that the current behavior was deliberately introduced in #7757 it should be discussed first whether we want to return to the previous behavior.
> 
> This should also not be done by adding new code (especially since that is the default behavior in media3 Timebars), but rather by reverting the previous PR.

Thanks for the context! I wasn't aware of #7757
Happy to discuss this further. If the team decides relative seeking is the way to go, I can close this PR and instead put together a revert of #7757, as suggested, rather than reintroducing it with custom code

dhanuarf:
> Thanks for the PR. Considering that the current behavior was deliberately introduced in #7757 it should be discussed first whether we want to return to the previous behavior.
> 
> This should also not be done by adding new code (especially since that is the default behavior in media3 Timebars), but rather by reverting the previous PR.

Hmm I think this PR is trying to introduce different behavior than the DefaultTimeBar (which is just regular seekbar).  It's mentioned above:
>  Drag horizontally: seek position moves relative to the player's position at drag start, not the absolute touch X.

And It was also mentioned then in the #7757 by the author as the "Youtube sytle"

YasinMoridi:
> > Thanks for the PR. Considering that the current behavior was deliberately introduced in #7757 it should be discussed first whether we want to return to the previous behavior.
> > This should also not be done by adding new code (especially since that is the default behavior in media3 Timebars), but rather by reverting the previous PR.
> 
> Hmm I think this PR is trying to introduce different behavior than the DefaultTimeBar (which is just regular seekbar). It's mentioned above:
> 
> > Drag horizontally: seek position moves relative to the player's position at drag start, not the absolute touch X.
> 
> And It was also mentioned then in the #7757 by the author as the "Youtube sytle"

Yes, brother, I think so. I checked #7757's diff — it only changed tap-to-seek into drag-to-seek (gating the seek behind a drag gesture). The actual position calculation still delegated to DefaultTimeBar's default (absolute, touch-X-based) logic. It never implemented relative/offset-based seeking.
So this PR isn't reintroducing something #7757 removed — it's a distinct feature (relative dragging from the drag's starting point) that wasn't present before or after #7757. Reverting #7757 wouldn't give us this behavior; it would just bring back tap-to-seek.

dhanuarf:
Though we decided in #7757 that we use the "Tiktok style", after testing it for a while I personally like this "Youtube style" better though. Making fine adjustment is much easier than the Tiktok style.

But there's seem to be a bug, it cancels seeking even when the finger only slightly move below the seekbar.

YasinMoridi:
> Though we decided in #7757 that we use the "Tiktok style", after testing it for a while I personally like this "Youtube style" better though. Making fine adjustment is much easier than the Tiktok style.
> 
> But there's seem to be a bug, it cancels seeking even when the finger only slightly move below the seekbar.

Thanks for testing it out and for the feedback! Glad to hear you prefer this style too — I agree, the relative dragging makes fine adjustments a lot easier than the previous approach.
I'll look into the cancel-on-slight-vertical-move bug and push a fix.

Bnyro:
Sorry, your pull request description looks kind of AI generated. I don't want to offend you, but please notice that we have a 0-tolerance AI policy (i.e. all code and text, including PRs, must be fully human-written).

YasinMoridi:
> Sorry, your pull request description looks kind of AI generated. I don't want to offend you, but please notice that we have a 0-tolerance AI policy (i.e. all code and text, including PRs, must be fully human-written).

Yes, brother, I'm not very fluent in English and I use artificial intelligence to help me translate.

PR #8578, opened 2026-07-09
Default "Delete only watched videos" to true in delete all downloads dialog

bermount:
The "Delete only watched videos" option in the DownloadsFragment delete all dialog to be checked by default.

1. User Behavior: I believe clearing out only watched content would occur with a much higher frequency than the action of wiping the entire download library.

2. Data Safety and Cost: If a user accidentally confirms the dialog contrary to their intent, the cost(data/bandwidth cost, time/effort cost, cognitive cost, ...) of losing their entire library including unwatched content is far higher than the inconvenience of having to re-run the process to remove the remaining unwatched videos. This change improves overall user experience and prioritizes data safety by defaulting to the more conservative option while still allowing users to opt for a full deletion if they choose.

Another option could be to set the default behavior to delete only watched videos, changing the checkbox to "Delete all" for users who explicitly want to clear the entire library.

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/fragments/DownloadsFragment.kt
+++ b/app/src/main/java/com/github/libretube/ui/fragments/DownloadsFragment.kt
@@ -323,13 +323,13 @@ class DownloadsFragmentPage : DynamicLayoutManagerFragment(R.layout.fragment_dow
     }
 
     private fun showDeleteAllDialog(context: Context, adapter: DownloadsAdapter) {
-        var onlyDeleteWatchedVideos = false
+        var onlyDeleteWatchedVideos = true
 
         MaterialAlertDialogBuilder(context)
             .setTitle(R.string.delete_all)
             .setMultiChoiceItems(
                 arrayOf(getString(R.string.delete_only_watched_videos)),
-                null
+                booleanArrayOf(true)
             ) { _, _, selected ->
                 onlyDeleteWatchedVideos = selected
             }


Bnyro:
I personally don't have an opinion on this change as I never delete all downloads at once. So if @FineFindus doesn't object, it's fine to merge.

In addition, I'd propose to remember the previous user selection for this option (as we do in many other places too), so that e.g. if a user deleted all the last time that's the default when opening the dialog the next time, for convenience.

bermount:
> In addition, I'd propose to remember the previous user selection for this option (as we do in many other places too), so that e.g. if a user deleted all the last time that's the default when opening the dialog the next time, for convenience.

I agree that remembering user selections is a great pattern for convenience, and it makes sense in most parts of the app.

However, for this specific dialog, I intentionally kept it stateless. The core purpose of this PR is to build a strict safety net against accidental data loss. If we save the state and a user's last action was a full wipe, the dialog will default to that option the next time. A quick, habitual click could then wipe their entire library, which reintroduces the exact irreversible cost this PR is trying to eliminate. 
The "cost" of that mistake is just too high: wasted network data and bandwidth to re-download videos, the time spent rebuilding the library, and the sheer frustration of losing unwatched videos. That heavily outweighs the minor inconvenience of having to re-run the delete process.

FineFindus (approved):
I agree that this is a better default.

> However, for this specific dialog, I intentionally kept it stateless. The core purpose of this PR is to build a strict safety net against accidental data loss. If we save the state and a user's last action was a full wipe, the dialog will default to that option the next time. A quick, habitual click could then wipe their entire library, which reintroduces the exact irreversible cost this PR is trying to eliminate.

This shouldn't be solved by reducing the blast radius of the mistake, but rather by adding an option to undo the deletion, similar to how the app offers the option to undo changes for other actions.

Bnyro:
In this case an undo option is not that trivial to implement though - we would need to start a timer of e.g. 10 seconds after which we do the actual video deletion. If a user closes the app in the meantime, things get ugly.

For subscriptions e.g., it's easy to implement because the undo action can simply re-subscribe to the channel as that action is not related to large data amounts.

I still think that the dialog that asks you for confirmation is enough of a warning and users are clever enough to understand that the deletion is irreversible.

Bnyro:
I decided against changing this default. "Delete" buttons are usually expected to delete everything visible on the screen, so I think that these changes would be counter-intuitive.

Nevertheless, thank you for the PR :+1:

PR #8590, opened 2026-07-12
fix: correct isFinished logic in DownloadItem

bermount:
Update isFinished property to correctly determine if a download is complete. The logic now ensures downloadSize is valid and checks if the current file size is greater than or equal to the expected total size.


Bnyro (approved):

Bnyro:
Good catch :+1:

PR #8595, opened 2026-07-14
feat: make the whole item list clickable in `AddChannelToGroupSheet`

dhanuarf:
Improve accessibility when selecting channel groups.


FineFindus (approved):
Thanks!

PR #8605, opened 2026-07-18
attempt to fetch playlist video from database in AddToPlaylistActivity

nuturs:
As of now, whenever an user attempts to add a video through the share menu, video info already available in database is re-fetched from youtube and the user is left with a frozen screen until this action is complete

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/activities/AddToPlaylistActivity.kt
+++ b/app/src/main/java/com/github/libretube/ui/activities/AddToPlaylistActivity.kt
@@ -2,13 +2,16 @@ package com.github.libretube.ui.activities
 
 import android.content.Intent
 import android.os.Bundle
+import android.util.Log
 import androidx.core.net.toUri
 import androidx.core.os.bundleOf
 import androidx.lifecycle.lifecycleScope
 import com.github.libretube.R
 import com.github.libretube.api.MediaServiceRepository
+import com.github.libretube.api.PlaylistsHelper
 import com.github.libretube.api.obj.StreamItem
 import com.github.libretube.constants.IntentData
+import com.github.libretube.extensions.TAG
 import com.github.libretube.extensions.toastFromMainDispatcher
 import com.github.libretube.helpers.IntentHelper
 import com.github.libretube.helpers.PreferenceHelper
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/github/libretube/ui/activities/AddToPlaylistActivity.kt
```diff
@@ -39,18 +42,27 @@ class AddToPlaylistActivity : BaseActivity() {
...
         lifecycleScope.launch(Dispatchers.IO) {
-            val videoInfo = if (PreferenceHelper.getToken().isEmpty()) {
-                try {
-                    MediaServiceRepository.instance.getStreams(videoId).toStreamItem(videoId)
-                } catch (e: Exception) {
-                    toastFromMainDispatcher(R.string.unknown_error)
-                    withContext(Dispatchers.Main) {
-                        finish()
+
+            var videoInfo = try {
```
alihrhera, line 46: Suggestion: Consider extracting the video fetching logic into a separate function that only returns the result. Keeping UI actions (`toast`, `finish`) outside the data-loading function would make the flow easier to test and maintain.



For example:

```kotlin
   val videoInfo = try {
    PlaylistsHelper.getVideoFromPlaylist(videoId)?.toStreamItem()
        ?: loadVideoInfo(videoId)
} catch (e: Exception) {
    Log.e(TAG(), "Failed to load playlist video", e)
    null
}

  if (videoInfo == null) {
        toast(R.string.unknown_error)
        finish()
        return@launch
    }
````
and keep the loading logic isolated:

```kotlin
private suspend fun loadVideoInfo(videoId: String): StreamItem? {
    return if (PreferenceHelper.getToken().isEmpty()) {
        MediaServiceRepository.instance
            .getStreams(videoId)
            .toStreamItem(videoId)
    } else {
        StreamItem(videoId)
    }
}
```

Bnyro (commented):
Thanks a lot for the pull request!

I'm not really sure if there is a lot of benefit in doing this in `AddToPlaylistActivity` because the add to playlist action there only adds one video to a playlist at once - where this could definitely be very useful is for cloning YouTube playlists or importing playlists from a YouTube playlist export file. But it doesn't hurt to have it here, so no reason to not merge this :)

I agree with the comment from @alihrhera, that makes the code easier to read. Good to merge from my side after that!

Bnyro:
We decided against merging this because there probably isn't a big reduction in API calls by this in comparison to the additional code complexity.

Thanks anyways!

PR #8606, opened 2026-07-18
add a preference to control the stepSize for playback speed slider

nuturs:
after implementing the following code, i realized that we could instead have an preference that would turn off the discrete steps by removing the stepSize attribute from the speed slider


FineFindus:
Thanks for the PR, we're currently not planning to add preferences for small customization features, see https://github.com/libre-tube/LibreTube/issues/8465 for more details.

Issue #8604, opened 2026-07-18
In tablet mode, the Material button border in the navigation bar deforms when switching


BugeStudioTeam:
### Steps to reproduce

1, Enter the app
2, Then switch to the button corresponding to any Fragment

### Expected behavior

In tablet mode, when switching, the border of the Material button in the navigation bar adaptively deforms according to its length.

### Actual behavior

In tablet mode, the Material button border in the navigation bar deforms when switching

### LibreTube version

31.4(70)

### Android version

Android 13 (33)

### Other details

None

Video:

<p align="center">
  <video src="https://github.com/user-attachments/assets/1d520662-2bb2-454c-ab8a-2da565fa3df6" width="100%" controls>
  </video>
</p>

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [x] I have written a short but informative title.
- [x] I will fill out all of the requested information in this form.

PR #8607, opened 2026-07-20
feat: dynamic indicator width for bottom navigation on tablets

alihrhera:
Implements a mechanism to adjust the `itemActiveIndicatorWidth` of the bottom navigation bar based on the menu item count and container width, specifically for devices with a screen width of 600dp or greater.
this is fix for #8604 In tablet mode, the Material button border in the navigation bar deforms when switching

Changes:
- Added `applyIndicatorForSelection` to calculate and set the active indicator width when an item is selected.
- Introduced `cachedIndicatorWidths` to store calculated widths, avoiding redundant calculations.
- Implemented `calculateSelectedItemSize` using a delayed post to ensure the view hierarchy is laid out before measuring child widths.
- Updated `setOnItemSelectedListener` to trigger the indicator adjustment on large screens.

[Uploading Screen_recording_20260721_015012.webm…]()

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/activities/MainActivity.kt
+++ b/app/src/main/java/com/github/libretube/ui/activities/MainActivity.kt
@@ -20,6 +20,7 @@ import androidx.core.os.bundleOf
 import androidx.core.splashscreen.SplashScreen.Companion.installSplashScreen
 import androidx.core.view.allViews
 import androidx.core.view.children
+import androidx.core.view.doOnPreDraw
 import androidx.core.view.isNotEmpty
 import androidx.core.widget.NestedScrollView
 import androidx.lifecycle.lifecycleScope
@@ -238,9 +239,8 @@ class MainActivity : AbstractPlayerHostActivity() {
         val menuView =
             binding.bottomNav.getChildAt(0) as? ViewGroup ?: return
         val selectedId = binding.bottomNav.selectedItemId
-        if (cachedIndicatorWidths[selectedId] != null) {
-            binding.bottomNav.itemActiveIndicatorWidth =
-                cachedIndicatorWidths[selectedId] ?: 0
+        cachedIndicatorWidths[selectedId]?.let { cachedWidth ->
+            binding.bottomNav.itemActiveIndicatorWidth = cachedWidth
             return
         }
         calculateSelectedItemSize(selectedId, menuView, visibleCount)
@@ -254,11 +254,14 @@ class MainActivity : AbstractPlayerHostActivity() {
         val selectedIndex = (0 until binding.bottomNav.menu.size)
             .firstOrNull { binding.bottomNav.menu[it].itemId == selectedId }
             ?: return
-        binding.bottomNav.postDelayed({
-            val selectedChild = menuView.getChildAt(selectedIndex) ?: return@postDelayed
-            binding.bottomNav.itemActiveIndicatorWidth = selectedChild.width / visibleCount
-            cachedIndicatorWidths[selectedId] = selectedChild.width / visibleCount
-        }, 200)
+
+        val selectedChild = menuView.getChildAt(selectedIndex) ?: return
+
+        selectedChild.doOnPreDraw {
+            val width = selectedChild.width / visibleCount
+            binding.bottomNav.itemActiveIndicatorWidth = width
+            cachedIndicatorWidths[selectedId] = width
+        }
     }
 
     /**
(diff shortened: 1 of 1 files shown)


FineFindus:
Thanks for the PR. This feels like a lot of code for something that I would presume is (and should be) natively handled by the material widgets (and looking at the docs is intended to do so)?

alihrhera:
> Thanks for the PR. This feels like a lot of code for something that I would presume is (and should be) natively handled by the material widgets (and looking at the docs is intended to do so)?

You're right that Material exposes itemActiveIndicatorWidth for a fixed size, but it doesn't recompute based on the number of currently visible items (when tabs are conditionally hidden). 
This code keeps the indicator spanning the full tab width dynamically in that case  happy to simplify further if we don't actually need that.

app/src/main/java/com/github/libretube/ui/activities/MainActivity.kt
```diff
@@ -205,6 +210,9 @@ class MainActivity : AbstractPlayerHostActivity() {
         }
 
         binding.bottomNav.setOnItemSelectedListener {
+            if (resources.configuration.smallestScreenWidthDp >= 600) {
```
Bnyro: Please move `600` to a constant `MINIMUM_SCREEN_SIZE_TABLET` or something

app/src/main/java/com/github/libretube/ui/activities/MainActivity.kt
```diff
@@ -81,7 +85,8 @@ class MainActivity : AbstractPlayerHostActivity() {
     private val searchViewModel: SearchViewModel by viewModels()
     private val downloadViewModel: DownloadsViewModel by viewModels()
     private val playlistViewModel: PlaylistViewModel by viewModels()
-
+    //this to save width of the selected item Indicator after calculate it
```
Bnyro:
```suggestion
    // save width of selected bottom bar item Indicator after calculating it
```

app/src/main/java/com/github/libretube/ui/activities/MainActivity.kt
```diff
@@ -225,6 +233,37 @@ class MainActivity : AbstractPlayerHostActivity() {
         showUserInfoDialogIfNeeded()
     }
 
+    private fun applyIndicatorForSelection() {
+        val visibleCount = (0 until binding.bottomNav.menu.size)
+            .count { binding.bottomNav.menu[it].isVisible }
+        val menuView =
+            binding.bottomNav.getChildAt(0) as? ViewGroup ?: return
+        val selectedId = binding.bottomNav.selectedItemId
+        cachedIndicatorWidths[selectedId]?.let { cachedWidth ->
```
Bnyro, line 242: I don't really get why we need to cache the widths? Why can't we just do `calculateSelectedItemSize` only?

app/src/main/java/com/github/libretube/ui/activities/MainActivity.kt
```diff
@@ -225,6 +233,37 @@ class MainActivity : AbstractPlayerHostActivity() {
...
+            binding.bottomNav.itemActiveIndicatorWidth = cachedWidth
+            return
+        }
+        calculateSelectedItemSize(selectedId, menuView, visibleCount)
+    }
+
+    private fun calculateSelectedItemSize(
```
Bnyro: This should be named something like `updateBottomBarSelectionBackgroundSize`, methods that are named `calculate` shouldn't have side effects

Bnyro (commented):
I really appreciate your efforts into this!

But to be honest, this looks like a bug that should be fixed at Android's material components library and not from our side. This here is just very hacky and I don't think it's worth maintaining this code because it makes everything unnecessarily complex.

app/src/main/java/com/github/libretube/ui/activities/MainActivity.kt
alihrhera: > I don't really get why we need to cache the widths? Why can't we just do `calculateSelectedItemSize` only?

The calculation itself is relatively expensive. On lower-end devices, it can sometimes introduce a slight delay, so I thought caching the result and reusing it would be a better approach for performance.

alihrhera:
> I really appreciate your efforts into this!
> 
> But to be honest, this looks like a bug that should be fixed at Android's material components library and not from our side. This here is just very hacky and I don't think it's worth maintaining this code because it makes everything unnecessarily complex.

I agree with your point that this would ideally be fixed in the Android Material Components library rather than handled on our side.

However, I don’t think this workaround adds much complexity or makes the code hard to maintain. It’s also isolated and can be easily removed once Google provides an official fix, so I think it’s a reasonable temporary solution until then.

PR #8629, opened 2026-07-30
refactor(settings): remove label visibility, playback during call, behavior when minimized, notification stream thumbnails and watch history size settings

Bnyro:
- label visibility: Seems very pointless? only a visual change probably almost nobody uses
- playback during call: this has various bad side effects, e.g. if you start playback in another app then LibreTube would continue playing at the same time
- behavior when minimized: picture in picture can be disabled in the Android settings for the app
- show stream thumbnails in notifications: merged with data saver mode
- watch history size: has only been useful while the app didn't support pagination in the watch history (i.e. the UI became laggy). not needed anymore and might be confusing if users search the history and don't find what they're looking for because by default, old videos automatically got wiped once the history reached a size of 100

### DIFF
--- a/app/src/main/java/com/github/libretube/db/DatabaseHelper.kt
+++ b/app/src/main/java/com/github/libretube/db/DatabaseHelper.kt
@@ -24,19 +24,6 @@ object DatabaseHelper {
     suspend fun addToWatchHistory(watchHistoryItem: WatchHistoryItem) =
         withContext(Dispatchers.IO) {
             Database.watchHistoryDao().insert(watchHistoryItem)
-            val maxHistorySize = PreferenceHelper.getString(
-                PreferenceKeys.WATCH_HISTORY_SIZE,
-                "100"
-            )
-            if (maxHistorySize == "unlimited") {
-                return@withContext
-            }
-
-            // delete the first watch history entry if the limit is reached
-            val historySize = Database.watchHistoryDao().getSize()
-            if (historySize > maxHistorySize.toInt()) {
-                Database.watchHistoryDao().delete(Database.watchHistoryDao().getOldest())
-            }
         }
 
     suspend fun getWatchHistoryPage(page: Int, pageSize: Int): List<WatchHistoryItem> {
--- a/app/src/main/java/com/github/libretube/helpers/NavBarHelper.kt
+++ b/app/src/main/java/com/github/libretube/helpers/NavBarHelper.kt
@@ -13,7 +13,6 @@ import com.github.libretube.R
 import com.github.libretube.constants.PreferenceKeys
 import com.github.libretube.ui.dialogs.NavBarItem
 import com.google.android.material.bottomnavigation.BottomNavigationView
-import com.google.android.material.navigation.NavigationBarView
 
 object NavBarHelper {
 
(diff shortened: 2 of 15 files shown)


app/src/main/java/com/github/libretube/db/DatabaseHelper.kt
```diff
@@ -24,19 +24,6 @@ object DatabaseHelper {
...
-            )
-            if (maxHistorySize == "unlimited") {
-                return@withContext
-            }
-
-            // delete the first watch history entry if the limit is reached
-            val historySize = Database.watchHistoryDao().getSize()
-            if (historySize > maxHistorySize.toInt()) {
-                Database.watchHistoryDao().delete(Database.watchHistoryDao().getOldest())
-            }
```
FineFindus, line 39: So watch history now doesn't have a limit anymore? I think we should set a high limit to avoid indefinite data growth (similar to how playlists are limited to 5k videos).

app/src/main/java/com/github/libretube/helpers/NavBarHelper.kt
```diff
@@ -73,15 +73,7 @@ object NavBarHelper {
...
-        val labelVisibilityMode = when (
-            PreferenceHelper.getString(PreferenceKeys.LABEL_VISIBILITY, "selected")
-        ) {
-            "always" -> NavigationBarView.LABEL_VISIBILITY_LABELED
-            "selected" -> NavigationBarView.LABEL_VISIBILITY_SELECTED
-            "never" -> NavigationBarView.LABEL_VISIBILITY_UNLABELED
-            else -> NavigationBarView.LABEL_VISIBILITY_AUTO
-        }
-        bottomNav.labelVisibilityMode = labelVisibilityMode
+        bottomNav.labelVisibilityMode = NavigationBarView.LABEL_VISIBILITY_SELECTED
```
FineFindus: I think we can set this in the UI xml now?

Bnyro: Good catch, thanks!

app/src/main/java/com/github/libretube/db/DatabaseHelper.kt
Bnyro: > similar to how playlists are limited to 5k videos
>
Are they really? Doesn't look like it: https://github.com/libre-tube/LibreTube/blob/master/app/src/main/java/com/github/libretube/repo/LocalPlaylistsRepository.kt#L41.

Maybe it makes more sense to allow users to delete all history entries that are older than X years? That's also problematic because we don't store the addition date yet btw.

I'd suspect that each video takes about `500B` by inspecting a sample video quickly, so 5k videos would only be `2.5MB`. Watching 5K videos takes you half a year if you watch 30 videos each day (which is already a lot of videos imo). Let's say 5MB per year, 50MB in 10 years. I don't think anyone will still use the app in 10 years to be honest (either YouTube will have broken it until then or we live in an AI apocalypse).

So I don't see how storage is an issue here realistically speaking.

FineFindus: > Are they really? Doesn't look like it: https://github.com/libre-tube/LibreTube/blob/master/app/src/main/java/com/github/libretube/repo/LocalPlaylistsRepository.kt#L41.

Yes, they are limited server-side: https://www.reddit.com/r/youtube/comments/m91k4t/why_is_there_a_5000_video_limit_to_playlists/

> I'd suspect that each video takes about 500B by inspecting a sample video quickly, so 5k videos would only be 2.5MB. Watching 5K videos takes you half a year if you watch 30 videos each day (which is already a lot of videos imo). Let's say 5MB per year, 50MB in 10 years. I don't think anyone will still use the app in 10 years to be honest (either YouTube will have broken it until then or we live in an AI apocalypse).

I thought it would be more (considering all the JVM overhead). If it's actually that little, if fine with that, otherwise we can still add a limit later on.

Bnyro: > Yes, they are limited server-side
>
Oh, you mean for YouTube, not LibreTube's internal database, okay.

> If it's actually that little, if fine with that, otherwise we can still add a limit later on.
>
I only tested it with a small SQLite database, I really can't tell how much overhead Android's Room ORM has. I expected it to be quite minimal, I think it mostly has computational overhead and not data storage space overhead.

I'd say we go without a limit for now and if we run into issues we can still add a limit.

PR #8637, opened 2026-08-03
fix: playlist videos list showing empty for some users

graffzepplin:
what was the issue:
  opening a playlist was showing the playlist name, thumbnail and
  video count correctly, but the video list itself was empty,
  "Nothing here" shown even though playlist have videos.

  checked logcat, no exception was thrown, so extractor call was
  success but returned no items. compared with NewPipe app, same
  playlist loads fine there.

  found NewPipeExtractorInstance.kt was calling NewPipe.init() without
  passing Localization/ContentCountry, so it fallback to extractor
  default locale instead of using device/app locale. NewPipe app
  passes this properly. this was causing youtube to return different
  playlist response for item list.

  fix:
  1) pass Localization and ContentCountry (from existing Language/Region
    preference) to NewPipe.init()
  2) also handled the case when playlist fetch throws exception, before
    this it was just stuck on loading spinner with no error, now shows
    error state with retry button

  tested on device, playlist which was not working before is loading
  fine now, tried few more playlist from different channel too, all
  working.

### DIFF
--- a/app/src/main/java/com/github/libretube/helpers/NewPipeExtractorInstance.kt
+++ b/app/src/main/java/com/github/libretube/helpers/NewPipeExtractorInstance.kt
@@ -1,16 +1,27 @@
 package com.github.libretube.helpers
+  
+  import android.content.Context
+  import com.github.libretube.util.NewPipeDownloaderImpl
+  import org.schabi.newpipe.extractor.NewPipe
+  import org.schabi.newpipe.extractor.ServiceList
+  import org.schabi.newpipe.extractor.StreamingService
+  import org.schabi.newpipe.extractor.localization.ContentCountry
+  import org.schabi.newpipe.extractor.localization.Localization
+  
+  object NewPipeExtractorInstance {
+      val extractor: StreamingService by lazy {
+          NewPipe.getService(ServiceList.YouTube.serviceId)
+      }   
+      
+      fun init(context: Context) {
+          @Suppress("DEPRECATION")
+          val locale = LocaleHelper.getAppLocale()
+          val region = PreferenceHelper.getTrendingRegion(context)
 
-import com.github.libretube.util.NewPipeDownloaderImpl
-import org.schabi.newpipe.extractor.NewPipe
-import org.schabi.newpipe.extractor.ServiceList
-import org.schabi.newpipe.extractor.StreamingService
-
-object NewPipeExtractorInstance {
-    val extractor: StreamingService by lazy {
-        NewPipe.getService(ServiceList.YouTube.serviceId)
-    }
-
-    fun init() {
-        NewPipe.init(NewPipeDownloaderImpl())
-    }
-}
+          NewPipe.init(
+              NewPipeDownloaderImpl(),
+              Localization(locale.language, locale.country.ifEmpty { null }),
+              ContentCountry(region)
+          )
+      }
+  }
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/github/libretube/helpers/NewPipeExtractorInstance.kt
```diff
@@ -1,16 +1,27 @@
...
-        NewPipe.init(NewPipeDownloaderImpl())
-    }
-}
\ No newline at end of file
+          NewPipe.init(
+              NewPipeDownloaderImpl(),
+              Localization(locale.language, locale.country.ifEmpty { null }),
+              ContentCountry(region)
+          )
+      }
```
FineFindus, line 26: Thanks for the PR.
I'm fairly surprised by this. I'll have to recheck, but I'm relatively confident that the extractor does not yet support using other languages for extraction (always using `en-GB`, see https://github.com/AudricV/NewPipeExtractor/tree/yt_support-non-english-interface-languages), in which case we should treat this as a bug in the extractor.

app/src/main/java/com/github/libretube/ui/fragments/PlaylistFragment.kt
```diff
@@ -133,13 +131,20 @@ class PlaylistFragment : DynamicLayoutManagerFragment(R.layout.fragment_playlist
...
             val response = try {
                 withContext(Dispatchers.IO) {
                     PlaylistsHelper.getPlaylist(playlistId)
                 }
             } catch (e: Exception) {
                 Log.e(TAG(), e.toString())
+                val binding = _binding ?: return@launch
+                binding.playlistProgress.isGone = true
+                binding.playlistError.isVisible = true
+                binding.playlistRetry.setOnClickListener { fetchPlaylist() }
```
FineFindus, line 147: I'm fine with showing an error, however, we should show a retry button, as it's unlikely to succeed, since the extraction is already broken (unless there is an A/B-test, but in this case I would prefer to get notified by users sooner, rather than later so we can fix the issue). It's also breaks with our design, as no other pages have retry action and retrying is already possible, simply by reopening the page.

FineFindus (changes requested):

app/src/main/java/com/github/libretube/helpers/NewPipeExtractorInstance.kt
graffzepplin: Hey, quick update on this one.

  I went back and tested it properly before making any more changes. I installed the current debug version of the app (v31.4), the same one I was using when I first found the bug, and tried the same playlists again with no changes at all. They still failed to load,
  showing "Nothing here" just like before, so the bug was really there.

  Then I tried the latest master code, but with my changes removed. Same playlists, same phone, everything else same. This time they loaded fine.

  So turns out the bug is already fixed on master, just not because of my change. Something else got updated in between (probably a library update) and that's what actually fixed it. My change was not doing anything for this specific bug, I just
  happened to test it after that other fix was already in place, so it looked like mine worked.

  Since my change isn't really fixing anything here, I think it's best if this PR gets closed. Feel free to close it from your side whenever you get a chance. Thanks for the review, and thanks for pointing out the language issue too, that made me look
  into this more closely and figure out what was actually going on.

  Also, since the whole reason I made this PR was the playlist not loading, and that is already fixed on master now, maybe it's a good time to push out a new release. People on the current release build are probably still facing the same empty playlist
  issue, and they don't need to wait on this PR for it, a new build would already fix it for them.

Issue #8155, opened 2026-02-01
Fullscreen button disappears when Auto-Fullscreen is enabled (vertical videos)


a-ettahiri:
### Steps to reproduce

When enabling Settings → Player → Auto-fullscreen in LibreTube v0.30.0, the fullscreen button disappears from the player UI when opening a video. This causes an issue specifically for vertical videos, where manual full-screen control is still required.

1. Open LibreTube v0.30.0
2. Go to Settings → Player
3. Enable Auto-fullscreen
4. Open any video (especially a vertical/portrait video)

### Expected behavior

- The fullscreen button should remain visible in the player UI.
- Users should be able to manually toggle fullscreen when needed, especially for vertical videos.

### Actual behavior

- The fullscreen button disappears entirely.
- Horizontal videos still enter fullscreen automatically when rotating the device, so this is not a major issue there.
- For vertical videos, full-screen must be triggered manually, but this becomes impossible because the full-screen button is missing.

### LibreTube version

0.30.0

### Android version

Android 13

### Other details

![Image](https://github.com/user-attachments/assets/dacc14f9-c09a-4a9e-8820-2ef01c3f51bf)

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [x] I have written a short but informative title.
- [x] I will fill out all of the requested information in this form.

ULTRA-LEGENDS:
Hi please Tap the three dots then go to Settings → Player → Behavior and disable the Auto-fullscreen option

a-ettahiri:
@ULTRA-LEGENDS i know it works that way, it becomes annoying to click it every time you open a new video instead of it being automatically turned when you rotate your phone, i like the auto-rotate future the problem become just when you need to watch a channel that have vertical videos and you don't find the full screen button

ULTRA-LEGENDS:
@FineFindus What do you think about this issue? Do you think adding a new option to address it would be better or modifying these settings?

FineFindus:
The full screen icon is purposefully hidden when auto-rotate is enabled, though I'm not sure why, that was before my time in the project. In any case adding a new option is nearly always the wrong answer.

FWIW videos can be fullscreened by swiping up on them. There is also a setting to auto-fullscreen shorts, which should probably be merged into the other setting or just made the default.

a-ettahiri:
[...]

PR #8658, opened 2026-08-19
fix: broken player layout when leaving fullscreen without rotating

prefixFelix:
Fixes the fullscreen issues reported in the https://github.com/libre-tube/LibreTube/issues/8155#issuecomment-3837620822. 

Leaving fullscreen in landscape via the exit-gesture (without rotating) kept the portrait layout inflated, so the video was cut off and the description and action buttons were pushed off screen (see video).
For this to happen auto-fullscreen must be enabled and either portrait or auto-rotation must be selected as screen orientation. 

The player view is now re-created so that the layout of the current orientation is used, and auto-fullscreen is suppressed until the device leaves landscape, so it doesn't immediately enter fullscreen again.

https://github.com/user-attachments/assets/cf9d1042-8108-4f6c-9f42-93e42ca2eaf5

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
+++ b/app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
@@ -39,6 +39,7 @@ import androidx.core.view.isVisible
 import androidx.fragment.app.Fragment
 import androidx.fragment.app.activityViewModels
 import androidx.fragment.app.commit
+import androidx.fragment.app.commitNow
 import androidx.fragment.app.viewModels
 import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.lifecycleScope
@@ -156,6 +157,12 @@ class PlayerFragment : Fragment(R.layout.fragment_player), CustomPlayerCallback 
      */
     private var playerLayoutOrientation = Int.MIN_VALUE
 
+    /**
+     * Whether auto fullscreen must not enter fullscreen for the current landscape orientation
+     * This is the case if the user left fullscreen without rotating the device.
+     */
+    private var isAutoFullscreenSuppressed = false
+
     // Activity that's active during PiP, can be used for controlling its lifecycle.
     private var pipActivity: Activity? = null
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
```diff
@@ -841,6 +852,46 @@ class PlayerFragment : Fragment(R.layout.fragment_player), CustomPlayerCallback
             !ThemeHelper.isDarkMode(requireContext())
     }
 
+    /**
+     * Leave fullscreen due to a user interaction instead of a rotation to portrait
+     */
+    private fun exitFullscreen() {
```
Bnyro, line 858: The naming doesn't really make it clear what the difference between `exitFullscreen` and `unsetFullscreen` is. Can't we just merge both into `unsetFullscreen` and do something like

```kotlin
if (PlayerHelper.autoFullScreenEnabled && orientation != playerLayoutOrientation) {
    recreateViewForCurrentOrientation()
}
```

?

app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
```diff
@@ -1426,9 +1480,12 @@ class PlayerFragment : Fragment(R.layout.fragment_player), CustomPlayerCallback
         if (PlayerHelper.autoFullscreenEnabled) {
             when (newConfig.orientation) {
                 // go to fullscreen mode
-                Configuration.ORIENTATION_LANDSCAPE -> setFullscreen()
-                // exit fullscreen if not landscape
-                else -> unsetFullscreen()
+                Configuration.ORIENTATION_LANDSCAPE -> if (!isAutoFullscreenSuppressed) setFullscreen()
```
Bnyro, line 1483: Please refactor the `when (newConfig.orientation)` into an `if` statement to also include `!isAutoFullscreenSuppressed`

Bnyro:
Feels kinda ugly to add yet another workaround for the autofullscreen option.

I would have preferred to remove it to make maintenance easier, but I can imagine that quite a lot of people are using it.

PR #8367, opened 2026-04-25
Support for LibreTube's new Sync server

Bnyro:
Please read https://github.com/libre-tube/LibreTube/issues/8363 for more information.

closes https://github.com/libre-tube/LibreTube/issues/8363

NOTE:
- The sync server is still in alpha and it's expected that we will wipe the server's database a few times to adapt the database scheme (too lazy to do migrations while it's not in production), so please don't use it to store your account data there yet (but for testing, of course, please do so!).

PR #8669, opened 2026-08-20
fix: unwatched scrollback now continues forever

ItsMeSamey:
Previously, the unwatched scrollback was super slow and would just stop in the middle because the filtering was done in kotlin and the recieved chunks may not even have any unwatched entries. This pr fixes this and makes the infinite scrolling in history smoother by
- adding prefetching
- moved filtering into SQL
- using cursor instead  of repeated `OFFSET` queries that rescan  the rows


ItsMeSamey:
the main issue only arises when we have a long contiguous chunk of videos all of which are watched. When we encounter this during scrolling, the pagination breaks. A hacky way atound this is to scroll the full [watched+unwatched] list to beyond that position, which caused the unwatched videos to render properly; but this was super tedious and janky.

app/src/main/java/com/github/libretube/db/dao/WatchHistoryDao.kt
```diff
@@ -1,19 +1,74 @@
...
 interface WatchHistoryDao {
     @Query("SELECT * FROM watchHistoryItem")
     suspend fun getAll(): List<WatchHistoryItem>
 
-    @Query("SELECT * FROM watchHistoryItem LIMIT :limit OFFSET :offset")
-    suspend fun getN(limit: Int, offset: Int): List<WatchHistoryItem>
+    @Query(
+        """
+        SELECT h.*, h.rowid AS historyRowId, p.position AS watchPosition,
+            EXISTS(SELECT 1 FROM download AS d WHERE d.videoId = h.videoId) AS isDownloaded
```
Bnyro: I don't really like to mix downloads and watch history logic, that will break https://github.com/libre-tube/LibreTube/pull/8367.

It would be better to drop this part here and create a new method like `areVideosDownloaded(videoIds: List<String>): List<Boolean>` in `DownloadDao`. That would require a second database call, I know, but that will continue to work after https://github.com/libre-tube/LibreTube/pull/8367 and the performance is still much better than doing one database query per item.

Bnyro: (same for `getFilteredPage`)

app/src/main/java/com/github/libretube/db/dao/WatchHistoryDao.kt
```diff
@@ -1,19 +1,74 @@
...
+    )
+    suspend fun getPage(limit: Int, cursor: Long): List<WatchHistoryPageItem>
+
+    @Query(
+        """
+        SELECT h.*, h.rowid AS historyRowId, p.position AS watchPosition,
+            EXISTS(SELECT 1 FROM download AS d WHERE d.videoId = h.videoId) AS isDownloaded
+        FROM watchHistoryItem AS h
+        LEFT JOIN watchPosition AS p ON p.videoId = h.videoId
+        WHERE h.rowid <= :cursor AND CASE WHEN
```
Bnyro: Duplicating the `isVideoWatched` logic into SQL is a bit ugly, because we need to keep the code in sync. But I don't have a better idea.

app/src/main/java/com/github/libretube/db/dao/WatchHistoryDao.kt
```diff
@@ -1,19 +1,74 @@
...
+        END = :watched
+        ORDER BY h.rowid DESC
+        LIMIT :limit
+        """
+    )
+    suspend fun getFilteredPage(
+        limit: Int,
+        cursor: Long,
+        watched: Int,
+        absoluteWatchedThreshold: Float,
```
Bnyro:
```suggestion
        absoluteWatchedThresholdSeconds: Float,
```

app/src/main/java/com/github/libretube/db/DatabaseHelper.kt
```diff
@@ -26,19 +27,42 @@ object DatabaseHelper {
...
+    suspend fun getWatchHistoryPage(
+        pageSize: Int,
+        statusFilter: Int = 0,
+        cursor: Long = Long.MAX_VALUE
+    ): WatchHistoryPage {
+        val watchHistoryDao = Database.watchHistoryDao()
+        val watched = when (statusFilter) {
+            0 -> null
+            1 -> 0
+            2 -> 1
```
Bnyro:
```suggestion
            0 -> STATUS_ALL
            1 -> STATUS_CONTINUE_WATCHING
            2 -> STATUS_FINISHED
```

Bnyro (commented):
I agree that it's a good idea to move the filtering into the database, thanks for looking into that.

However, the changes seem very complex for what they are doing, especially the changes to `WatchHistoryModel`.

app/src/main/java/com/github/libretube/ui/models/WatchHistoryModel.kt
```diff
@@ -1,84 +1,137 @@
...
+
             PreferenceHelper.putInt(PreferenceKeys.SELECTED_HISTORY_STATUS_FILTER, value)
-            selectedStatus.value = value
+            statusFilter = value
+            fetchJob?.cancel()
+            fetchJob = null
+            filterGeneration++
+            cursor = Long.MAX_VALUE
+            reachedEnd = false
+            watchHistory.value = emptyList()
```
Bnyro: I think it would be better to move keep `selectedStatus` a `MutableStateFlow` and move this logic into an observer that listens for `selectedStatus` changes. The setter of `selectedStatusFilter` is only supposed to set its value (as the name implies), not start data loading tasks.

app/src/main/java/com/github/libretube/ui/models/WatchHistoryModel.kt
```diff
@@ -1,84 +1,137 @@
...
-        }
-    }
+        val requestedCursor = cursor
+        val generation = filterGeneration
+        val requestedStatus = selectedStatusFilter
+        fetchJob = viewModelScope.launch {
+            val page = withContext(Dispatchers.IO) {
+                DatabaseHelper.getWatchHistoryPage(
+                    pageSize = HISTORY_PAGE_SIZE,
+                    statusFilter = requestedStatus,
```
Bnyro:
```suggestion
        fetchJob = viewModelScope.launch {
            val page = withContext(Dispatchers.IO) {
                DatabaseHelper.getWatchHistoryPage(
                    pageSize = HISTORY_PAGE_SIZE,
                    statusFilter = statusFilter,
```

app/src/main/java/com/github/libretube/ui/models/WatchHistoryModel.kt
```diff
@@ -1,84 +1,137 @@
...
+                else -> false
+            }
+        }
+
+        watchHistory.value = history.toMutableList().apply {
+            removeAt(index)
+            if (refreshedRow != null) {
+                add(if (refreshedRow.rowId > history.first().rowId) 0 else index, refreshedRow)
+            }
+        }
```
Bnyro, line 94: This feels very error-prone, it might be better to just pass a boolean `isVideoWatched` from `VideoOptionsBottomSheet` down to `WatchHistoryAdapter` and only update the state in the `WatchHistoryAdapter`'s item UI and add/remove it to `watchHistory.value` via the `onWatchStatusChanged` callback.

ItsMeSamey:
Simplified `WatchHistoryModel` as much as i could. Still has to be a little complex because am doing prefething instead of blocking the ui for loading entries.

Thanks for the review :)

app/src/main/java/com/github/libretube/db/dao/WatchHistoryDao.kt
ItsMeSamey: Download status is now seperately fetched.

app/src/main/java/com/github/libretube/ui/models/WatchHistoryModel.kt
ItsMeSamey: Done

app/src/main/java/com/github/libretube/ui/models/WatchHistoryModel.kt
ItsMeSamey: Done

app/src/main/java/com/github/libretube/enums/WatchHistoryStatus.kt
```diff
@@ -0,0 +1,7 @@
+package com.github.libretube.enums
+
+enum class WatchHistoryStatus(val isWatched: Boolean?) {
+    ALL(null),
+    CONTINUE_WATCHING(false),
+    FINISHED(true)
+}
```
FineFindus, line 7: This is quite confusing, it's not clear to me what `isWatched = null` is meant to mean (I assume both watched and not watched videos?). Is probably better to use a different abstractions here to make the intent clear (e.g. using an enum or bitset, or maybe this can be entire avoided?)

app/src/main/java/com/github/libretube/ui/models/WatchHistoryModel.kt
```diff
@@ -1,84 +1,99 @@
...
+                pageSize = HISTORY_PAGE_SIZE,
+                statusFilter = selectedStatus.value,
+                cursor = currentCursor
+            )
+            val downloaded = DatabaseHolder.Database.downloadDao()
+                .areVideosDownloaded(page.items.map(WatchHistoryItem::videoId))
+
+            page.rows.forEachIndexed { index, row ->
+                val item = row.item
+                if (downloaded[index]) downloadedVideoIds += item.videoId else downloadedVideoIds -= item.videoId
```
FineFindus, line 66:
```suggestion
                if (downloaded[index]) {
                    downloadedVideoIds += item.videoId
                } else {
                    downloadedVideoIds -= item.videoId
                }
```

nit: formatting, for readability

app/src/main/java/com/github/libretube/db/DatabaseHelper.kt
```diff
@@ -26,19 +28,27 @@ object DatabaseHelper {
...
+        statusFilter: WatchHistoryStatus = WatchHistoryStatus.ALL,
+        cursor: Long = Long.MAX_VALUE
+    ): WatchHistoryPage {
+        val rows = Database.watchHistoryDao().getPage(
+            limit = pageSize,
+            cursor = cursor,
+            watched = statusFilter.isWatched,
+            absoluteWatchedThresholdSeconds = ABSOLUTE_WATCHED_THRESHOLD,
+            relativeWatchedThreshold = RELATIVE_WATCHED_THRESHOLD
+        )
```
FineFindus, line 46:
```suggestion
    suspend fun getWatchHistoryPage(
        pageSize: Int,
        statusFilter: WatchHistoryStatus = WatchHistoryStatus.ALL,
        cursor: Long? = null,
    ): WatchHistoryPage {
        val rows = Database.watchHistoryDao().getPage(
            limit = pageSize,
            cursor = cursor ?: Long.MAX_VALUE,
            watched = statusFilter.isWatched,
            absoluteWatchedThresholdSeconds = ABSOLUTE_WATCHED_THRESHOLD,
            relativeWatchedThreshold = RELATIVE_WATCHED_THRESHOLD
        )
```

I think API-wise this may be better, since it's indicates that a cursor is optional and hides the cursor impl from the caller (just as an idea though).

FineFindus (commented):
Thanks for the PR!

> adding prefetching

Did you do any testing, whether prefetching is still necessary, after moving the filtering to SQL? Also, I think there is a native view thingy that handles the prefetching already, rather than needing to implement it yourself.

Bnyro:
> Did you do any testing, whether prefetching is still necessary, after moving the filtering to SQL? Also, I think there is a native view thingy that handles the prefetching already, rather than needing to implement it yourself.
>
Good point, we could/should use [paging 3](https://developer.android.com/topic/libraries/architecture/paging/v3-overview) like in https://github.com/libre-tube/LibreTube/pull/5528.

app/src/main/java/com/github/libretube/enums/WatchHistoryStatus.kt
Bnyro: I agree, but that's just the value that is passed to the database query.

Adding more abstractions here would make everything more complex. I've added a doc comment what `isWatched` means.

Bnyro (approved):
Thanks a lot @ItsMeSamey!

PR #8681, opened 2026-08-20
chore: created nightly.svg widget from universal.svg

IlliterateLexicon:

Issue #8655, opened 2026-08-18
An Easier Way to Install & Run Either Nightly or Quick Patches


key6oardWarrior:
### Describe your suggested feature

Some less technical users may not know how or may not have the time to learn how to compile an Android app. So in the event there was an issue that required a hot fix, but it won't be available until the next release this will give them an easier way to install this patch. This could also increase the number of testers this app has since there could be more users using the non-release versions of the app.

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [x] I have written a short but informative title.
- [x] I will fill out all of the requested information in this form.

headquarter8302:
Seconding this. Simply putting an evergreen `nightly` release that gets updated by CI periodically is enough for half of the use cases (e.g. using an app installation helper like Obtainium).

A lot of errors are fixed in nightly, but ordinary users get no indication of a fix being available

dftf-stu:
Yes, like the vast-majority of people in the world, I _never_ compile from source: I'm always using a pre-compiled APK.

Which means during the current "`source error`" on every video issue, I've no-choice* but to wait until an APK for the stable 31.5 release drops, as it seems the idea of rushing-out a minor 31.4.1 release APK to just address this one issue doesn't seem to be on-the-table. 😢

And it's also odd how there is no pinned-ticket at the top of the _Issues_ list here, given how prominent an issue this currently is.

| * okay, you could go to Menu > Settings > Backup and Restore > Export subscriptions , then import that into another app (BravePipe, 
|    NewPipe, PipePipe, Skytube, etc.) in the meantime, but it does get tedious having to keep switching apps as Google break them!

IlliterateLexicon:
There do appear to be builds updated occasionally from the repo "libre-tube/NightlyBuilds" however no tags / releases are published so obtainium doesnt seem to work. 

Manually installing the nightly build appears to have mostly resolved the "source error" issue as far as I can tell.

dftf-stu:
The suggestion from @IlliterateLexicon should indeed work: go to [https://github.com/libre-tube/NightlyBuilds](https://github.com/libre-tube/NightlyBuilds) and then tap the "DOWNLOAD" graphic, and it links to an APK file.

I still think though, given how prominent this issue is, they should just publish a 31.4.1 update on this site (which could then go-out to alternate Android stores, like F-Droid and IzzySoft), rather than make people wait for the fix to be included in 32.0

[...]

PR #8691, opened 2026-08-24
ci: create / update nightly pre-releases

IlliterateLexicon:
uses https://github.com/ncipollo/release-action to create a release in NightlyBuilds

view the test releases here https://github.com/illiteratelexicon/nightlybuilds/releases

possible fix for https://github.com/libre-tube/LibreTube/issues/8655 (it should allow obtainium or similar to pull nightly apks)

(edit) 
also changes the url in the readme to point to the latest release as mentioned here https://github.com/libre-tube/LibreTube/pull/8681


IlliterateLexicon:
it seems to fail when checking out the nightly repo with "Error: Input required and not supplied: token" however the token appears to be correctly passed afiak (https://github.com/IlliterateLexicon/LibreTube/blob/master/.github/workflows/ci.yml#L70). 

(I think it may be due to github not passing secrets to workflows triggered by forked prs?)

"works on my repo TM"

IlliterateLexicon:
Lmk if anymore work needs to happen for this to get merged. Thanks you!

Bnyro:
Thanks for the PR. I'll look into it after my vacation (in a week or so)

IlliterateLexicon:
Have a good vacation! ^^

.github/workflows/ci.yml
```diff
@@ -61,19 +64,31 @@ jobs:
...
         run: |
-          mv .github/uploader.py .
-          echo "GH_REPO = '${{ github.repository }}'" > config_file.py
-          git clone --filter=blob:none https://github.com/libre-tube/NightlyBuilds nightly
-          rm -rf nightly/*.apk
-          mv app/build/outputs/apk/debug/*.apk nightly/
           cd nightly
-          python ../uploader.py
+          git tag nightly-$DATE_$TIME
+          git push origin nightly-$DATE_$TIME
```
FineFindus: We use should a single `nightly` tag, rather than creating a new one for each release, similar to [Ghostty's tip tag](https://github.com/ghostty-org/ghostty/releases/tag/tip)

FineFindus (changes requested):

.github/workflows/ci.yml
IlliterateLexicon: I updated it to check if the 'nighty' tag exists and only make it if not.
I also made it a pre-release which I think is ideal. (this means to download it from obtainium you need to mark "Include prereleases") 

I also realized ive removed the "Upload to archive" step that uploads the apk to the repo. Should that be re-added?

Additionally since the release is being created in NightlyBuilds the source code archives that automatically get created (Source Code.zip/tar) dont actually contain the source code for libre-tube and instead just contain the contents of NightlyBuild

IlliterateLexicon:
I think it may be worth considering creating the nightly release as a periodically updated pre-release in the main repo while continuing to push apks to the nightly repo as before.

That way the history of the apks is perserved in nightly builds, the apks arent stored in the main repo history, and the source code packaged with the releases would be accurate instead of just including another version of the binary.

It would also allow the same repo to be used for nightly/regular releases in obtainium (enabling pre-releases on the repo would download the nightly version, disabling them would download regular)

IlliterateLexicon:
readded upload to archive step and am now using the repo it clones instead of checking out.

I belive the only change that would need to be made to push the pre-release to the main repo instead of the nightly one is the remove input "repo" on line 93 if we do end up going that route.

.github/workflows/ci.yml
Bnyro: > I also realized ive removed the "Upload to archive" step that uploads the apk to the repo. Should that be re-added?
>
No, it's correct to remove it.

If we only have a single nightly release, we could possibly even put that release into the main LibreTube repo so that we don't need the NightlyBuilds repo anymore? But I don't have a preference on that.

Bnyro:
> I think it may be worth considering creating the nightly release as a periodically updated pre-release in the main repo while continuing to push apks to the nightly repo as before.
>
I think it's fine to just delete the NightlyBuilds repo then, it's simpler for users if everything is in the LibreTube main repo.

IlliterateLexicon:
Alright. The only downside I see Is that there would be no access to historical nightly apks, but that honestly seems more consistent with other repos.

I'm leaving camping tommorw morning for a few days, but might edit it quickly tonight.

.github/workflows/ci.yml
FineFindus: > I updated it to check if the 'nighty' tag exists and only make it if not.

That shouldn't be necessary, since creating the tag is a one time action and it should always exist after that.

FineFindus:
> I think it's fine to just delete the NightlyBuilds repo then, it's simpler for users if everything is in the LibreTube main repo.

I agree that it would be simpler, but with the current workflow we would create a new release on every commit (which would notify everyone that is subscribed for releases). Maybe it's better to follow Ghostty's model here (again), use a single nightly release and update it with new artifacts?

IlliterateLexicon:
> use a single nightly release and update it with new artifacts?

I think that's already the case.

Using inputs `allowUpdates` and `updateOnlyUnreleased` (only update pre-releases) from https://github.com/ncipollo/release-action

IlliterateLexicon:
I belive it's now working and creates a "evergreen" pre-release under the tag nightly

.github/workflows/ci.yml
```diff
@@ -60,20 +65,31 @@ jobs:
...
+          token: ${{ secrets.GH_TOKEN }}
+          tag:  nightly
+
+          makeLatest: true
+          prerelease: true
+          allowUpdates: true
+          updateOnlyUnreleased: true
+          
+          artifacts: app/build/outputs/apk/debug/*.apk
+          name: LibreTube Nightly ${{ env.DATE }}
```
FineFindus, line 88:
```suggestion
        with:
          token: ${{ secrets.GH_TOKEN }}
          tag:  nightly
          makeLatest: true
          prerelease: true
          allowUpdates: true
          updateOnlyUnreleased: true
          artifacts: app/build/outputs/apk/debug/*.apk
          name: LibreTube Nightly ${{ env.DATE }}
```

To keep a common formatting between the CI job files.

FineFindus (approved):

Bnyro (approved):
Thank you!

I noticed that the link in the README for downloading the nightly builds still pointed to the NightlyBuilds repo, fixed that.

Let's see if it works by merging :)

Bnyro:
I switched to `github.token` instead of `secrets.GH_TOKEN` because that one has the permissions needed.

PR #8730, opened 2026-08-29
fix: Open the trends page in Piped mode

### LINKED ISSUE #8290: Trends Seccion -- Selecting Text
### Steps to reproduce

1. Open Libretube
2. Select "Trends" text seccion instead of logo option
3. A emergency window appears only showing these section with nothing (just logo option)
4. Click in any region / country

### Expected behavior

Instead of waiting or create any emergency window, just its has same working of pressing logo option

### Actual behavior

Its creates a emergency window and when you select any region / country its turns off and closes / reject from application.

### LibreTube version

0.30.0

### Android version

Android 16

### Other details

Error Log:
4
java.lang.IndexOutOfBoundsException: Index 0 out of bounds for length 0
	at jdk.internal.util.Preconditions.outOfBounds(Preconditions.java:64)
	at jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Preconditions.java:70)
	at jdk.internal.util.Preconditions.checkIndex(Preconditions.java:266)
	at java.util.Objects.checkIndex(Objects.java:391)
	at java.util.ArrayList.get(ArrayList.java:434)
	at androidx.work.impl.utils.CancelWorkRunnable$forId$1.invoke(Unknown Source:226)
	at com.github.libretube.ui.dialogs.LoginDialog$$ExternalSyntheticLambda2.onClick(Unknown Source:51)
	at androidx.preference.PreferenceFragmentCompat$1.handleMessage(Unknown Source:810)
	at android.os.Handler.dispatchMessage(Handler.java:110)
	at android.os.Looper.loopOnce(Looper.java:273)
	at android.os.Looper.loop(Looper.java:363)
	at android.app.ActivityThread.main(ActivityThread.java:10060)
	at java.lang.reflect.Method.invoke(Native Method)
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:632)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:975)

PDTA: Once trying again 'Trends' section dissapear on my home view.

### Acknowledgements

mvanhorn:
Change `PipedMediaServiceRepository.getTrendingCategories()` to expose `TrendingCategory.LIVE` as its single supported pager category, matching the existing Piped `getTrending` implementation that ignores category distinctions and fetches the instance's general trending feed. Opening the full Trends page from the home section in Piped mode produces an empty pager, and using its region action then crashes with an `IndexOutOfBoundsException`.

In a unit test, a fresh `PipedMediaServiceRepository` returns exactly `TrendingCategory.LIVE` from `getTrendingCategories()`, not an empty list or the multi-category NewPipe set; In Piped mode, selecting the Trends heading from Home opens a single-page trends feed while the category tab strip remains hidden.

Closes #8290

### DIFF
--- a/app/src/test/java/com/github/libretube/api/PipedMediaServiceRepositoryTest.kt
+++ b/app/src/test/java/com/github/libretube/api/PipedMediaServiceRepositoryTest.kt
@@ -1,0 +1,13 @@
+package com.github.libretube.api
+
+import org.junit.Assert.assertEquals
+import org.junit.Test
+
+class PipedMediaServiceRepositoryTest {
+    @Test
+    fun getTrendingCategoriesReturnsLiveCategory() {
+        val repository = PipedMediaServiceRepository()
+
+        assertEquals(listOf(TrendingCategory.LIVE), repository.getTrendingCategories())
+    }
+}
--- a/app/src/main/java/com/github/libretube/api/PipedMediaServiceRepository.kt
+++ b/app/src/main/java/com/github/libretube/api/PipedMediaServiceRepository.kt
@@ -16,7 +16,7 @@ import com.github.libretube.helpers.PreferenceHelper
 import retrofit2.HttpException
 
 open class PipedMediaServiceRepository : MediaServiceRepository {
-    override fun getTrendingCategories(): List<TrendingCategory> = emptyList()
+    override fun getTrendingCategories(): List<TrendingCategory> = listOf(TrendingCategory.LIVE)
 
     override suspend fun getTrending(region: String, category: TrendingCategory): List<StreamItem> =
         api.getTrending(region)


app/src/test/java/com/github/libretube/api/PipedMediaServiceRepositoryTest.kt
FineFindus, line 1: I don't think this tests adds anything of value, if we can't trust that a function returns a hardcoded value anymore we have larger issues :)

FineFindus (changes requested):
Thanks for the PR. Did you use AI?

I don't think this is a good solution, considering that the Piped trending API will always fail. It would be better to hide the trending page if Piped is enabled.

mvanhorn:
Yes, I used a coding agent for this PR. I review the output before submitting, but the first draft was agent-assisted.

On the substance, you're right and I'd rather not defend the approach. If the Piped trending API always fails, catching the failure is treating the symptom, and hiding the trending page when Piped is enabled is the actual fix. Fair point on the test too, asserting a hardcoded return value proves nothing.

I'll close this. If you'd like the hide-the-page version as a separate PR, say the word and I'll open one; otherwise I'll leave it to you.

therealresonix:
> Thanks for the PR. Did you use AI?
> 
> I don't think this is a good solution, considering that the Piped trending API will always fail. It would be better to hide the trending page if Piped is enabled.

So #8290 is not fixable?
@FineFindus

FineFindus:
> So https://github.com/libre-tube/LibreTube/issues/8290 is not fixable?

It's of course fixable.

PR #8733, opened 2026-08-30
feat: rework autoplay

### LINKED ISSUE #8715: Remove autoplay playlists toggle

In the interest of https://github.com/libre-tube/LibreTube/issues/8465 this could be removed. Most would assume the playlist to autoplay the next item. If desired a person could always still loop the current video.

### Other details

_No response_

### Acknowledgements

### LINKED ISSUE #8722: Remove insert related videos toggle

In the interest of https://github.com/libre-tube/LibreTube/issues/8465 this could be removed. Most people don't expect related videos to be added to their queue.

### Other details

_No response_

### Acknowledgements

FineFindus:
Reworks and simplifies autoplay. When a video ends, the play now tries to choose the next video from the
following options (in order):
- an explicitly selected next video
- the next queued video (e.g. from the queue or a playlist/channel)
- a related video, if available and autoplay is enabled

If none of these are available, the video endscreen is shown.

### Future enhancements

- Always enable the autoplay countdown (similar to the official app). Alternatively we could also remove it, considering it add a lot of complexity to handling autoplay
- The official app uses different videos for autoplay than related streams. It may be worthwhile to extract them
- [Disable autoplay if related videos are hidden?](https://github.com/libre-tube/LibreTube/issues/8722#issuecomment-5459453250
)

Closes: https://github.com/libre-tube/LibreTube/issues/8722
Closes: https://github.com/libre-tube/LibreTube/issues/8715


app/src/main/java/com/github/libretube/helpers/PlayerHelper.kt
```diff
@@ -369,8 +363,7 @@ object PlayerHelper {
         }
 
     fun isAutoPlayEnabled(isPlaylist: Boolean = false): Boolean {
-        return autoPlayEnabled || (isPlaylist && PreferenceHelper
-            .getBoolean(PreferenceKeys.AUTOPLAY_PLAYLISTS, false))
+        return isPlaylist || autoPlayEnabled
```
Bnyro: I don't really like that, because it effectively means that the autoplay toggle has no effect when playing a playlist?

Can't we just only take `autoPlayEnabled` (so we could drop the `isAutoplayEnabled` method)?

app/src/main/java/com/github/libretube/services/AbstractPlayerService.kt
```diff
@@ -219,6 +221,31 @@ abstract class AbstractPlayerService : MediaLibraryService(), MediaLibrarySessio
...
+            exoPlayer?.seekTo(0)
+            return
+        }
+
+        if (!shouldHandleAutoplay) {
+            return
+        }
+
+        val nextVideo = videoId ?: PlayingQueue.getNext()
+        ?: (if (PlayerHelper.isAutoPlayEnabled()) relatedStreams?.firstOrNull { !it.isLive }?.url?.toID() else null)
```
Bnyro:
```suggestion
        ?: relatedStreams?.firstOrNull { !it.isLive }?.url?.toID()?.takeIf { PlayerHelper.isAutoPlayEnabled() }
```

app/src/main/java/com/github/libretube/ui/fragments/PlayerFragment.kt
```diff
@@ -1232,7 +1232,10 @@ class PlayerFragment : Fragment(R.layout.fragment_player), CustomPlayerCallback
             }
         }
         playerBackgroundBinding.autoplayCountdown.startCountdown {
-            PlayingQueue.getNext()?.let { playNextVideo(it) }
+            val nextVideo = PlayingQueue.getNext()
```
Bnyro: Hmm, it's a bit ugly to duplicate the logic here. Probably fine to move it to a helper method or just drop the autoplay countdown.

Bnyro (commented):
Looks great! I agree that it's much cleaner to not add related streams to the queue.

> https://github.com/libre-tube/LibreTube/issues/8722#issuecomment-5459453250
>
I think that if related streams are hidden, we should also not add them to the queue (as you suggested). If a user doesn't want to see related streams, they certainly also don't want them to be played automatically.

Bnyro (commented):
I think that https://github.com/libre-tube/LibreTube/pull/8733/changes/34b841b94fff5679d3ddd9f46fb56e017ee9fb87 (Always autoplay playlists) is no longer the case now? So the commit should be squashed into the autoplay refactor?

Otherwise lgtm 👍

FineFindus:
> I think that https://github.com/libre-tube/LibreTube/commit/34b841b94fff5679d3ddd9f46fb56e017ee9fb87 (Always autoplay playlists) is no longer the case now? So the commit should be squashed into the autoplay refactor?

No, it's still the case.

> I think that if related streams are hidden, we should also not add them to the queue (as you suggested). If a user doesn't want to see related streams, they certainly also don't want them to be played automatically.

I agree, added in cbbbb203617bc0444ddf38b1ae9aa7a71eed42f8, though I'm overall not sure if it's even worth keeping the preference in general.

Bnyro:
> No, it's still the case.
>
But isn't it very inconvenient that even if you disable autoplay, it still continues playing? For me that would be very unexpected and confusing.

FineFindus:
No, it think it's rather expected (and kind of the point of using a queue/playlist). If a user doesn't want to automatically play the next video, we already have to "Play automatically" setting.

PR #8736, opened 2026-08-30
fix: Showing shorts in feed when using full local mode

### LINKED ISSUE #8427: Shorts don't appear in the subscriptions feed
### Steps to reproduce

All videos and livestreams appear as intended except the shorts form content(in the subscriptions feed). I'm using full local mode and local feed extraction. Shorts filter on. No piped backend.

### Expected behavior

Shorts form content should be showing up in the feed.

### Actual behavior

Doesn't show up.

### LibreTube version

31.1 and 31.2

### Android version

Android 16

### Other details

I have tried clearing cache, data and reimporting subscriptions. Shorts do not appear no matter what. I've also refreshed the feed with or without the shorts filter on, experimenting. Does not mean shorts can't be watched. They just don't show up in the local feed. I saw that this is an existing issue and only want to bring it up. I'm no expert, only interested in this project and wish to help. Do take down this ticket if duplicates are not allowed.

### Acknowledgements

woutersmeenk:
RSS entries for shorts use /shorts/<id> links while the shorts tab returns /watch?v=<id>, so the date fallback never matched and every short was dropped by the upload date filter.

Normalize both forms to the video ID via toID() and use that as key  for feedInfo.

Fixes a regression from a04265764901f26262b71c7d4160f1f92c565bca

Fixes #8427

### DIFF
--- a/app/src/main/java/com/github/libretube/extensions/ToID.kt
+++ b/app/src/main/java/com/github/libretube/extensions/ToID.kt
@@ -14,6 +14,7 @@ fun String.toID(): String {
         .removePrefix(YOUTUBE_MUSIC_URL)
         .removePrefix(YOUTUBE_SHORT_URL)
         .replace("/watch?v=", "") // videos
+        .replace("/shorts/", "") // shorts
         .replace("/channel/", "") // channels
         .replace("/playlist?list=", "") // playlists
         // channel urls for different categories than the main one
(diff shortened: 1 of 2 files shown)


FineFindus (approved):
Good catch! Thanks for the fix.

woutersmeenk:
Found it using AI. Thanks for making a great app!

FineFindus:
That's unfortunate to hear. We don't [allow AI contributions](https://github.com/libre-tube/LibreTube/blob/master/README.md?plain=1#L118) because the copyright situation of such code is unclear, and we believe that the extensive use of AI code might cause maintainability and code quality issues in the future.

We've discussed this internally and decided against reverting this PR as we believe that the code changes are small enough to not present a copyright issue and we're hoping that you only used AI to find the issue, but not to write the fix.

You're of course welcome to contribute using non-AI code. Thanks for understanding.

PR #8743, opened 2026-09-02
refactor: remove option to disable showing skip buttons

### LINKED ISSUE #8725: Remove skip buttons toggle

In the interest of https://github.com/libre-tube/LibreTube/issues/8465 this could be removed. 

### Other details

_No response_

### Acknowledgements

Bnyro:
The option only exists because the skip buttons don't look that nice.

I think that having an option for the skip buttons that is disabled by default is definitely unexpected because it is part of the core functionality to switch between videos.

I thought about making the buttons look better, but idk how, so I kept them looking as before.

Pref migration missing to not conflict with https://github.com/libre-tube/LibreTube/pull/8733. Thus also a draft for now.
closes #8725

### DIFF
--- a/app/src/main/res/xml/player_settings.xml
+++ b/app/src/main/res/xml/player_settings.xml
@@ -52,13 +52,6 @@
             app:valueTo="60.0" />
 
         <SwitchPreferenceCompat
-            android:defaultValue="false"
-            android:icon="@drawable/ic_next"
-            android:summary="@string/skip_buttons_summary"
-            app:key="skip_buttons"
-            app:title="@string/skip_buttons" />
-
-        <SwitchPreferenceCompat
             android:disableDependentsState="true"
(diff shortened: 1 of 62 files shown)


app/src/main/res/xml/player_settings.xml
```diff
@@ -51,13 +51,6 @@
             app:valueFrom="5.0"
             app:valueTo="60.0" />
 
-        <SwitchPreferenceCompat
-            android:defaultValue="false"
-            android:icon="@drawable/ic_next"
-            android:summary="@string/skip_buttons_summary"
-            app:key="skip_buttons"
-            app:title="@string/skip_buttons" />
```
FineFindus, line 59: We should also remove these strings, as they are not used anywhere else.

FineFindus (approved):

PR #8749, opened 2026-09-03
chore: remove unused resources

Bnyro:
Automatically done by Intellij, only tested by rebuilding the app.


FineFindus:
If it builds, it's probably fine, but I've had some trouble in the past, where Intellij would make used resources as unused and then remove them.

Bnyro:
Yeah, that's why I built the app and clicked around a bit.

PR #8751, opened 2026-09-03
fix(poToken): use page-specific context

FineFindus:
Fixes the poToken generation by extracting the `eventId` from the homepage.

Port of https://github.com/InfinityLoop1308/PipePipeClient/pull/86

### DIFF
--- a/app/src/main/java/com/github/libretube/api/poToken/PoTokenWebView.kt
+++ b/app/src/main/java/com/github/libretube/api/poToken/PoTokenWebView.kt
@@ -16,6 +16,8 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.suspendCancellableCoroutine
 import kotlinx.coroutines.withContext
+import okhttp3.OkHttpClient
+import okhttp3.Request
 import java.time.Instant
 import kotlin.coroutines.Continuation
 import kotlin.coroutines.resume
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/github/libretube/api/poToken/PoTokenWebView.kt
```diff
@@ -89,15 +102,20 @@ class PoTokenWebView private constructor(
         }
 
         CoroutineScope(Dispatchers.IO).launch(exceptionHandler) {
-            val responseBody = makeBotguardServiceRequest(
-                "https://www.youtube.com/api/jnn/v1/Create",
-                listOf(REQUEST_KEY)
-            )
-            val parsedChallengeData = parseChallengeData(responseBody)
+            val pageHtml = makeRequest("https://youtube.com/")
+            val (eventId, challengeData) = parseYoutubePageAttestation(pageHtml)
```
FineFindus, line 106: I'm not sure happy that this requires us the make 3 sequential HTTP calls (one of them to the homepage), but I'm also not sure how to avoid them …

Figim:
Pipepipe has also made some changes to the PipePipe Client. Newpipe uses a old Potoken creation system. PipePipe uses SABR Potoken. 

This needs to be fixed to play kids videos. Because the visionos client cannot play kids videos. 

1. You need to copy "potoken.js" to the assets section. Change newpipe extractor potoken.xml 

https://github.com/Priveetee/PipePipeClient/blob/d7bbd20da5cb87b5f88c8fb1c440e100189975d2/app/src/main/assets/sabr_po_token.js

2. ​Instead of porting PipePipe's entire codebase, we should extract ONLY the core JS/WebView logic into a standalone class without any PipePipe-specific extractor dependencies. Change in the Potoken generator 

 https://github.com/Priveetee/PipePipeClient/blob/d7bbd20da5cb87b5f88c8fb1c440e100189975d2/app/src/main/java/org/schabi/newpipe/player/datasource/LocalDomPoTokenGenerator.kt

PR #8752, opened 2026-09-04
 ci(nightly): upload only signed APK 

FineFindus:
Also moves the nightly release action into a separate workflow, which allows us to reduce the permission scope to only that workflow.


Bnyro (approved):
Good idea! Not tested, but looks right 👍

FineFindus:
> Not tested, but looks right 👍

Neither did I ;)

Bnyro:
Worked, nice :)

Issue #8083, opened 2026-01-16
Custom download folder + don't split audio and video + name downloaded files after movie titles


rafnov:
### Describe your suggested feature

My suggestions for LibreTube's better usability: 

1. Let us set custom download folder like other apps do. I was very much surprised when I had to look for downloads in Android/Data/com.github.libretube, but that anyway didn't solve my problem as following came... 
2. Split video from audio files. What's the point? Newpipe/Tubular may combine those media and do the following... 
3. Name downloaded files after human readable movie title, like other apps do. That's so obvious it would let us choosing the proper video we want, later, via file manager, not only Libretube player. 

Please, implement those features for better comfort and usability. I don't want separate app for streaming, another one for local playing and another one for downloading... That's ridiculous, but there is no fully customizable app atm, that does that all. 
Examples: FreeTube broken downloader that gives to many choices and splits video from audio too. 
Newpipe/Tubular: unacceptable accent colours, dependent on a service chosen. 
Please make LibreTube better than them. 
Also, implementing all those three features would let us to easily use our downloads across all our systems and devices.

FineFindus:
Thanks for the suggestion. This has been suggested before and is [something that is explicitly not planned](https://github.com/libre-tube/LibreTube/blob/master/ROADMAP.md#not-planned), see https://github.com/libre-tube/LibreTube/issues/3572 for more details.

Issue #3572, opened 2023-04-16
Export downloads


luckkmaxx:
### Describe your suggested feature

Right now all downloaded media is under Android/data/
Wich is unreachable to other players or apps, and AFAIK cannot be deleted because the app crashes trying to go to "downloads" section.

So maybe the best solution is a new button just like "Download" but "Export", wich will place the selected media in:
  - A folder set under settings.
  - Or a folder selected during the export action.

My case is I want to download music and move to my personal music folder in my device.

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

luckkmaxx:
> NewPipe allows you to set the Download location. I have mine set to `/Podcasts`.

Hi! Thanks for your reply. Im not searching for an app capable to do it. I figure out there are apps that does it, but Im just suggesting implement this here, because is my preferred system right now (to have my own server instance is a must).
But as I said, thanks for your time and help.

Bnyro:
The thing is that downloads are separated into video, audio and subtitles. These are all different files, hence I don't think you would really benefit from having a different location as video and audio can only get played together inside the app. For audio files only, it might make sense, I agree, however it'd probably be confusing for some users and end up in issues being raised that downloaded videos don't have audio.

luckkmaxx:
Ok I understand. I didnt noticed that everything is dowloaded separately since I only downloaded audios until now.

So maybe to be able to export content there has to be another tool (integrated or not) to merge all parts in one file (just like "Tubemate" does, if you know it).

Well, here is my idea, that for sure isnt a priority right now, cuz there are many other relevant fixes and improvements to do.

Thanks!

dhammel:
@Bnyro Is there a reason why audio/video seperated in different files? I can use a different tool on pc to join them, but it would actually be easier to just use yt-dlp, or newpipe to download since it does this automatically. Libretube can replace both the official youtube app and newpipe for me if downloads could save to a folder of my choosing and be joined together.

The downloads feature in libretube is for the most part useless in its current state, and still would need a 2nd app to handle downloading. (Especially the speeds are way too slow over piped proxy, that alone may be worth using another downloader)

Is there some reason you don't seem interested in implementing this?

Bnyro:
Don't ping me and read closed issues.

Bnyro:
[...]

PR #8764, opened 2026-09-12
feat(downloads): add support for exporting downloaded video including audio

Bnyro:
Because it was requested a lot of times (e.g. https://github.com/libre-tube/LibreTube/issues/3572, https://github.com/libre-tube/LibreTube/issues/8083, ...), I decided to give this a chance now.

If both audio and video are downloaded, we have to mux them together using media3. See https://developer.android.com/media/media3/transformer/getting-started.

Muxing audio and video is extremely slow for some reason, don't ask me why.

### DIFF
--- a/app/src/main/java/com/github/libretube/ui/dialogs/DownloadExportDialog.kt
+++ b/app/src/main/java/com/github/libretube/ui/dialogs/DownloadExportDialog.kt
@@ -1,0 +1,106 @@
+package com.github.libretube.ui.dialogs
+
+import android.app.Dialog
+import android.net.Uri
+import android.os.Bundle
+import androidx.annotation.OptIn
+import androidx.appcompat.app.AlertDialog
+import androidx.core.net.toUri
+import androidx.core.view.isGone
+import androidx.core.view.isVisible
+import androidx.fragment.app.DialogFragment
+import androidx.media3.common.util.UnstableApi
+import com.github.libretube.R
+import com.github.libretube.constants.IntentData
+import com.github.libretube.db.DatabaseHolder
+import com.github.libretube.enums.FileType
+import com.github.libretube.extensions.toAndroidUri
+import com.github.libretube.extensions.toastFromMainThread
+import com.github.libretube.util.MediaExporter
+import com.google.android.material.dialog.MaterialAlertDialogBuilder
+import com.google.android.material.loadingindicator.LoadingIndicator
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
+import java.io.File
+
+class DownloadExportDialog : DialogFragment() {
+    @OptIn(UnstableApi::class)
+    private suspend fun exportDownload(videoId: String, outputUri: Uri, onCompletion: () -> Unit) {
+        val activity = requireActivity()
+
+        val downloadItems =
+            DatabaseHolder.Database.downloadDao().getDownloadById(videoId)!!.downloadItems
+        val video = downloadItems.firstOrNull { it.type == FileType.VIDEO }
+        val audio = downloadItems.firstOrNull { it.type == FileType.AUDIO }
+        if (audio == null && video == null) return
+
+        if (audio == null) {
+            MediaExporter.copyFileContents(activity, video!!.path.toAndroidUri(), outputUri)
+            onCompletion()
+        } else if (video == null) {
+            MediaExporter.copyFileContents(activity, audio.path.toAndroidUri(), outputUri)
+            onCompletion()
+        } else {
+            // output stream has to be opened shortly after the file picker was shown, otherwise
+            // the URI gets invalid
+            val outputStream = activity.contentResolver.openOutputStream(outputUri)!!
+
+            // we write everything to a temporary file first because the `MediaExporter`
+            // can't write to URIs directly, only to paths
+            val tempFile = File.createTempFile("mux", ".mp4", activity.cacheDir)
+
+            MediaExporter.muxMedia(
+                activity,
+                audio.path.toAndroidUri(),
+                video.path.toAndroidUri(),
+                tempFile.path,
+                onCompletion = {
+                    // copy to user-specified file
+                    MediaExporter.copyFileContents(
+                        activity.contentResolver.openInputStream(tempFile.toUri())!!,
+                        outputStream
+                    )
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/github/libretube/ui/dialogs/DownloadExportDialog.kt
```diff
@@ -0,0 +1,106 @@
...
+            .setTitle(R.string.exporting)
+            .setView(loadingIndicator)
+            .setPositiveButton(R.string.okay, null)
+            .setCancelable(false)
+            .show()
+            .also {
+                it.getButton(Dialog.BUTTON_POSITIVE)?.isGone = true
+                dialog = it
+            }
+    }
```
FineFindus, line 105: I think the code flow here is rather confusing (first the function returns, then the `exportDownload` callback is run?)

app/src/main/java/com/github/libretube/ui/dialogs/DownloadExportDialog.kt
```diff
@@ -0,0 +1,106 @@
...
+            MediaExporter.copyFileContents(activity, video!!.path.toAndroidUri(), outputUri)
+            onCompletion()
+        } else if (video == null) {
+            MediaExporter.copyFileContents(activity, audio.path.toAndroidUri(), outputUri)
+            onCompletion()
+        } else {
+            // output stream has to be opened shortly after the file picker was shown, otherwise
+            // the URI gets invalid
+            val outputStream = activity.contentResolver.openOutputStream(outputUri)!!
+
```
FineFindus, line 48:
```suggestion
      when {
          audio != null && video != null -> exportMultiMedia(...)
          video != null -> {
                MediaExporter.copyFileContents(activity, video!!.path.toAndroidUri(), outputUri)
                onCompletion()
            }
          audio != null -> {
                MediaExporter.copyFileContents(activity, audio!!.path.toAndroidUri(), outputUri)
                onCompletion()
            }
        }
```

I think this would be easier to read when using a subject-less `when` expression, e.g. something like this.

app/src/main/java/com/github/libretube/ui/sheets/DownloadOptionsBottomSheet.kt
```diff
@@ -16,22 +18,41 @@ import com.github.libretube.helpers.NavigationHelper
...
+        videoId = streamItem.url!!.toID()
         val downloadTab = arguments?.serializable<DownloadTab>(IntentData.downloadTab)!!
         val playlistId = arguments?.getString(IntentData.playlistId)
 
         val options = mutableListOf(
             R.string.playOnBackground,
             R.string.share,
-            R.string.delete
+            R.string.delete,
+            R.string.export
```
FineFindus, line 55:
```suggestion
            R.string.delete,
            R.string.export,
```

nit: to avoid the double line change next time

app/src/main/java/com/github/libretube/ui/dialogs/DownloadExportDialog.kt
```diff
@@ -0,0 +1,106 @@
...
+        return MaterialAlertDialogBuilder(requireContext())
+            .setTitle(R.string.exporting)
+            .setView(loadingIndicator)
+            .setPositiveButton(R.string.okay, null)
+            .setCancelable(false)
+            .show()
+            .also {
+                it.getButton(Dialog.BUTTON_POSITIVE)?.isGone = true
+                dialog = it
+            }
```
FineFindus, line 104: Considering the export may take some time, maybe it's better to do it in a background job with a progress notification (rather than blocking the user from doing anything else during that time)?

FineFindus (dismissed):
> Muxing audio and video is extremely slow for some reason, don't ask me why.

Knowing more than I wanted to know about `media3` internals, muxing is probably implemented in pure Java…

I'm honestly still don't think that this is something that should be implemented, as users who need such features are almost always better served by dedicated downloading apps.

app/src/main/java/com/github/libretube/ui/dialogs/DownloadExportDialog.kt
Bnyro: I thought about that - but it would make the code more complicated because we have to implement a background service and I'm too lazy to implement it properly with notifications etc ;)

Bnyro:
> I'm honestly still don't think that this is something that should be implemented, as users who need such features are almost always better served by dedicated downloading apps.
>
Well, if you already have things downloaded in LibreTube, then it is probably faster to just export that instead of installing a downloader app first and figuring out how it works.

But yeah, I agree that this is not LibreTube's main focus (as we already discussed in other issues), and we still default to only keeping downloads internal to LibreTube.

app/src/main/java/com/github/libretube/ui/dialogs/DownloadExportDialog.kt
Bnyro: Good point! I wasn't happy with my approach, this is much better

app/src/main/java/com/github/libretube/ui/dialogs/DownloadExportDialog.kt
Bnyro: True - that was needed previously because I did some hacky stuff with the dialog view - cleaned it up now.

app/src/main/java/com/github/libretube/ui/dialogs/DownloadExportDialog.kt
Bnyro: Let's keep it as is for now and move it to a service if people start complaining.

app/src/main/java/com/github/libretube/util/MediaExporter.kt
```diff
@@ -0,0 +1,74 @@
...
+        }
+    }
+
+    fun copyFileContents(input: InputStream, output: OutputStream) {
+        output.use { outputStream ->
+            input.use { inputStream ->
+                inputStream.copyTo(outputStream)
+            }
+        }
+    }
```
FineFindus:
```suggestion
    fun copyFileContents(input: InputStream, output: OutputStream) {
        input.copyTo(output)
    }
```

Kotlin has a [`copyTo`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.io/copy-to.html) function.

FineFindus (approved):

PR #8787, opened 2026-09-22
fix(downloads): open a populated tab by default

### LINKED ISSUE #8786: Select a populated tab when opening Downloads

Open Downloads on a populated tab when Video is empty. With only audio downloads, select Audio automatically. Preserve an explicitly selected or restored tab.

### Other details

On Pixel 5 / Android 14, with three audio downloads and no videos or playlists, Library → Downloads shows “Nothing here” in Video. All three downloads are visible after switching to Audio.

Reproduced on master c08566c5b (32.1). The same UI assertion fails before the change and passes afterward on the device.

### Acknowledgements

ethicnology-agent:
With only audio downloads, opening Downloads currently shows an empty Video tab. Select the first populated tab on entry, keeping Video when it has items or all tabs are empty. Preserve restored state and any selection made while the database query runs.

Fixes #8786.

Validation: the same UI assertion fails on master and passes with this change on Pixel 5 / Android 14. Six unit tests cover audio-only, mixed media, playlist-only, empty and unstarted downloads.


FineFindus:

However, as the pull request unfortunately looks like it's AI-generated, we must assume that the same applies for the code submitted here, and hence we must reject this PR.

Thanks for understanding.

